PR #188, opened 2017-12-12
[PROTOTYPE] Add SparseListIndexProxy prototype

alexander-irbis:
Development of ideas of #185

This code ain't supposed to be production-ready or great, it's a (throwaway) prototype. Though it has to work correctly.


dmitry-timofeev:
:+1: 

Closing as MerkleDB will have another set of collections, and sparse list is unlikely to be needed before then.

PR #219, opened 2018-02-26
WIP Add a test leaking a native peer [ECR-892]:

dmitry-timofeev:
## Overview 
The problem with most index constructors is that they first
create a native peer, and then perform some checks, e.g., in
ProofListIndex constructor:

```java
super(nativeCreate(checkIndexName(name),
  view.getViewNativeHandle()), name, view, serializer);
```

Therefore, the parent receives a native handle, and then checks
some of the other arguments (serializer, in this case). If these checks
fail, we get a leak of a native peer.

A possible solution is to simplify error checking with static factory
methods, that do not have such a restriction as constructors that the first call
in the method must be a call to `super`. To make it even more reliable, we may
also wrap the code in try/finally, destroying the native peer.


dmitry-timofeev:
Closing as #244 made this test obsolete.

PR #228, opened 2018-03-22
Add an experimental QA service controller accepting transactions in JSON.

dmitry-timofeev:
## Overview 

Transactions in JSON are probably not the best choice for a QA-service,
for it’s not convenient to create JSON messages with some integer ids.


dmitry-timofeev:
Closing, we now have a controller accepting transactions in JSON from the light client in the cryptocurrency demo.

PR #230, opened 2018-03-26
Add Guava beta checker

dmitry-timofeev:
## Overview

It ensures that our library code does not depend on beta APIs
in Guava, which may change in any release.

Also, fix a Beta API use in HashCode.

See https://github.com/google/guava/wiki/PhilosophyExplained#beta-apis


dmitry-timofeev:
The following is likely to be the solution for this issue: https://groups.google.com/forum/#!topic/error-prone-discuss/O4VJTvvjBkM

PR #245, opened 2018-05-02
Make destruction of native objects implicit/v2 [ECR-595]

dmitry-timofeev:


dmitry-timofeev:
Closing as we went the other way : #244

PR #250, opened 2018-05-09
ECR-1164 Add flat proof

MakarovS:
https://jira.bf.local/browse/ECR-1164


exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/DbKeyComparisonParameterizedTest.java
```diff
@@ -0,0 +1,68 @@
...
+  public DbKey secondKey;
+
+  @Parameter(2)
+  public boolean expectedResult;
+
+  @Parameter(3)
+  public String description;
+
+  @Test
+  public void commonPrefixTest() {
```
dmitry-timofeev: This name is off :upside_down_face:

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/DbKeyComparisonParameterizedTest.java
```diff
@@ -0,0 +1,68 @@
...
+    int comparisonResult = firstKey.compareTo(secondKey);
+    assertThat(comparisonResult > 0, equalTo(expectedResult));
+  }
+
+  @Parameters(name = "{index} = {3}")
+  public static Collection<Object[]> testData() {
+    return Arrays.asList(
+        // "A | B -> C" reads "C is a common prefix of A and B"
```
dmitry-timofeev: the comment is off :upside_down_face:

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/DbKeyComparisonParameterizedTest.java
```diff
@@ -0,0 +1,68 @@
...
+        parameters(
+            DbKey.newBranchKey(createUserKey(0b1100), 4),
```
dmitry-timofeev: it would be great to migrate these tests to the new method `com.exonum.binding.storage.proofs.map.DbKeyCommonPrefixParameterizedTest.branchKeyFromPrefix`

dmitry-timofeev: We may actually gather the various factory method used in various tests in a utility class under /test directory (e.g., DbKeys#branchKeyFromPrefix, DbKeys#leafKeyFromPrefix, + other)

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/DbKeyCommonPrefixParameterizedTest.java
```diff
@@ -0,0 +1,97 @@
...
+            branchKeyFromPrefix("11101"),
+            branchKeyFromPrefix("11111"),
+            branchKeyFromPrefix("111"),
+            "[11101] | [11111] -> [111]"),
+        parameters(
+            branchKeyFromPrefix("11111"),
+            branchKeyFromPrefix("10111"),
+            branchKeyFromPrefix("1"),
+            "[11111] | [11101] -> [1]"));
+  }
```
dmitry-timofeev: I'd add some tests with leaf keys.

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/DbKeyCommonPrefixParameterizedTest.java
```diff
@@ -0,0 +1,97 @@
...
+  public void commonPrefixTest() {
+    DbKey actualCommonPrefixKey = firstKey.commonPrefix(secondKey);
+    assertThat(actualCommonPrefixKey, equalTo(expectedResultKey));
+  }
+
+  @Parameters(name = "{index} = {3}")
+  public static Collection<Object[]> testData() {
+    return Arrays.asList(
+        // "A | B -> C" reads "C is a common prefix of A and B"
+        parameters(
```
dmitry-timofeev: I'd consider adding section comments to help avoid duplicate cases and explain what you intend to test:

// # Not a prefix:

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/DbKeyCommonPrefixParameterizedTest.java
```diff
@@ -0,0 +1,97 @@
...
+            branchKeyFromPrefix("01"),
+            branchKeyFromPrefix("10"),
+            branchKeyFromPrefix(""),
+            "[01] | [10] -> []"),
+        parameters(
```
dmitry-timofeev: // # Prefixes of various length:

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/DbKeyCommonPrefixParameterizedTest.java
```diff
@@ -0,0 +1,97 @@
...
+            branchKeyFromPrefix("11101"),
+            branchKeyFromPrefix("11111"),
+            branchKeyFromPrefix("111"),
+            "[11101] | [11111] -> [111]"),
+        parameters(
+            branchKeyFromPrefix("11111"),
+            branchKeyFromPrefix("10111"),
+            branchKeyFromPrefix("1"),
+            "[11111] | [11101] -> [1]"));
```
dmitry-timofeev: I've added some test cases (some are failing :warning: ):

```java
        // ## Multi-byte keys
        parameters(
            branchKeyFromPrefix("1111 1111 | 10_11"),
            branchKeyFromPrefix("1111 1111 | 10_00"),
            branchKeyFromPrefix("1111 1111 | 10"),
            "[1111 1111 | 10_11] | [1111 1111 | 10_00] -> [1111 1111 | 10]"),
        // ## One is full prefix of another:
        parameters(
            branchKeyFromPrefix("11"),
            branchKeyFromPrefix("111"),
            branchKeyFromPrefix("11"),
            "[11] | [111] -> [11]"),
        parameters(
            branchKeyFromPrefix("10"),
            branchKeyFromPrefix("100"),
            branchKeyFromPrefix("10"),
            "[10] | [100] -> [10]"),
        parameters(
            branchKeyFromPrefix("100"),
            branchKeyFromPrefix("10"),
            branchKeyFromPrefix("10"),
            "[100] | [10] -> [10]"),
        parameters(
            branchKeyFromPrefix("0000"),
            branchKeyFromPrefix("00"),
            branchKeyFromPrefix("00"),
            "[0000] | [00] -> [00]"),
        parameters(
            branchKeyFromPrefix("0"),
            branchKeyFromPrefix("000"),
            branchKeyFromPrefix("0"),
            "[0] | [000] -> [0]"),
        // ## Equal keys:
        parameters(
            branchKeyFromPrefix("00"),
            branchKeyFromPrefix("00"),
            branchKeyFromPrefix("00"),
            "[00] | [00] -> [00]"),
        parameters(
            branchKeyFromPrefix("11"),
            branchKeyFromPrefix("11"),
            branchKeyFromPrefix("11"),
            "[11] | [11] -> [11]")
        );
```

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedMapProof.java
```diff
@@ -0,0 +1,15 @@
+package com.exonum.binding.storage.proofs.map.flat;
+
+import java.util.List;
+
+public interface UncheckedMapProof {
```
dmitry-timofeev: Please add Javadoc

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/CheckedMapProofEntry.java
```diff
@@ -0,0 +1,42 @@
+package com.exonum.binding.storage.proofs.map.flat;
+
+import java.util.Arrays;
+
+/**
+ * A map entry corresponding to a leaf node in the map tree.
```
dmitry-timofeev: I'd use this because this class does not have any connection to a leaf node:
```
A map entry: a key-value pair. This entry does not permit null keys and values
```

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/CheckedMapProofEntry.java
```diff
@@ -0,0 +1,42 @@
...
+  private final byte[] key;
+
+  private final byte[] value;
+
+  CheckedMapProofEntry(byte[] key, byte[] value) {
+    this.key = key;
+    this.value = value;
+  }
+
+  public byte[] getKey() {
```
dmitry-timofeev: /** Returns the key in this entry. */

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/CheckedMapProofEntry.java
```diff
@@ -0,0 +1,42 @@
...
+    return key;
+  }
+
+  public byte[] getValue() {
```
dmitry-timofeev: /** Returns the value in this entry. */

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/MapProofEntryLeaf.java
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  MapProofEntryLeaf(DbKey dbKey, byte[] value, HashFunction hashFunction) {
+    super(dbKey);
+    checkArgument(dbKey.getNodeType() == Type.LEAF, "dbKey should be of type LEAF", dbKey);
+    this.value = value;
+    this.hashFunction = hashFunction;
+  }
+
+  public byte[] getValue() {
```
dmitry-timofeev: /** Returns the value in this entry. */

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/CheckedFlatMapProof.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * A checked flat map proof, which does not include any intermediate nodes.
+ */
+public class CheckedFlatMapProof implements CheckedMapProof {
+
+  @Nullable
+  private List<CheckedMapProofEntry> entryList;
```
dmitry-timofeev: `entryList` and `rootHash` seem to no longer be nullable.

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -0,0 +1,169 @@
...
+
+  @Override
+  public CheckedMapProof check() {
+    ProofStatus orderCheckResult = orderCheck();
+    List<CheckedMapProofEntry> leafList;
+    if (orderCheckResult != ProofStatus.CORRECT) {
+      return CheckedFlatMapProof.invalid(orderCheckResult);
+    }
+    if (proofList.isEmpty()) {
+      leafList = extractEntries(proofList);
```
dmitry-timofeev: Just `= emptyList();`?

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/flat/CheckedFlatMapProofTest.java
```diff
@@ -0,0 +1,123 @@
...
+import com.exonum.binding.hash.HashCode;
+import com.exonum.binding.hash.HashFunction;
+import com.exonum.binding.hash.Hashing;
+import com.exonum.binding.storage.proofs.map.DbKey;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import org.junit.Test;
+
+public class CheckedFlatMapProofTest {
```
dmitry-timofeev: I'd rename it into **Un**checked as it tests mostly that class — this way navigation will work properly!

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/flat/CheckedFlatMapProofTest.java
```diff
@@ -0,0 +1,123 @@
...
+        leaf,
+        createBranchMapEntry(thirdKey)
+    );
+    UncheckedMapProof uncheckedFlatMapProof = new UncheckedFlatMapProof(entries);
+
+    CheckedMapProof checkedMapProof = uncheckedFlatMapProof.check();
+    assertThat(entries, equalTo(uncheckedFlatMapProof.getProofList()));
+
+    CheckedMapProofEntry checkedEntry = new CheckedMapProofEntry(valueKey.getKeySlice(), VALUE);
+    assertThat(Collections.singletonList(checkedEntry), equalTo(checkedMapProof.getEntries()));
```
dmitry-timofeev: I think it is the other way around:

assertThat(checkedMapProof.getEntries()/*: actual */, equalTo(Collections.singletonList(checkedEntry)/* expected: */));

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/flat/CheckedFlatMapProofTest.java
```diff
@@ -0,0 +1,123 @@
...
+        createBranchMapEntry(firstKey),
+        leaf,
+        createBranchMapEntry(thirdKey)
+    );
+    UncheckedMapProof uncheckedFlatMapProof = new UncheckedFlatMapProof(entries);
+
+    CheckedMapProof checkedMapProof = uncheckedFlatMapProof.check();
+    assertThat(entries, equalTo(uncheckedFlatMapProof.getProofList()));
+
+    CheckedMapProofEntry checkedEntry = new CheckedMapProofEntry(valueKey.getKeySlice(), VALUE);
```
dmitry-timofeev: expectedEntry?

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/flat/CheckedFlatMapProofTest.java
```diff
@@ -0,0 +1,123 @@
...
+    assertThat(Collections.singletonList(checkedEntry), equalTo(checkedMapProof.getEntries()));
+    assertTrue(checkedMapProof.containsKey(valueKey.getKeySlice()));
```
dmitry-timofeev: There are four places where valueKey.getKeySlice() is used (essentially, that's the `secondKey`). Every time it forces the reader to think 'what is a key slice in this case'. What if we rename the `secondKey` to `valueKey`, and present `valueKey` to `valueDbKey` to avoid confusion?

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/flat/CheckedFlatMapProofTest.java
```diff
@@ -0,0 +1,123 @@
...
+    assertThat(VALUE, equalTo(checkedMapProof.get(valueKey.getKeySlice())));
+  }
+
+  @Test
+  public void mapProofWithOneElementShouldBeValid() {
+    byte[] key = createPrefixed(bytes(0b10), DbKey.KEY_SIZE);
+    DbKey valueKey = DbKey.newLeafKey(key);
```
dmitry-timofeev: dbKey?

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/flat/CheckedFlatMapProofTest.java
```diff
@@ -0,0 +1,123 @@
...
+        .putObject(HASH_FUNCTION.hashBytes(VALUE), hashCodeFunnel())
+        .hash();
+    List<MapProofEntry> entries =
+        Collections.singletonList(createLeafMapEntry(valueKey.getKeySlice(), VALUE));
+    UncheckedMapProof uncheckedFlatMapProof = new UncheckedFlatMapProof(entries);
+    CheckedMapProof checkedMapProof = uncheckedFlatMapProof.check();
+
+    assertThat(expectedRootHash, equalTo(checkedMapProof.getMerkleRoot()));
+
+    CheckedMapProofEntry checkedEntry = new CheckedMapProofEntry(valueKey.getKeySlice(), VALUE);
```
dmitry-timofeev: Just `key` (same as in the previous test case)?

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/flat/CheckedFlatMapProofTest.java
```diff
@@ -0,0 +1,123 @@
...
+    assertThat(Collections.singletonList(checkedEntry), equalTo(checkedMapProof.getEntries()));
```
dmitry-timofeev: Same regarding the order of arguments (assert that actual matches sth).

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -0,0 +1,176 @@
...
+ */
+public class UncheckedFlatMapProof implements UncheckedMapProof {
+
+  private static final HashFunction HASH_FUNCTION = Hashing.defaultHashFunction();
+
+  private final List<MapProofEntry> proofList;
+
+  private final Set<byte[]> requestedKeys;
+
+  UncheckedFlatMapProof(List<MapProofEntry> proofList, Set<byte[]> requestedKeys) {
```
dmitry-timofeev: The unchecked map proof already contains some keys for which it provides a proof as MapProofEntries. It shall not carry the _requested_ keys — it is the caller of the map validator who supplies them.

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -0,0 +1,176 @@
...
+    return new UncheckedFlatMapProof(mapProofEntries, expectedKeys);
+  }
+
+  @Override
+  public List<MapProofEntry> getProofList() {
+    return proofList;
+  }
+
+  @Override
+  public CheckedMapProof check() {
```
dmitry-timofeev: The validator method has to check that the actual proof proves that the requested keys are either in the map or not in the map.

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -0,0 +1,176 @@
...
+  UncheckedFlatMapProof(List<MapProofEntry> proofList, Set<byte[]> requestedKeys) {
+    this.proofList = proofList;
+    this.requestedKeys = requestedKeys;
+  }
+
+  @SuppressWarnings("unused") // Native API
+  static UncheckedFlatMapProof fromUnsorted(MapProofEntry[] proofList, byte[][] requestedKeys) {
+    List<MapProofEntry> mapProofEntries = Arrays.asList(proofList);
+    mapProofEntries.sort(Comparator.comparing(MapProofEntry::getDbKey));
+    Set<byte[]> expectedKeys = new HashSet<>(Arrays.asList(requestedKeys));
```
dmitry-timofeev: I’ve done the same mistake — an array does not override equals and hashCode, therefore, cannot be reliably used in a _hash_ set. If we need to have a set of byte[], then we have to add a ByteString class or use a specialized set (e.g., a TreeSet + a comparator).

I’d consider adding a ByteString interface with the minimum required methods and static factory methods + ArrayByteString implementing that interface — this way it will be easier to replace it if needed.

Some existing implementations (overkill for our needs at the moment, just for reference): 
* https://github.com/google/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/ByteString.java or
* https://github.com/kbolino/ByteString/blob/master/src/main/java/com/kbolino/libraries/bytestring/ByteString.java

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
MakarovS: Unchecked map proof won't contain absent keys. Without requested keys it contains only list of MapProofEntries and from it we have no reliable way to know what keys were requested. Plus we can't derive missing keys without knowing what keys were requested.

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/CheckedFlatMapProof.java
```diff
@@ -0,0 +1,128 @@
...
+        .stream()
+        .map(CheckedMapProofAbsentEntry::getKey)
+        .collect(Collectors.toList());
+  }
+
+  @Override
+  public boolean containsKey(byte[] key) {
+    checkValid();
+    checkThatKeyIsRequested(key);
+    for (CheckedMapProofEntry entry: entries) {
```
bullet-tooth: can be replaced by:
```
return entries.stream().anyMatch(entry -> Arrays.equals(entry.getKey(), key));
```

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/CheckedFlatMapProof.java
```diff
@@ -0,0 +1,128 @@
...
+  public HashCode getRootHash() {
+    checkValid();
+    return rootHash;
+  }
+
+  @Override
+  public byte[] get(byte[] key) {
+    checkValid();
+    checkThatKeyIsRequested(key);
+    for (CheckedMapProofEntry entry: entries) {
```
bullet-tooth: Can be replaced by:
```
    entries.stream()
        .filter(entry -> Arrays.equals(entry.getKey(), key))
        .map(CheckedMapProofEntry::getValue)
        .findFirst()
        .orElse(null);
```

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/CheckedFlatMapProof.java
```diff
@@ -0,0 +1,128 @@
...
+    checkValid();
+    return rootHash.equals(expectedRootHash);
+  }
+
+  private void checkValid() {
+    checkState(status == ProofStatus.CORRECT, "Proof is not valid: %s", status);
+  }
+
+  private void checkThatKeyIsRequested(byte[] key) {
+    for (byte[] entryKey :
```
bullet-tooth: can be refactored to Java 8 streams:
```
    Stream.concat(
        entries.stream().map(CheckedMapProofEntry::getKey),
        absentEntries.stream().map(CheckedMapProofAbsentEntry::getKey))
        .filter(entryKey -> Arrays.equals(entryKey, key))
        .findFirst()
        .orElseThrow(
            () -> new IllegalArgumentException("Key that wasn't among requested keys was checked"));
```

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -0,0 +1,225 @@
...
+
+  /**
+   * Compute hashes of leaf entries and merge them into list of branches.
+   */
+  private void mergeLeavesWithBranches() {
+    proofList.addAll(branches);
+    proofList.addAll(
+        leaves
+            .stream()
+            .map(
```
bullet-tooth: format to 1 line

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -0,0 +1,225 @@
...
+    contour.push(newEntry);
+    return commonPrefix;
+  }
+
+  /*
+   * Check that no entry has a prefix among the paths in the proof entries. Both found and absent
+   * keys are checked.
+   */
+  private boolean prefixesCheck() {
+    List<DbKey> allLeaves =
```
bullet-tooth: How about? 
```
    return Stream.concat(
        leaves.stream().map(MapProofEntryLeaf::getDbKey), 
        absentLeaves.stream().map(MapProofAbsentEntryLeaf::getDbKey))
        .anyMatch(leafEntryKey -> branches.stream()
            .map(MapProofEntry::getDbKey)
            .anyMatch(anotherEntryKey -> leafEntryKey.commonPrefix(anotherEntryKey).equals(anotherEntryKey))
        );
```

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -0,0 +1,225 @@
...
+      }
+    }
+    return true;
+  }
+
+  private List<CheckedMapProofEntry> convertIntoCheckedProofEntries() {
+    return leaves
+        .stream()
+        .map(l -> new CheckedMapProofEntry(l.getDbKey().getKeySlice(), l.getValue()))
+        .collect(Collectors.toList());
```
bullet-tooth: static import

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -27,53 +27,49 @@
...
-    List<MapProofEntryLeaf> leavesList = Arrays.asList(leaves);
-    List<MapProofAbsentEntryLeaf> absentLeavesList = Arrays.asList(absentLeaves);
-    return new UncheckedFlatMapProof(branchesList, leavesList, absentLeavesList);
-  }
-
-  @Override
-  public List<MapProofEntry> getProofList() {
-    return new ArrayList<>(proofList);
+      MapEntry[] leaves,
+      byte[] absentLeaves) {
```
dmitry-timofeev: Shan't it be byte[][]?

dmitry-timofeev: I'd avoid using the same name (= leaf) for different types: we can say that both MapEntry and missing keys _correspond_ to leaf nodes in the map tree, but we can't say that they _are_ leaf nodes (as a leaf node is a 34-byte _database_ key + value). Therefore I'd suggest using `entries` and `missingKeys`.

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -152,13 +143,18 @@ private ProofStatus preCheck() {
...
+                absentLeaves.stream().map(DbKey::newLeafKey))
+            .collect(toList());
+    return requestedKeys
+        .stream()
+        .anyMatch(leafEntryKey -> proof.stream()
             .map(MapProofEntry::getDbKey)
-            .anyMatch(branchEntryKey ->
-                leafEntryKey.commonPrefix(branchEntryKey).equals(branchEntryKey))
+            .anyMatch(proofEntryKey ->
+                leafEntryKey.commonPrefix(proofEntryKey).equals(proofEntryKey))
```
dmitry-timofeev: I’d suggest adding a `DbKey#isPrefixOf(other: DbKey)` for easier reading, testing and faster implementation (you don't need to create a common prefix to see if one key is a prefix of another):

```java
  public boolean isPrefixOf(DbKey other) {
    return this.keyBits().isPrefixOf(other.keyBits());
  }
```

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -27,53 +27,49 @@
...
   }
 
   @Override
   public CheckedMapProof check() {
-    ProofStatus orderCheckResult = preCheck();
+    if (prefixesIncluded()) {
+      return CheckedFlatMapProof.invalid(ProofStatus.INVALID_STRUCTURE);
+    }
+    mergeLeavesWithBranches();
+    ProofStatus orderCheckResult = orderCheck();
```
dmitry-timofeev: We shall check the order before merging, shan't we? The list being sorted _after_ merging must be our internal invariant.

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProofTest.java
```diff
@@ -22,79 +21,76 @@
...
         expectedEntriesList,
         containsInAnyOrder(
             actualCheckedEntriesList.toArray()));
 
-    assertTrue(checkedMapProof.containsKey(firstValueKey));
-    assertThat(VALUE, equalTo(checkedMapProof.get(firstValueKey)));
-    assertThat(secondValue, equalTo(checkedMapProof.get(secondValueKey)));
-    assertThat(fourthValue, equalTo(checkedMapProof.get(fourthValueKey)));
+    assertTrue(checkedMapProof.containsKey(firstKey));
+    assertThat(FIRST_VALUE, equalTo(checkedMapProof.get(firstKey)));
```
dmitry-timofeev: Here and in other places expected/actual are confused:
```java
assertThat(actual, matches(expected));
```

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProofTest.java
```diff
@@ -141,60 +136,71 @@ public void mapProofWithEqualEntriesOrderShouldBeInvalid() {
...
     CheckedMapProof checkedMapProof = uncheckedFlatMapProof.check();
     assertThat(checkedMapProof.getStatus(), equalTo(ProofStatus.CORRECT));
   }
 
   @Test
-  public void mapProofWithAbsentKeyShouldBeCorrect() {
-    byte[] firstKey = createPrefixed(bytes(0b001101), DbKey.KEY_SIZE);
-    byte[] valueKey = createPrefixed(bytes(0b011101), DbKey.KEY_SIZE);
-    byte[] absentKey = createPrefixed(bytes(0b111101), DbKey.KEY_SIZE);
+  public void mapProofWithSingleProofEntryShouldBeInvalid() {
```
dmitry-timofeev: I'd suggest a clearer name:
`mapProofWithSingleBranchProofEntryShouldBeInvalid`

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProofTest.java
```diff
@@ -141,60 +136,71 @@ public void mapProofWithEqualEntriesOrderShouldBeInvalid() {
...
+  @Test
+  public void mapProofWithIncludedPrefixesShouldBeInvalid() {
+    DbKey firstDbKey = DbKeyTestUtils.branchKeyFromPrefix("11");
+    DbKey secondDbKey = DbKeyTestUtils.branchKeyFromPrefix("01");
+    byte[] absentKey = DbKeyTestUtils.keyFromString("111111");
+
+    UncheckedMapProof uncheckedFlatMapProof =
+        new UncheckedFlatMapProof(
+            Arrays.asList(
+                new MapProofEntry(firstDbKey, HashCode.fromBytes(firstDbKey.getKeySlice())),
```
dmitry-timofeev: createBranchMapEntry(firstDbKey)?

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProofTest.java
```diff
@@ -141,60 +136,71 @@ public void mapProofWithEqualEntriesOrderShouldBeInvalid() {
...
+            Collections.singletonList(absentKey));
+
+    CheckedMapProof checkedMapProof = uncheckedFlatMapProof.check();
+    assertThat(checkedMapProof.getStatus(), equalTo(ProofStatus.INVALID_STRUCTURE));
   }
 
-  private static MapProofEntryLeaf createLeafMapEntry(byte[] key, byte[] value) {
-    byte[] prefixedKey = createPrefixed(key, DbKey.KEY_SIZE);
-    return new MapProofEntryLeaf(DbKey.newLeafKey(prefixedKey), value);
+  private static MapProofEntry createBranchMapEntry(DbKey key) {
```
dmitry-timofeev: proof _might_ contain proof nodes, corresponding to leaves, plus this method does not have branch-specific logic.
I'd suggest `createProofEntry(DbKey key)` or `createMapProofEntry(DbKey key)`

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProofTest.java
```diff
@@ -141,60 +136,71 @@ public void mapProofWithEqualEntriesOrderShouldBeInvalid() {
...
+    return new MapProofEntry(
+        key,
+        HashCode.fromBytes(key.getKeySlice()));
   }
 
-  private static MapProofAbsentEntryLeaf createAbsentLeafMapEntry(byte[] key) {
-    byte[] prefixedKey = createPrefixed(key, DbKey.KEY_SIZE);
-    return new MapProofAbsentEntryLeaf(DbKey.newLeafKey(prefixedKey));
+  private static MapEntry createLeafMapEntry(byte[] key, byte[] value) {
```
dmitry-timofeev: just `createMapEntry`?

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -0,0 +1,201 @@
...
+    if (orderCheckResult != ProofStatus.CORRECT) {
+      return CheckedFlatMapProof.invalid(orderCheckResult);
+    }
+    mergeLeavesWithBranches();
+    if (proofList.isEmpty()) {
+      return CheckedFlatMapProof.correct(
+          getEmptyProofListHash(), Collections.emptyList(), missingKeys);
+    } else if (proofList.size() == 1) {
+      // One element proof is correct only if the node is a leaf
+      if (proof.size() == 1 && proof.get(0).getDbKey().getNodeType() != Type.LEAF) {
```
dmitry-timofeev: This code is rather hard to understand. It also seems more complicated than it is:
image

I’d go with something like that:
```java

  private CheckedMapProof checkSingletonProof() {
    assert proof.size() + entries.size() == 1;

    if (proof.size() == 1) {
      // There are no entries, therefore, the proof node must correspond to a leaf.
      MapProofEntry entry = proof.get(0);
      DbKey.Type nodeType = entry.getDbKey().getNodeType();
      if (nodeType == Type.BRANCH) {
        return CheckedFlatMapProof.invalid(ProofStatus.NON_TERMINAL_NODE);
      } else {
        // TODO: calculate hash properly.
      }
    } else {
      // TODO: calculate hash properly.
    }
  }

```

It also must be decided where we calculate the branch (root) hash: here or determine in `merge…`. In the first case we have a merge that does not work correctly for singleton maps, doing meaningless operations. In the second case we have a rather complicated merge method.

Probably, it would be better to refactor in the following way:
```
def check:
  checkNoPrefixes
  checkOrder
  if isEmptyTree():
    return checkEmptyTree()
  elif isSingletonTree():
    return checkSingletonTree()
  else:
    return checkTree() // Performs merge internally
```

The first case (calculating it here after meaningless merge) is also acceptable.

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -0,0 +1,201 @@
...
+
+  /**
+   * Compute hashes of leaf entries and merge them into list of branches.
+   */
+  private void mergeLeavesWithBranches() {
+    proofList.addAll(proof);
+    List<MapProofEntry> leafEntries =
+        entries
+            .stream()
+            .map(e -> new MapProofEntry(DbKey.newLeafKey(e.getKey()), getSingleLeafHash(e)))
```
dmitry-timofeev: I'm not sure it is correct to use `getSingleLeafHash`: 
> Hash of leaf node  is defined as *hash( __[serialized\_value]__ )*.

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/ProofStatus.java
```diff
@@ -0,0 +1,25 @@
+package com.exonum.binding.storage.proofs.map.flat;
+
+/**
+ * Possible statuses of a checked map proof.
+ */
+public enum ProofStatus {
+  CORRECT("Proof has a valid structure"),
+  NON_TERMINAL_NODE("Leaf entry is of branch type"),
```
dmitry-timofeev: "Proof entry in a singleton proof is of branch type (must be a leaf)"

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/DbKey.java
```diff
@@ -187,4 +194,88 @@ public int getNumSignificantBits() {
...
+    }
+    int commonPrefixSize = Math.min(firstSetBitIndex, minPrefixSize);
+    byte[] resultingByteArray = this.keyBits().getKeyBits().get(0, firstSetBitIndex).toByteArray();
+    byte[] newArray = new byte[DbKey.KEY_SIZE];
+    System.arraycopy(resultingByteArray, 0, newArray, 0, resultingByteArray.length);
+    return newBranchKey(newArray, commonPrefixSize);
+  }
+
+  /**
+   * Returns whether this {@code DbKey} is a prefix of that {@code DbKey}.
```
dmitry-timofeev: Returns true if …

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -0,0 +1,249 @@
...
+      DbKey.Type nodeType = entry.getDbKey().getNodeType();
+      if (nodeType == Type.BRANCH) {
+        return CheckedFlatMapProof.invalid(ProofStatus.NON_TERMINAL_NODE);
+      } else {
+        return CheckedFlatMapProof.correct(getSingleEntryProofHash(entry), entries, missingKeys);
+      }
+    } else {
+      // The proof consists of a single leaf with a required key
+      MapEntry entry = entries.get(0);
+      return CheckedFlatMapProof.correct(getMapEntryHash(entry), entries, missingKeys);
```
dmitry-timofeev: `getMapEntryHash(entry)` does not seem correct here: we must include a dbkey and a hash of value. I’ve fixed this in #456

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedMapProof.java
```diff
@@ -0,0 +1,42 @@
...
+ * It's used to get a checked map proof.
+ * Example usage:
+ * <pre><code>
+ * byte[] key = "The key for which I want a proved value";
+ * HashCode expectedRootHash = // get a known root hash from block proof //
+ * UncheckedMapProof proof = requestProofForKey(key);
+ * // Convert to checked
+ * CheckedMapProof checkedProof = proof.check();
+ * // Check the root hash
+ * if (compareWithRootHash(HashCode expectedRootHash)) {
```
dmitry-timofeev: This code is not accurate, `compareWithRootHash` does not seem to be a valid Java call.

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/CheckedMapProof.java
```diff
@@ -0,0 +1,83 @@
...
+ * throw IllegalStateException.
+ * Example usage:
+ * <pre><code>
+ * byte[] key = "The key for which I want a proved value";
+ * HashCode expectedRootHash = // get a known root hash from block proof //
+ * UncheckedMapProof proof = requestProofForKey(key);
+ * // Convert to checked
+ * CheckedMapProof checkedProof = proof.check();
+ * // Check the root hash
+ * if (checkedProof.compareWithRootHash(HashCode expectedRootHash)) {
```
dmitry-timofeev: `checkedProof.compareWithRootHash(HashCode expectedRootHash)` does not work in Java :-)

dmitry-timofeev: `checkedProof.compareWithRootHash(expectedRootHash)`?

MakarovS: sorry, missed that)

MakarovS:
@dmitry-timofeev Thanks for your help! Sorry it took so long.
Spikes for design tasks are a great idea that could save quite some time in this case, I'm glad we are doing those now.

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -0,0 +1,269 @@
...
+      }
+    }
+    return ProofStatus.CORRECT;
+  }
+
+  /**
+   * Check if any entry has a prefix among the paths in the proof entries. Both found and absent
+   * keys are checked.
+   */
+  private boolean prefixesIncluded() {
```
slowli: I'm not sure this is the complete check; we should also check that there are no prefixes of _proof keys_ in other proof keys, too. In the Rust counterpart [this check is performed](https://github.com/exonum/exonum/blob/35617bda142b2d44bf8ea8de753403534e7a9ab3/exonum/src/storage/proof_map_index/proof.rs#L489-L496) concurrently with checking ordering of proof keys. (Technically speaking, skipping this check can lead to an exception when calculating the index Merkle root hash - at least it does in the Rust version - but it may not be desired behavior.)

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProofTest.java
```diff
@@ -0,0 +1,254 @@
...
+    UncheckedMapProof uncheckedFlatMapProof =
+        new UncheckedFlatMapProof(
+            emptyList(), emptyList(), emptyList());
+
+    CheckedMapProof checkedMapProof = uncheckedFlatMapProof.check();
+    assertThat(checkedMapProof.getStatus(), equalTo(ProofStatus.CORRECT));
+  }
+
+  @Test
+  public void mapProofWithAbsentKeyShouldBeCorrect() {
```
slowli: Nit: there's "valid" / "correct" discrepancy among test naming.

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProofTest.java
```diff
@@ -0,0 +1,254 @@
...
+            emptyList());
+
+    CheckedMapProof checkedMapProof = uncheckedFlatMapProof.check();
+    assertThat(checkedMapProof.getStatus(), equalTo(ProofStatus.INVALID_ORDER));
+  }
+
+  @Test
+  public void mapProofWithSingleBranchProofEntryShouldBeInvalid() {
+    DbKey firstDbKey = DbKeyTestUtils.branchKeyFromPrefix("1011111");
+    byte[] absentKey = DbKeyTestUtils.keyFromString("101111");
```
slowli: I'm not completely sure, but allowing to specify missing keys as arbitrary byte sequences (instead of only "full" keys corresponding to leaves) looks weird. OTOH, I suppose there are no reasons for it to not work, provided that the length of the key is checked to be less or equal to the length of a leaf key.

dmitry-timofeev: As I remember, `keyFromString` must create a proper 32-byte key, that has the given _bit_ prefix (and an all-zero tail). It used to be a private method which name made perfect sense in that context — but in a wider context it might be confusing.

I am not sure though that we check the size of keys explicitly ­— but we certainly do _implicitly_ here (an exception will occur if it is not 32-byte long):
https://github.com/exonum/exonum-java-binding/pull/250/files/02985fb9adabfd387830e8e4dbdcd78e365dc26f#diff-2bbe03c0a0570e463c254c07245b4bd3R123

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/ProofStatus.java
```diff
@@ -24,7 +24,8 @@
   NON_TERMINAL_NODE("Proof entry in a singleton proof is of branch type (must be a leaf)"),
   INVALID_STRUCTURE("Proof has invalid structure"),
   INVALID_ORDER("Proof entries are placed in the wrong order"),
-  DUPLICATE_PATH("There are entries with duplicate keys");
+  DUPLICATE_PATH("There are entries with duplicate keys"),
+  EMBEDDED_PATH("Some keys in a proof are prefixes of others");
```
dmitry-timofeev: Shall we separate these two errors:
* A key in the proof is a prefix of the key of another proof entry.
* A key in the proof is a prefix of a requested key.

Combining them makes a difficult-to-understand description.

BTW, `INVALID_STRUCTURE` is no longer used.

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -0,0 +1,271 @@
...
+   *         {@code ProofStatus.INVALID_ORDER} if any following key key is lesser than the previous
+   *         {@code ProofStatus.DUPLICATE_PATH} if there are two equal keys
+   * @see DbKey#compareTo(DbKey)
+   */
+  private ProofStatus orderCheck() {
+    for (int i = 1; i < proof.size(); i++) {
+      DbKey key = proof.get(i - 1).getDbKey();
+      DbKey nextKey = proof.get(i).getDbKey();
+      int comparisonResult = nextKey.compareTo(key);
+      if (comparisonResult < 0) {
```
dmitry-timofeev: If we compare the keys and check for prefixes in the same order, does it make the branches clearer:
```java
      int comparisonResult = key.compareTo(nextKey);
      if (comparisonResult < 0) {
        if (key.isPrefixOf(nextKey)) {
          return ProofStatus.EMBEDDED_PATH;
        }
      } else if (comparisonResult == 0) {
        return ProofStatus.DUPLICATE_PATH;
      } else {
        return ProofStatus.INVALID_ORDER;
      }
```

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/ProofStatus.java
MakarovS: I don't think we should - these are both cases of the same error, they don't differ in any significant way. In rust it's the same `EmbeddedPaths` error.
I could change the description to the one we use in rust - `One path in the proof is a prefix of another path.` - would that be better?

dmitry-timofeev: They differ in the source of a key that has a prefix in `proofs` (and the call stack) — but we can use a more generic description.

PR #306, opened 2018-06-18
ECR-1708 Change configuration in deploy plugins and add a badge

MakarovS:
## Overview

Some changes to deploy configuration - now we explicitly state what artefacts we want to deploy, rather than stating what artefacts we want to skip.
Also add configuration to `maven-gpg-plugin` so that it works with newer versions of gpg - https://myshittycode.com/2017/08/07/maven-gpg-plugin-prevent-signing-prompt-or-gpg-signing-failed-no-such-file-or-directory-error/ (2.2.8 in our case).
And add a badge :)

---
See: https://jira.bf.local/browse/ECR-1708

### Definition of Done


exonum-java-binding-core/pom.xml
```diff
@@ -203,6 +203,14 @@
           <excludeFilterFile>findbugs-exclude.xml</excludeFilterFile>
         </configuration>
       </plugin>
+
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-deploy-plugin</artifactId>
+        <configuration>
+          <skip>false</skip>
```
dmitry-timofeev: Do we need 'skip' in configurations of the modules we deploy?

pom.xml
```diff
@@ -209,7 +209,6 @@
         <executions>
           <execution>
             <id>attach-javadocs</id>
-            <phase>deploy</phase>
```
dmitry-timofeev: You removed this because it's the default, correct?

MakarovS: No, both javadoc and sources plugins bind by default to `package` phase and that's how we want it.

dmitry-timofeev: Does this configuration make the build always generate Javadocs at 'package' phase?

MakarovS: > Does this configuration make the build always generate Javadocs at 'package' phase for every module?

True! I changed that so now we need to explicitly specify for which project we want to generate sources and javadoc.
We could also move those two plugins to `deploy` profile (where we have maven-gpg-plugin) if we don't need them unless we deploy.
But I think we could leave it be as we will probably change our deployment process (using maven release plugin).

pom.xml
```diff
@@ -204,12 +204,12 @@
         <artifactId>maven-javadoc-plugin</artifactId>
         <version>3.0.1</version>
         <configuration>
+          <skip>true</skip>
```
dmitry-timofeev: Why are `maven-javadoc-plugin`, `maven-source-plugin` and `deploy` here and not in `pluginManagement`, if they are not needed in each module?

exonum-java-binding-core/pom.xml
```diff
@@ -275,6 +291,13 @@
       <scope>provided</scope>
     </dependency>
 
+    <!--It is needed for Javadoc generation with JDK 9-->
+    <dependency>
+      <groupId>javax.annotation</groupId>
+      <artifactId>javax.annotation-api</artifactId>
+      <version>${javax-annotation-api.version}</version>
```
dmitry-timofeev: Is it needed in the default scope? Will `provided` scope fix the problem, so it is not packaged anywhere?

Asking because it is GPL, we cannot include it (but can migrate probably to better alternatives).

MakarovS:
> Would yo add please a jira task to skip building javadocs in the dev environment?

Sure. My vpn doesn't work at the moment so I can't access jira, but I will.

PR #314, opened 2018-06-20
Add modify_runpath script [ECR-1660]

vitvakatu:
## Overview
**DO NOT MERGE**

---
See: https://jira.bf.local/browse/ECR-1660

### DIFF
--- a/exonum-java-binding-core/rust/ejb-app/modify_runpath.sh
+++ b/exonum-java-binding-core/rust/ejb-app/modify_runpath.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+# Modifies RUNPATH of ejb-app executable to make it able to locate java_binding.so and Rust std lib .so
+# EJB App executable must be in the current directory, dynamic libraries in the `lib` subdirectory.
+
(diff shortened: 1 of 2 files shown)


exonum-java-binding-core/rust/ejb-app/modify_runpath.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+# Modifies RUNPATH of ejb-app executable to make it able to locate java_bindings.so and Rust std lib .so
+# EJB App executable must be in the current directory, dynamic libraries in the `lib` subdirectory.
+
+chrpath -v || { echo "No chrpath installed. Aborting"; exit 1; }
+
+# Convert RPATH to RUNPATH
```
dmitry-timofeev: We have to do that because the app also needs libjvm on LD_LIBRARY_PATH, which is traversed by the loader _before_ runpath, but after rpath, don't we?

vitvakatu: Rpath is deprecated, moreover user can change location of libraries usong LD_LIBRARY_PATH

PR #294, opened 2018-06-13
ECR-1551 Add read request to the cryptocurrency service API

MakarovS:
## Overview

Add GET request to the cryptocurrency service API and refactor ApiController to some extent, similar to QaService controller.

---
See: https://jira.bf.local/browse/ECR-1551

### Definition of Done

- [x] There are no TODOs left in the code
- [x] Change is covered by automated tests
- [x] The [coding guidelines](https://google.github.io/styleguide/javaguide.html) are followed
- [x] Public API has Javadoc
- [x] Method preconditions are checked and documented in the Javadoc of the method
- [ ] The [continuous integration build](https://www.travis-ci.com/exonum/exonum-java-binding) passes

PR #315, opened 2018-06-20
ECR-1703 Add text/plain content-type to string replies

MakarovS:
## Overview

Add text/plain content-type to string replies.

---
See: https://jira.bf.local/browse/ECR-1703

### Definition of Done


MakarovS:
I've added it here https://github.com/exonum/exonum-java-binding/pull/294

PR #318, opened 2018-06-20
Eliminate excessive memory allocations [ECR-1506].

alexander-irbis:
## Overview

Removed intermediate memory allocations of small arrays.

---
See: https://jira.bf.local/browse/ECR-1506

### Definition of Done

### DIFF
--- a/exonum-java-binding-core/rust/src/storage/proof_map_index.rs
+++ b/exonum-java-binding-core/rust/src/storage/proof_map_index.rs
@@ -12,20 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use exonum::storage::{Snapshot, Fork, ProofMapIndex};
-use exonum::storage::proof_map_index::{ProofMapIndexIter, ProofMapIndexKeys, ProofMapIndexValues,
-                                       PROOF_MAP_KEY_SIZE};
-use jni::JNIEnv;
+use exonum::storage::proof_map_index::{
+    ProofMapIndexIter, ProofMapIndexKeys, ProofMapIndexValues, PROOF_MAP_KEY_SIZE,
+};
+use exonum::storage::{Fork, ProofMapIndex, Snapshot};
 use jni::objects::{JClass, JObject, JString};
 use jni::sys::{jboolean, jbyte, jbyteArray, jint, jobject};
+use jni::JNIEnv;
 
 use std::mem;
 use std::panic;
 use std::ptr;
 
-use JniResult;
-use storage::db::{View, ViewRef, Value};
+use storage::db::{Value, View, ViewRef};
 use utils::{self, Handle, PairIter};
+use JniResult;
 
 type Key = [u8; PROOF_MAP_KEY_SIZE];
 type Index<T> = ProofMapIndex<T, Key, Value>;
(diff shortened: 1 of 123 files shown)


exonum-java-binding-core/rust/src/storage/proof_map_index.rs
```diff
@@ -442,11 +442,10 @@ pub extern "system" fn Java_com_exonum_binding_storage_indices_ProofMapIndexProx
...
-    assert_eq!(PROOF_MAP_KEY_SIZE, bytes.len());
+    let array_length = env.get_array_length(array)?;
+    assert_eq!(PROOF_MAP_KEY_SIZE as jint, array_length);
 
-    let mut key = Key::default();
-    key.copy_from_slice(&bytes);
-    Ok(key)
+    let mut key = [0 as jbyte; PROOF_MAP_KEY_SIZE];
+    env.get_byte_array_region(array, 0, &mut key[..])?;
+    Ok(unsafe { ::std::mem::transmute(key) })
```
vitvakatu: please don't use `::` at the beginning.

exonum-java-binding-core/rust/src/storage/proof_map_index.rs
```diff
@@ -442,11 +442,10 @@ pub extern "system" fn Java_com_exonum_binding_storage_indices_ProofMapIndexProx
 }
 
 fn convert_to_key(env: &JNIEnv, array: jbyteArray) -> JniResult<Key> {
-    // TODO: Optimize copying and allocations.
-    let bytes = env.convert_byte_array(array)?;
-    assert_eq!(PROOF_MAP_KEY_SIZE, bytes.len());
+    let array_length = env.get_array_length(array)?;
+    assert_eq!(PROOF_MAP_KEY_SIZE as jint, array_length);
```
vitvakatu: Shouldn't we return error instead of panic?

alexander-irbis: A panic will be intercepted and converted into a Java Error.
I suppose, if here is a Java array of a wrong length, things are totally broken and the application should be terminated.

exonum-java-binding-core/rust/src/utils/conversion.rs
```diff
@@ -12,20 +12,21 @@
...
-    let bytes = env.convert_byte_array(array)?;
-    Ok(Hash::from_slice(&bytes).expect(
-        "Unable to create `Hash` from the slice",
-    ))
+    let array_length = env.get_array_length(array)?;
+    assert_eq!(HASH_SIZE as jint, array_length);
+
+    let mut bytes = [0 as jbyte; HASH_SIZE];
+    env.get_byte_array_region(array, 0, &mut bytes[..])?;
+    Ok(Hash::from(unsafe { ::std::mem::transmute(bytes) }))
```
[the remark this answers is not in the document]
alexander-irbis: The additional JNI call was hidden under the hood of jni-rs before.
Only thing was removed here is a redundant memory allocation with Vec.
But this `unsafe` unfortunately currently is the only way to convert an array of i8 into an array of u8 in Rust. Hopefully in the future this will be improved.

alexander-irbis:
It's decided not to add `unsafe`.
Complication of the code is not proportional to the benefit.

PR #322, opened 2018-06-24
Use latest rustfmt-preview [ECR-1158]

vitvakatu:
See: https://jira.bf.local/browse/ECR-1158

### Definition of Done


vitvakatu:
I think yes, we can fix any PR easily.

PR #328, opened 2018-06-25
ECR-1367 Import Guava tests

MakarovS:
## Overview

Imported Guava tests with almost no changes.

---
See: https://jira.bf.local/browse/ECR-1367

### Definition of Done

- [x] There are no my TODOs left in the code
- [x] Change is covered by automated [tests](../CONTRIBUTING.md#tests)
- [x] The [coding guidelines](../CONTRIBUTING.md#the-code-style) are followed
- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-proofs/src/main/java/com/exonum/binding/hash/AbstractNonStreamingHashFunction.java
+++ b/exonum-java-proofs/src/main/java/com/exonum/binding/hash/AbstractNonStreamingHashFunction.java
@@ -1,0 +1,148 @@
+/*
+ * Copyright (C) 2011 The Guava Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+/*
+ * Copyright 2018 The Exonum Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.exonum.binding.hash;
+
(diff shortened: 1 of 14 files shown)


dmitry-timofeev:
Which version do these files correspond to?

exonum-java-proofs/src/test/java/com/exonum/binding/hash/HashingTest.java
```diff
@@ -1,4 +1,19 @@
-/* 
+/*
```
dmitry-timofeev: Probably, ours is first.

MakarovS:
> Which version do these files correspond to?

`25.1-jre` - the one we use.

exonum-java-proofs/src/main/java/com/exonum/binding/hash/AbstractNonStreamingHashFunction.java
```diff
@@ -0,0 +1,148 @@
...
+
+/**
+ * Skeleton implementation of {@link HashFunction}, appropriate for non-streaming algorithms. All
+ * the hash computation done using {@linkplain #newHasher()} are delegated to the {@linkplain
+ * #hashBytes(byte[], int, int)} method.
+ *
+ * @author Dimitris Andreou
+ */
+@Immutable
+abstract class AbstractNonStreamingHashFunction extends AbstractHashFunction {
```
dmitry-timofeev: Is this (and the second below) class required to test some of the functionality we already have?

MakarovS: Yes, we need that for `AbstractStreamingHasher`, which in turn we need for `FunnelsTest`.

MakarovS:
> These classes use JUnit 3 APIs :)

Do you suggest we refactor those? I tried to keep any changes to minimum.

exonum-java-proofs/src/test/java/com/exonum/binding/hash/HashingTest.java
MakarovS: In other guava classes we have Guava license first, ours second. Do you think we should change it there as well?

exonum-java-proofs/src/main/java/com/exonum/binding/hash/AbstractNonStreamingHashFunction.java
dmitry-timofeev: I see, do you think it makes sense to put these files in `test/` folder, or keep here in `main/` to be consistent with the source of this package?

dmitry-timofeev:
> Do you suggest we refactor those? I tried to keep any changes to minimum.

No, certainly not. We may consider doing that system-wide when will migrate to JUnit 5 in the future.

exonum-java-proofs/src/main/java/com/exonum/binding/hash/AbstractNonStreamingHashFunction.java
MakarovS: I think we should keep those in main for consistency with original Guava lib. Plus someone might want to use those outside of tests.

dmitry-timofeev: OK, the latter is unlikely for they are package-private.

PR #330, opened 2018-06-26
Add native support for flat proofs [ECR-1350]

vitvakatu:
## Overview

Native getProof method.
Does not support proof of absence. 
Does not includes tests (yet)
Formatting and code readability is awful (yet)

---
See: https://jira.bf.local/browse/XYZ

### Definition of Done

- [ ] Change is covered by automated [tests](../CONTRIBUTING.md#tests)
- [x] The [coding guidelines](../CONTRIBUTING.md#the-code-style) are followed
- [x] Public API has Javadoc
- [ ] Add unit tests
- [x] Polish code

### DIFF
--- a/exonum-java-binding-core/rust/src/storage/proof_map_index.rs
+++ b/exonum-java-binding-core/rust/src/storage/proof_map_index.rs
@@ -12,10 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use exonum::storage::proof_map_index::{
-    ProofMapIndexIter, ProofMapIndexKeys, ProofMapIndexValues, PROOF_MAP_KEY_SIZE,
-};
-use exonum::storage::{Fork, ProofMapIndex, Snapshot};
 use jni::objects::{JClass, JObject, JString};
 use jni::sys::{jboolean, jbyteArray, jobject};
 use jni::JNIEnv;
@@ -23,12 +19,27 @@ use jni::JNIEnv;
 use std::panic;
 use std::ptr;
 
+use exonum::crypto::Hash;
+use exonum::storage::proof_map_index::{
+    ProofMapIndexIter, ProofMapIndexKeys, ProofMapIndexValues, ProofPath, PROOF_MAP_KEY_SIZE,
+};
+use exonum::storage::{Fork, ProofMapIndex, Snapshot, StorageValue};
+
 use storage::db::{Value, View, ViewRef};
 use utils::{self, Handle, PairIter};
 use JniResult;
 
 type Key = [u8; PROOF_MAP_KEY_SIZE];
 type Index<T> = ProofMapIndex<T, Key, Value>;
+
+const MAP_PROOF_ENTRY_BRANCH: &str =
+    "com/exonum/binding/storage/proofs/map/flat/MapProofEntryBranch";
+const MAP_PROOF_ENTRY_LEAF: &str = "com/exonum/binding/storage/proofs/map/flat/MapProofEntryLeaf";
+const MAP_PROOF_ENTRY: &str = "com/exonum/binding/storage/proofs/map/flat/MapProofEntry";
+const UNCHECKED_FLAT_MAP_PROOF: &str =
+    "com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof";
+const UNCHECKED_FLAT_MAP_PROOF_SIG: &str =
+    "([Lcom/exonum/binding/storage/proofs/map/flat/MapProofEntry;)Lcom/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof;";
 
 enum IndexType {
     SnapshotIndex(Index<&'static Snapshot>),
@@ -163,14 +174,93 @@ pub extern "system" fn Java_com_exonum_binding_storage_indices_ProofMapIndexProx
 ) -> jobject {
     let res = panic::catch_unwind(|| {
         let key = convert_to_key(&env, key)?;
-        env.ensure_local_capacity(512)?;
-        let _proof = match *utils::cast_handle::<IndexType>(map_handle) {
+        let proof = match *utils::cast_handle::<IndexType>(map_handle) {
             IndexType::SnapshotIndex(ref map) => map.get_proof(key),
             IndexType::ForkIndex(ref map) => map.get_proof(key),
         };
-        unimplemented!("ECR-1350")
+
+        let unchecked_entries = proof.proof_unchecked();
+        let branches: Vec<_> = unchecked_entries
+            .into_iter()
+            .map(|(path, hash)| create_java_map_proof_entry_branch(&env, &path, &hash))
+            .collect::<JniResult<_>>()?;
+        let local_refs = 100 + 2 * branches.len();
+        env.ensure_local_capacity(local_refs as i32)?;
+
+        // TODO: avoid checking proofs (ECR-1802)
+        let checked_proof = proof.check().unwrap();
+        let leaves: Vec<_> = checked_proof
+            .all_entries()
+            .into_iter()
+            .map(|(key, optional_value)| {
+                if optional_value.is_none() {
+                    unimplemented!("Proofs for missing keys are not yet supported");
+                }
+                let path = ProofPath::new(key);
+                let value: Vec<u8> = optional_value.cloned().unwrap().into_bytes();
+                create_java_map_proof_entry_leaf(&env, &path, &value)
+            })
+            .collect::<JniResult<_>>()?;
+        let array = create_java_array_map_proof_entry(&env, &leaves, &branches)?;
+        let unchecked_flat_map_proof = create_java_unchecked_flat_map_proof(&env, array)?;
+        Ok(unchecked_flat_map_proof.into_inner())
     });
     utils::unwrap_exc_or(&env, res, ptr::null_mut())
+}
+
+fn create_java_map_proof_entry_branch<'a>(
+    env: &'a JNIEnv,
+    proof_path: &ProofPath,
+    hash: &Hash,
+) -> JniResult<JObject<'a>> {
+    let proof_path: JObject = env.byte_array_from_slice(proof_path.as_bytes())?.into();
+    let hash: JObject = utils::convert_hash(env, hash)?.into();
+    env.new_object(
+        MAP_PROOF_ENTRY_BRANCH,
+        "([B[B)V",
+        &[proof_path.into(), hash.into()],
+    )
+}
+
+fn create_java_map_proof_entry_leaf<'a>(
+    env: &'a JNIEnv,
+    proof_path: &ProofPath,
+    value: &[u8],
+) -> JniResult<JObject<'a>> {
+    let proof_path: JObject = env.byte_array_from_slice(proof_path.as_bytes())?.into();
+    let value: JObject = env.byte_array_from_slice(value)?.into();
+    env.new_object(
+        MAP_PROOF_ENTRY_LEAF,
+        "([B[B)V",
+        &[proof_path.into(), value.into()],
+    )
+}
+
+fn create_java_array_map_proof_entry<'a, 'b: 'a>(
+    env: &'a JNIEnv,
+    leaves: &'b [JObject],
+    branches: &'b [JObject],
+) -> JniResult<JObject<'a>> {
+    let length = leaves.len() + branches.len();
+    let element_class = MAP_PROOF_ENTRY;
+    let array = env.new_object_array(length as i32, element_class, JObject::null())?;
+    for (index, entity) in branches.iter().chain(leaves.iter()).enumerate() {
+        env.set_object_array_element(array, index as i32, *entity)?;
+    }
+    Ok(array.into())
+}
+
+fn create_java_unchecked_flat_map_proof<'a>(
+    env: &'a JNIEnv,
+    entries: JObject,
+) -> JniResult<JObject<'a>> {
+    let java_proof = env.call_static_method(
+        UNCHECKED_FLAT_MAP_PROOF,
+        "fromUnsorted",
+        UNCHECKED_FLAT_MAP_PROOF_SIG,
+        &[entries.into()],
+    )?;
+    java_proof.l()
 }
 
 /// Returns the pointer to the iterator over a map keys and values.


exonum-java-binding-core/rust/src/storage/proof_map_index.rs
```diff
@@ -163,15 +173,75 @@ pub extern "system" fn Java_com_exonum_binding_storage_indices_ProofMapIndexProx
...
+        let proof = match *utils::cast_handle::<IndexType>(map_handle) {
             IndexType::SnapshotIndex(ref map) => map.get_proof(key),
             IndexType::ForkIndex(ref map) => map.get_proof(key),
         };
-        unimplemented!("ECR-1350")
+        let unchecked_entries = proof.proof_unchecked();
+        let branches: Vec<_> = unchecked_entries.into_iter()
+            .map(|(path, hash)| create_java_map_proof_entry_branch(&env, path, hash))
+            .collect::<Result<Vec<_>>>()?;
+        let checked_proof = proof.check().unwrap();
```
dmitry-timofeev: I'd probably add a comment explaining that checking a proof is a temporary measure.

exonum-java-binding-core/rust/src/storage/proof_map_index.rs
```diff
@@ -163,15 +173,75 @@ pub extern "system" fn Java_com_exonum_binding_storage_indices_ProofMapIndexProx
...
+        &[proof_path.into(), value.into()]
+    )
+}
+
+fn create_java_array_map_proof_entry<'a, 'b: 'a>(env: &'a JNIEnv, leaves: &'b [JObject], branches: &'b [JObject]) -> Result<JObject<'a>> {
+    let length = leaves.len() + branches.len();
+    let element_class = MAP_PROOF_ENTRY;
+    let array = env.new_object_array(length as i32, element_class, JObject::null())?;
+    let mut index = 0;
+    for branch in branches {
```
dmitry-timofeev: @MakarovS , any ordering will do, is that correct?

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -7,11 +7,8 @@
 import com.exonum.binding.hash.HashFunction;
 import com.exonum.binding.hash.Hashing;
 import com.exonum.binding.storage.proofs.map.DbKey;
-import java.util.ArrayDeque;
-import java.util.ArrayList;
-import java.util.Deque;
-import java.util.List;
-import java.util.Optional;
+
+import java.util.*;
```
dmitry-timofeev: Wildcard imports are not allowed — style checks will fail.

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -22,8 +19,8 @@
 
   private final List<MapProofEntry> proofList;
 
-  UncheckedFlatMapProof(List<MapProofEntry> proofList) {
-    this.proofList = proofList;
+  UncheckedFlatMapProof(MapProofEntry[] proofList) {
```
dmitry-timofeev: Please add
```java
@SuppressWarnings("unused") // Native API
```

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -22,8 +19,8 @@
...
+    this.proofList = Arrays.asList(proofList);
```
dmitry-timofeev: @MakarovS any preconditions to check? E.g., that it is non-empty?

MakarovS: Right now if we receive an empty `proofList` we consider it a correct one. Is it correct?

exonum-java-binding-core/rust/src/storage/proof_map_index.rs
MakarovS: We need specific ordering, like stated in the algorithm here https://wiki.bf.local/display/EXN/Flat+proofs+for+ProofMapIndex
If we look at the proof from the example there, we need `H_1, V_1, H_2, H_3, H_4` ordering.

vitvakatu: thats not good tbh

dmitry-timofeev: I think there is a misunderstanding: the algorithm checking the proof sorts the entries as its  first step, if I remember correctly. If that’s correct, we may use any order here. However, there might be some *preferred* order.

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
dmitry-timofeev: It did the same to me in Rust with its optimize imports feature 🙂

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
dmitry-timofeev: Can’t say, got to see the spec or the reference implementation 🤷‍♂️

exonum-java-binding-core/rust/src/storage/proof_map_index.rs
vitvakatu: Well, we can sort them either on Java or on Rust side. What do you choose? I think Rust would have better performance here.

dmitry-timofeev: That’s a good question. Unless we say that a class representing a proof must get a sorted list of entries (i.e., it’s a precondition), it will have to always sort the input. I’m not sure we shall impose such a precondition - it might make unit testing a little more complex as we will have to always provide a sorted list of entries. Sorting of a sorted array will definitely be faster, but the compound operation (create and check) might not be.

MakarovS: Hm, I expected proof to be passed to Java side already sorted and we check that it is.

exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/flat/UncheckedFlatMapProof.java
```diff
@@ -22,8 +25,11 @@
 
   private final List<MapProofEntry> proofList;
 
-  UncheckedFlatMapProof(List<MapProofEntry> proofList) {
-    this.proofList = proofList;
+  @SuppressWarnings("unused") // Native API
+  UncheckedFlatMapProof(MapProofEntry[] proofList) {
```
dmitry-timofeev: I'd better keep a constructor accepting a `List` of entries and not doing anything with them; and add a new one accepting an array of entries, sorting them, and creating an `UncheckedFlatMapProof`. E.g.:
```java
@SuppressWarnings("unused") // Native API
static UncheckedFlatMapProof fromUnsorted(MapProofEntry[] proofEntries) {
  // do your job
  return new UncheckedFlatMapProof(sortedProofEntries);
}
```
or, if you prefer inlining, you may add a method that sorts an array and returns it as a list, and just call
```
this(sort(proofEntries)));
```

exonum-java-proofs/src/test/java/com/exonum/binding/storage/proofs/map/flat/CheckedFlatMapProofTest.java
```diff
@@ -30,11 +28,11 @@ public void mapProofShouldBeValid() {
 
     DbKey valueKey = DbKey.newLeafKey(secondKey);
     MapProofEntryLeaf leaf = createLeafMapEntry(valueKey.getKeySlice(), VALUE);
-    List<MapProofEntry> entries =
-        Arrays.asList(
-            createBranchMapEntry(firstKey),
-            leaf,
-            createBranchMapEntry(thirdKey));
+    MapProofEntry[] entries = {
```
dmitry-timofeev: I'd keep using lists here — (1) they are more convenient; (2) tests might not need any sorting.

exonum-java-binding-core/rust/src/storage/proof_map_index.rs
```diff
@@ -163,15 +174,91 @@ pub extern "system" fn Java_com_exonum_binding_storage_indices_ProofMapIndexProx
     let res = panic::catch_unwind(|| {
         let key = convert_to_key(&env, key)?;
         env.ensure_local_capacity(512)?;
```
dmitry-timofeev: It probably has to be _f(num(branches) + num(leaves))_? Or we have to delete local refs as we add them to the _java_ array (i.e., do not store them in an intermediate vector.

dmitry-timofeev:
> Decided to do sorting on Java side due to simplicity comparing to possible Rust implementation.

And I was just curious how you would do that in Rust :upside_down_face:

exonum-java-binding-core/rust/src/storage/proof_map_index.rs
```diff
@@ -162,16 +173,92 @@ pub extern "system" fn Java_com_exonum_binding_storage_indices_ProofMapIndexProx
 ) -> jobject {
     let res = panic::catch_unwind(|| {
         let key = convert_to_key(&env, key)?;
-        env.ensure_local_capacity(512)?;
```
dmitry-timofeev: You must either call this or delete-as-you-go. Just have to pick the right size. I'd use _2 * (num(branches) + num(leaves))_

vitvakatu: :thinking: But we don't know the count of branches and leaves at this moment. What about using `auto_local`?

dmitry-timofeev: Yes, it will complicate thing. We can get the number of branches just below, so let's apply a simple estimate of _100 + 2 * num(branches)_ (you won't request a single proof for a hundred keys, will you?).

slowli:
Guys, I'm out of context here, so my comment may possibly be not productive, but still: all proofs produced by the core (i.e., `ProofMapIndex::get_proof` and `ProofMapIndex::get_multiproof`) *are* sorted. Non-sorted proofs can only be obtained from external sources.

dmitry-timofeev:
@slowli, thanks, that's good news.

@MakarovS , as you keep leaves and branches together, is that correct that there still a need to merge the two types of nodes into a single sorted list?

exonum-java-binding-core/rust/src/storage/proof_map_index.rs
```diff
@@ -251,12 +254,13 @@ fn create_java_unchecked_flat_map_proof<'a>(
     env: &'a JNIEnv,
     entries: JObject,
 ) -> JniResult<JObject<'a>> {
-    env.call_static_method(
+    let jvalue = env.call_static_method(
```
dmitry-timofeev: `proof` or `java_proof` or inline — `jvalue` is too generic.

exonum-java-binding-core/rust/src/utils/mod.rs
```diff
@@ -24,8 +24,10 @@ mod pair_iter;
 mod resource_manager;
 
 pub use self::conversion::{convert_hash, convert_to_hash, convert_to_string};
-pub use self::errors::{check_error_on_exception, get_and_clear_java_exception, panic_on_exception,
-                       unwrap_jni, unwrap_jni_verbose};
+pub use self::errors::{
+    check_error_on_exception, get_and_clear_java_exception, panic_on_exception, unwrap_jni,
```
dmitry-timofeev: Why do these changes appear in every PR? Haven't we merged them?

vitvakatu: We updated rustfmt to the latest version, but a week later next version was released which slightly changed syntax for imports. I suppose no commit with fix were merged yet

dmitry-timofeev:
> What merge do you mean?

In native there are two apparently sorted lists — one of structures corresponding to branches and one to leaves. But in Java we keep both types in a single list, which, as I understand, must also be sorted. Hence a merge operation :-)

dmitry-timofeev:
Why don’t we integrate this into master so that we don’t have to review it again with flat proofs? It will remain unused until the corresponding native methods are added to the map.

dmitry-timofeev:
> Why don’t we integrate this into master so that we don’t have to review it again with flat proofs? It will remain unused until the corresponding native methods are added to the map.

Not that easy — got to take out changes to Java and apply them in the ECR-1164 branch.

MakarovS:
> In native there are two apparently sorted lists — one of structures corresponding to branches and one to leaves. But in Java we keep both types in a single list, which, as I understand, must also be sorted. Hence a merge operation :-)

I thought that is what happens here https://github.com/exonum/exonum-java-binding/pull/330/files#diff-9e9e72b32ca730e1ad382d0fcbe47f26R247 no?

> Not that easy — got to take out changes to Java and apply them in the ECR-1164 branch.

I can add Java changes to ECR-1164.

dmitry-timofeev:
> I thought that is what happens here https://github.com/exonum/exonum-java-binding/pull/330/files#diff-9e9e72b32ca730e1ad382d0fcbe47f26R247 no?

That operation does not preserve a total order among all items. The order is restored in the Java side through sorting.

> integrate this into master 

Ilya, WDYT?

vitvakatu:
@dmitry-timofeev I don't see any problems - we can merge this PR now and work with java in a separate one.

dmitry-timofeev:
OK, then would you please revert changes to Java and send the patch to @MakarovS ?

vitvakatu:
@dmitry-timofeev should be ready.

dmitry-timofeev:
There are 16 files changed, must be a single file :upside_down_face:

vitvakatu:
:thinking: Indeed, my fault. Sorry, I was forced to do a rebase. I just copied a single file from one branch to another.

PR #336, opened 2018-07-03
Improve build on Mac (with SIP) [ECR-1793]

alexander-irbis:
## Overview

Replaces a usage of LD_LIBRARY_PATH in run time with the rpath linker option. 

---
See: https://jira.bf.local/browse/ECR-1793

### Definition of Done

- [x] Build instructions are updated


run_profile
```diff
@@ -0,0 +1,19 @@
...
+# hence this a bit complex thing.
+export JAVA_HOME="$(mvn --version | grep 'Java home' | sed 's/.*: //')"
+echo "JAVA_HOME=${JAVA_HOME}"
+
+# Find the directory containing libjvm (the relative path has changed in Java 9).
+JAVA_LIB_DIR="$(find ${JAVA_HOME} -type f -name libjvm.* | xargs -n1 dirname)"
+echo "JAVA_LIB_DIR=${JAVA_LIB_DIR}"
+
+# Modify path where an executable look for libjvm.
+export RUSTFLAGS="${RUSTFLAGS:-} -C link-args=-Wl,-rpath,${JAVA_LIB_DIR}"
```
vitvakatu: waaait, it means that user will need to compile App from sources on his machine which is unacceptable

alexander-irbis: Travis already compiles it, nothing changes here.
These scripts are used to build EJB for tests, whoever does it.

dmitry-timofeev: What are the effects, if any, on other artefacts that we build (ITs and App)?

run_profile
```diff
@@ -0,0 +1,19 @@
...
+# Unfortunately, a simple `which java` will not work for some users (e.g., jenv),
+# hence this a bit complex thing.
+export JAVA_HOME="$(mvn --version | grep 'Java home' | sed 's/.*: //')"
+echo "JAVA_HOME=${JAVA_HOME}"
+
+# Find the directory containing libjvm (the relative path has changed in Java 9).
+JAVA_LIB_DIR="$(find ${JAVA_HOME} -type f -name libjvm.* | xargs -n1 dirname)"
+echo "JAVA_LIB_DIR=${JAVA_LIB_DIR}"
+
+# Modify path where an executable look for libjvm.
```
dmitry-timofeev: Please document it in more detail, including any important implications of this solution.

run_all_tests.sh
```diff
@@ -7,6 +7,8 @@
 # Fail immediately in case of errors and/or unset variables
 set -eu -o pipefail
 
+source ./run_profile
```
dmitry-timofeev: Please document scripts thoroughly.

run_profile
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+# Environment for building and running tests with dynamically loading libraries.
```
dmitry-timofeev: Sets the environment variables required for … ?

run_native_integration_tests.sh
```diff
@@ -4,20 +4,9 @@
 #
 # ¡Keep it MacOS/Ubuntu compatible!
 
-# Fail immediately in case of errors and/or unset variables
+# Fail immediately in case of errors and/or unset variables.
 set -eu -o pipefail
 
-# Use the Java that Maven uses.
```
dmitry-timofeev: No vars required here? Can I run this script directly and see it work?

run_profile
dmitry-timofeev: We don't have a build script 'for tests'. If I understand this correctly, the final executable (or shared library, IDK) will have a path to some libjvm that was used on whatever system we build it. It might not be a problem if there is no JDK by this path on the running system, but will be if there is and it is not the one user wants to use.

alexander-irbis: I fear I don't understand what is "not other".
This affects all targets, which built with this script.
Any executable.

run_all_tests.sh
```diff
@@ -7,18 +7,9 @@
...
-# See build definitions of the modules for more.
-mvn install \
-  -DskipTests \
-  --activate-profiles ci-build \
-  -Drust.compiler.version="stable"
+# Run all java tests and native unit tests.
+./run_maven_tests.sh
 
-# Run native integration tests that require a JVM.
+# Run native integration tests that require prepared classpaths for fake classes.
```
dmitry-timofeev: require ejb-fakes module and/or JVM.

tests_profile
```diff
@@ -0,0 +1,47 @@
...
+# - JAVA_HOME - path to currently selected JDK.
+#   jni-rs with enabled the invocation feature fails to build without JAVA_HOME.
+# - JAVA_LIB_DIR - path to libjvm.
+# - RUST_LIB_DIR - path to Rust stdlib.
+# - RUST_VERSION - Rust version for building tests.
+# - RUSTFLAGS - Rust build options augmented with JAVA_LIB_DIR and RUST_LIB_DIR.
+#   That adds paths into search paths where native code at runtime will look for
+#   dynamically loadable libraries.
+#   On Linux it is RUNPATH, so LD_LIBRARY_PATH is examined before.
+#   On MacOS it extends @rpath, so LD_LIBRARY_PATH works as usually.
```
dmitry-timofeev: as usually — how? Can I override rpath or not?

tests_profile
```diff
@@ -0,0 +1,47 @@
...
+# ¡Keep it MacOS/Ubuntu compatible!
+#
+# This profile contains:
+# - JAVA_HOME - path to currently selected JDK.
+#   jni-rs with enabled the invocation feature fails to build without JAVA_HOME.
+# - JAVA_LIB_DIR - path to libjvm.
+# - RUST_LIB_DIR - path to Rust stdlib.
+# - RUST_VERSION - Rust version for building tests.
+# - RUSTFLAGS - Rust build options augmented with JAVA_LIB_DIR and RUST_LIB_DIR.
+#   That adds paths into search paths where native code at runtime will look for
```
dmitry-timofeev: absolute paths?

tests_profile
```diff
@@ -0,0 +1,47 @@
...
+#   dynamically loadable libraries.
+#   On Linux it is RUNPATH, so LD_LIBRARY_PATH is examined before.
+#   On MacOS it extends @rpath, so LD_LIBRARY_PATH works as usually.
+#   This affects all executables (application, tests, libjava_bindings).
+#   It worth nothing that recent releases of MacOS with enabled by default SIP
+#   clear LD_LIBRARY_PATH when new shell started.
+#   Such behavior greatly affects Maven and all that it runs.
```
dmitry-timofeev: It does not affect Maven, it makes it impossible to pass LD_LIBRARY_PATH to the application from the command interpreter.

tests_profile
```diff
@@ -0,0 +1,47 @@
...
+# - JAVA_LIB_DIR - path to libjvm.
+# - RUST_LIB_DIR - path to Rust stdlib.
+# - RUST_VERSION - Rust version for building tests.
+# - RUSTFLAGS - Rust build options augmented with JAVA_LIB_DIR and RUST_LIB_DIR.
+#   That adds paths into search paths where native code at runtime will look for
+#   dynamically loadable libraries.
+#   On Linux it is RUNPATH, so LD_LIBRARY_PATH is examined before.
+#   On MacOS it extends @rpath, so LD_LIBRARY_PATH works as usually.
+#   This affects all executables (application, tests, libjava_bindings).
+#   It worth nothing that recent releases of MacOS with enabled by default SIP
```
dmitry-timofeev: It is worth noting?

tests_profile
```diff
@@ -0,0 +1,47 @@
...
+# Version of Rust used to build tests.
+export RUST_VERSION="stable"
+echo "RUST_VERSION: ${RUST_VERSION}"
+
+# Find the directory containing Rust libstd.
+export RUST_LIB_DIR=$(rustup run ${RUST_VERSION} rustc --print sysroot)/lib
+echo "RUST_LIB_DIR: ${RUST_LIB_DIR}"
+
+# Linking options with additional paths for the application and tests.
+export RUSTFLAGS="${RUSTFLAGS:-} -C link-args=-Wl,-rpath,${RUST_LIB_DIR},-rpath,${JAVA_LIB_DIR}"
```
dmitry-timofeev: I don't think it is correct: you do not change RUSTFLAGS if it is set. But what if I change the JVM (interactively using jenv, or re-install to other directory, etc), or rust version, and run the script?

run_all_tests.sh
alexander-irbis: JVM we need not only for ITs now.

tests_profile
alexander-irbis: On MacOS a path can be relative to the current directory, relative to the current executable, relative to the calling binary (when loading dylib from other executable), relative to @rpath, and absolute.
In this case we add absolute paths into @rpath.
Lets make `RUSTFLAGS=... /wrong_path/Library/Java/JavaVirtualMachines/jdk1.8.0_171.jdk/Contents/Home/jre/lib/server`
```
cargo build --tests

target/debug/deps/integration_tests-1d8668cc32025316
dyld: Library not loaded: @rpath/libjava_bindings.dylib
...
LD_LIBRARY_PATH="target/debug/deps" target/debug/integration_tests-1d8668cc32025316
dyld: Library not loaded: @rpath/libjvm.dylib
...
LD_LIBRARY_PATH="target/debug/deps:/Library/Java/JavaVirtualMachines/jdk1.8.0_171.jdk/Contents/Home/jre/lib/server" target/debug/integration_tests-1d8668cc32025316

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Ok, let's look what is in load commands:
```
otool -l target/debug/integration_tests-1d8668cc32025316
...
Load command 14
          cmd LC_LOAD_DYLIB
      cmdsize 56
         name @rpath/libjava_bindings.dylib (offset 24)
   time stamp 2 Thu Jan  1 03:00:02 1970
      current version 0.0.0
compatibility version 0.0.0
...
Load command 16
          cmd LC_LOAD_DYLIB
      cmdsize 48
         name @rpath/libjvm.dylib (offset 24)
   time stamp 2 Thu Jan  1 03:00:02 1970
      current version 1.0.0
compatibility version 1.0.0
...
Load command 22
          cmd LC_RPATH
      cmdsize 96
         path /Users/alexandrsemichenkov/.rustup/toolchains/stable-x86_64-apple-darwin/lib (offset 12)
Load command 23
          cmd LC_RPATH
      cmdsize 96
         path /wrong_path/Library/Java/JavaVirtualMachines/jdk1.8.0_171.jdk/Contents/Home/jre/lib/server (offset 12)
```

tests_profile
alexander-irbis: It is possible to pass LD_LIBRARY_PATH to the application from the command interpreter.
An impossible thing is to pass LD_LIBRARY_PATH to the command interpreter.
Maven is running with the command interpreter.

tests_profile
alexander-irbis: I take into account existing value of RUSTFLAGS and add "-C link-args" to it.
But it will no affect a scoupe outer to that, where was called the "source" command.
After restart the script re-evaluates new JAVA_HOME and new JAVA_LIB_DIR.
Only conflict we can expect is a presense of the "-C link-args" option before tests_profile, it will be overrided by our one.
But correct parsing and modifiing of RUSTFLAGS will not be easy task.
Should we do it now, or should we allow a user with such nontrivial environment resolve conflict on his own?

run_all_tests.sh
dmitry-timofeev: Yes, but ITs require either one or both.

tests_profile
dmitry-timofeev: I see, I misunderstood the script.

> Only conflict we can expect is a presense of the "-C link-args" option before tests_profile, it will be overrided by our one.

Not appended to, but overridden? That's a good point, if that happens, we will get quite a weird linker error some thousand lines of output below. Can we error iff RUSTFLAGS is set _and_ contains "link-args=", so that we detect a prospective linker error ASAP?

run_all_tests.sh
alexander-irbis: All tests require JVM, it makes no sense to mention it separately here.

run_ejb_app_tests.sh
```diff
@@ -1,27 +1,30 @@
...
-# Find the directory containing libjvm (the relative path has changed in Java 9)
-export LD_LIBRARY_PATH="$(find ${JAVA_HOME} -type f -name libjvm.* | xargs -n1 dirname)"
-echo "LD_LIBRARY_PATH=${LD_LIBRARY_PATH}"
+# Compile all Java modules by default to ensure that ejb-fakes module, which is required
+# by native ITs, is up-to-date. This safety net takes about a dozen seconds,
+# so if the Java artefacts are definitely up-to-date, it may be skipped.
+if [ "$#" -eq 0 ]; then
+  # Compile Java artefacts.
+  echo "Compiling the Java artefacts…"
+  mvn -DskipTests compile --quiet
```
dmitry-timofeev: @vitvakatu , @alexander-irbis ,
Do ejb-app tests depend on any Java classes? If not, I believe this code is not needed.

tests_profile
```diff
@@ -0,0 +1,56 @@
...
+echo "JAVA_HOME=${JAVA_HOME}"
+
+# Find the directory containing libjvm (the relative path has changed in Java 9).
+export JAVA_LIB_DIR="$(find ${JAVA_HOME} -type f -name libjvm.\* | xargs -n1 dirname)"
+echo "JAVA_LIB_DIR=${JAVA_LIB_DIR}"
+
+# Version of Rust used to build tests.
+## Stable works well unless you want benchmarks.
+# TODO: stable does not work well until ECR-1839 is resolved
+export RUST_COMPILER_VERSION="${RUST_COMPILER_VERSION:-1.26.2}"
```
dmitry-timofeev: Shan't we update the .travis.yml to use the same variable?

dmitry-timofeev:
Will test locally, there might be some updates needed to the build instructions.

run_ejb_app_tests.sh
vitvakatu: AFAIK not

run_maven_tests.sh
```diff
@@ -0,0 +1,20 @@
...
+# Run unit and integration tests in ci-build profile. This profile includes:
+#  - Java unit & integration tests, including ci-only & slow non-critical tests,
+#    which are excluded in the default profile.
+#  - Checkstyle checks as errors.
+#  - Native unit & integration tests that do not require a JVM.
+# See build definitions of the modules for more.
+mvn install \
+  -DskipTests \
+  --activate-profiles ci-build \
+  -Drust.compiler.version="${RUST_COMPILER_VERSION}"
```
vitvakatu: There should be 1.26.2 I beleive

tests_profile
```diff
@@ -0,0 +1,69 @@
...
+
+# Find the directory containing Rust libstd.
+export RUST_LIB_DIR=$(rustup run ${RUST_COMPILER_VERSION} rustc --print sysroot)/lib
+echo "RUST_LIB_DIR: ${RUST_LIB_DIR}"
+
+# Linking options with additional paths for the application and tests.
+# This may override previous `-C link-args` if it appears in RUSTFLAGS before.
+RUSTFLAGS="${RUSTFLAGS:-}"
+if [[ $RUSTFLAGS = *"-C link-args="* ]]; then
+    echo "Error: the '-C link-args' RUSTFLAGS option will be overridden. Merge is not yet supported."
```
dmitry-timofeev: We discovered a problem with this code on Friday: you will usually run this script in your terminal so that you can use `mvn foo` of `cargo baz`. But the second time you use it (directly or indirectly), it will close your terminal :smirk_cat:

vitvakatu: Let's just print warning message then

dmitry-timofeev: Yep, that's probably the sanest option we have.

tests_profile
```diff
@@ -0,0 +1,69 @@
...
+#
+# [1] https://amir.rachum.com/blog/2016/09/17/shared-libraries/#runtime-search-path
+# [2] https://github.com/aidansteele/osx-abi-macho-file-format-reference#header-structure-and-load-commands
+
+
+# Use the Java that Maven uses.
+#
+# Unfortunately, a simple `which java` will not work for some users (e.g., jenv),
+# hence this a bit complex thing.
+export JAVA_HOME="$(mvn --version | grep 'Java home' | sed 's/.*: //')"
```
atrykush: better the value provided by the running version of java
```bash
export JAVA_HOME="$(java -XshowSettings:properties -version 2>&1 > /dev/null | grep 'java.home' | awk '{print $3}')"
```

PR #337, opened 2018-07-03
ECR-1523 Add code coverage integration with coveralls

MakarovS:
## Overview

Add code coverage integration with coveralls.

---
See: https://jira.bf.local/browse/ECR-1523

### Definition of Done


.travis.yml
```diff
@@ -18,7 +18,8 @@ env:
     - RUST_NIGHTLY_VERSION=nightly-2018-06-29
     - RUST_CLIPPY_VERSION=0.0.211
     - EJB_RUST_BUILD_DIR="$TRAVIS_BUILD_DIR/exonum-java-binding-core/rust/"
-
+  matrix:
+    secure: Fp22iaJpttsIArAyWmdCGNtljIALTYRVKO7O+H2hgBkwHHqrU7+15sbaq3xzhz4YNWNfuFMIkFUBgd/KYHgAuNDDrtm2agib13C0lQT1NFQO9ccmNCJNsXQrYrXGwpnNqPKp0YmfBfgNwzEpBerlbtvzV/T/RZukT/403XxwxU9y5tHfQokwVLibqP2jJsxdihTfCKIOs+o6hBfArmsn+e+panEv17ZrCjOmBIM/W70Rf2rEM26wFnYsfnAUTCkpl4Ong0SYNpZZxNMtw61W8ApDY8bpz7cKUxCv7SmD3kO7Y+TTHWfWYx6FNXtUpE1vCi6I7fZAY16rViTWOX55NCeFQz56XER7ArJQZtC/nC1lZ9tGKtcofu2Rq7WUoRuTwvLTaf6VzAP/CUj0DUxkV+8WUggl3s/Im7Y9rn8Aqvh8LReZmqzTY+dJ0hFG4DLoLtl71eTEnNoumi5UleBhJPaei3wPNPHg1WlOmhFyhRCsbIIGiyFtSj/faLmdc7tN/sBFANb0g4Exl0mRNvB0IfS1gM6XouEGUTlVree68p11PnsGJGs/QaUB9F9AAGVKTZ2kz7sqkCDdGmLxzbdidYDHZtYWfOIYSJCQsA09n2Txi0fwNByKfl/spdyMmtI1uGeT803rhN9vu0NGrQFG3mU7mqO33fUDEStIQ6/xn0A=
```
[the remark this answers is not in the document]
MakarovS: Yes, as we have to integrate our builds with coveralls using its REPO_TOKEN which we want to remain private, travis allows us to encode it this way and use it as a global env variable.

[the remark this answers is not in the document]
dmitry-timofeev: :ok_man: , please add what this particular token corresponds to.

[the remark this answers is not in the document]
MakarovS: Sure, I will!

exonum-java-binding-core/pom.xml
```diff
@@ -144,7 +144,10 @@
           <excludes>
             <exclude>**/*IntegrationTest.java</exclude>
           </excludes>
-          <argLine>${java.vm.assertionFlag}</argLine>
+          <argLine>
+            ${argLine}
```
dmitry-timofeev: I don't see where the 'argLine' property is set. Is it a standard one?

exonum-java-binding-core/pom.xml
```diff
@@ -144,7 +144,10 @@
...
+            ${java.vm.assertionFlag}
+          </argLine>
```
dmitry-timofeev: Shall this property be set for failsafe running ITs :arrow_down: ?

pom.xml
```diff
@@ -270,6 +261,31 @@
 
       <build>
         <plugins>
+          <plugin>
+            <groupId>org.jacoco</groupId>
+            <artifactId>jacoco-maven-plugin</artifactId>
+            <version>0.8.1</version>
+            <configuration>
+              <destFile>${basedir}/target/coverage-reports/jacoco-unit.exec</destFile>
+              <dataFile>${basedir}/target/coverage-reports/jacoco-unit.exec</dataFile>
```
dmitry-timofeev: Is dest file = data file on purpose?

pom.xml
```diff
@@ -270,6 +261,31 @@
 
       <build>
         <plugins>
+          <plugin>
+            <groupId>org.jacoco</groupId>
+            <artifactId>jacoco-maven-plugin</artifactId>
```
dmitry-timofeev: Is it always enabled (i.e., when I run tests locally)?

pom.xml
MakarovS: Yes, those two being the same is expected

pom.xml
MakarovS: No, expected behaviour is that it is only enabled when `ci-build` profile is active.
I really want to test it all, but I guess we can't unless we stop skipping tests (coverage report isn't prepared if we skip all tests). So this will have to wait until fixing that `LD_LIBRARY_PATH` issue.

exonum-java-binding-core/pom.xml
MakarovS: jacoco:prepare-agent prepares it.
> Prepares a property pointing to the JaCoCo runtime agent that can be passed as a VM argument to the application under test. Depending on the project packaging type by default a property with the following name is set:
tycho.testArgLine for packaging type eclipse-test-plugin and
argLine otherwise.

https://www.eclemma.org/jacoco/trunk/doc/prepare-agent-mojo.html

exonum-java-binding-core/pom.xml
MakarovS: Good catch, I'll look into that.

pom.xml
MakarovS: That's what I hoped for - 0% coverage just to be sure that we did everything and integrated with coveralls correctly. But when we skip test, we don't get report with 0% - we don't get any report at all and coveralls plugin doesn't do anything :(
I'll try to use this https://github.com/exonum/exonum-java-binding/pull/336 for now and try to connect with coveralls locally.

MakarovS:
This one is blocked until we stop skipping tests in travis.

dmitry-timofeev:
FYI: https://twitter.com/_godin_/status/1033047553481293824

.travis.yml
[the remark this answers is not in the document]
dmitry-timofeev: Shall it be under matrix?

.travis/run_travis_job.sh
```diff
@@ -47,5 +47,5 @@ then
     echo 'Rust checks are completed.'
 else
     cd "${TRAVIS_BUILD_DIR}"
-    ./run_all_tests.sh
+    ./run_all_tests_with_coverage.sh
```
dmitry-timofeev: We shall run coverage on Mac only — as Linux currently skips some tests.

exonum-java-binding-core/pom.xml
dmitry-timofeev: ^

exonum-java-binding-service-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -18,6 +18,8 @@
         <hamcrest.version>1.3</hamcrest.version>
         <mockito-core.version>2.22.0</mockito-core.version>
         <exonum-bom.version>0.2</exonum-bom.version>
+        <!--This property is set by jacoco and used in the surefire config further down-->
+        <argLine></argLine>
```
dmitry-timofeev: Why do we set it in archetype — it's the template of a service, and there is no jacoco configuration (and I believe there must not be).

pom.xml
```diff
@@ -309,6 +309,50 @@
       <properties>
         <checkstyle.severity>error</checkstyle.severity>
       </properties>
+
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.jacoco</groupId>
+            <artifactId>jacoco-maven-plugin</artifactId>
```
dmitry-timofeev: What are the implications of always running the tests with jacoco? Why don't we enable the coverage collection on demand?

dmitry-timofeev:
@MakarovS , does it collect correctly the coverage of Java code by the native ITs? I see it requires to set some command line arguments when running Java tests, but none are passed to the JVM when started from native ITs.

pom.xml
dmitry-timofeev: Sorry, I didn't notice, that's good then.

dmitry-timofeev:
Coveralls has a rather unfriendly UI: they don't have a coverage map; they collapse the packages, so one has to click all the way to `com.acme.foo.bar.….Baz`. Perhaps, the next PR will migrate this to Codecov :-)

MakarovS:
> Coveralls has a rather unfriendly UI: they don't have a coverage map; they collapse the packages, so one has to click all the way to com.acme.foo.bar.….Baz. Perhaps, the next PR will migrate this to Codecov :-)

Maybe, I've chosen it because we already have several of our exonum projects using it. We'll need to look into that.

dmitry-timofeev:
> Maybe, I've chosen it because we already have several of our exonum projects using it. We'll need to look into that.

A bit later — as we start using it

run_all_tests_with_coverage.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+# Run all tests, generate code coverage report and post it on coveralls.
+
+# Run all java tests, native unit tests and native integration tests.
+./run_all_tests.sh
+
+# Generate a coverage report.
+mvn org.jacoco:jacoco-maven-plugin:report org.eluder.coveralls:coveralls-maven-plugin:report
```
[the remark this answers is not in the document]
MakarovS: > Also, shan't we use https://www.eclemma.org/jacoco/trunk/doc/report-aggregate-mojo.html ?

Why should we? Coveralls aggregates it for us, no?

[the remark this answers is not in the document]
dmitry-timofeev: > Why should we? Coveralls aggregates it for us, no?

In this script probably not, but if you'd like to see an _aggregated_ coverage in your local project during development (i.e., by unit and integration tests, possibly, from different modules), you won't have that with individual reports.

pom.xml
```diff
@@ -326,43 +326,43 @@
...
                   <goal>report</goal>
                 </goals>
                 <configuration>
-                  <dataFile>${basedir}/target/coverage-reports/jacoco-unit.exec</dataFile>
+                  <dataFile>${project.build.directory}/coverage-reports/jacoco-unit.exec</dataFile>
                 </configuration>
               </execution>
               <execution>
-                <id>pre-integration-test</id>
+                <id>pre-integration-test-prepare-jacoco-agent</id>
```
dmitry-timofeev: :disappointed:

dmitry-timofeev:
@MakarovS , 

What are the implications of always running the tests with jacoco in terms of time it takes to perform `./run_all_tests.sh`? 

As it is in a ci-build profile, the coverage data will be always collected when `run_all_tests.sh` is executed.

MakarovS:
@dmitry-timofeev I didn't measure the overhead that jacoco agent adds, I can look into it, but I expect it to be insignificant. But as I mentioned above - https://github.com/exonum/exonum-java-binding/pull/337#discussion_r220420333 I thought that we always want to measure coverage in ci-builds. If we run all tests and don't want ci-build profile to be activated, maybe we should use different script for that?

dmitry-timofeev:
> I thought that we always want to measure coverage in ci-builds. If we run all tests and don't want ci-build profile to be activated, maybe we should use different script for that?

Maybe a different script or a different profile. There are a couple of reasons for that: first, see a0ade213ac2a1af00e0587421b20fbb93322fae3 — run_all_tests is used interactively and enables CI profile; second, and you see that in .travis.yml — coverage info is not used in any but a single build.

Would you please add a task to consider separating coverage-collection and ci-build profile?

pom.xml
MakarovS: Sorry I tried to come up with a better name, but it's often difficult to think of something other than phase + goal.
In most code samples I've come across it's just default-<phase>

MakarovS:
> Would you please add a task to consider separating coverage-collection and ci-build profile?

Sure! Though maybe we should do something about run_all_tests enabling CI profile, as it's not always the case. As you mentioned above -

> if you'd like to see an aggregated coverage in your local project during development (i.e., by unit and integration tests, possibly, from different modules), you won't have that with individual reports.

So the developer often wants to run tests locally and it's strange to use CI build. What do you think?

PR #344, opened 2018-07-06
Fix rustfmt on travis.

alexander-irbis:
## Overview

`cargo fmt -- --write-mode=diff` does not fail tests now if bad formatting is detected.
`cargo fmt -- --write-mode=check` should be used instead.

### Definition of Done


dmitry-timofeev:
> cargo fmt -- --write-mode=diff does not fail tests now if bad formatting is detected.

Wait, why? I think it shall fail build?

dmitry-timofeev:
Or you made it fail?

alexander-irbis:
This behavior has changed in the rustfmt itself.

PR #346, opened 2018-07-06
Allow failures of CI build on JDK 10:

dmitry-timofeev:
## Overview
There is a flaky native IT that fails quite frequently.

See: ECR-1734

### Definition of Done


dmitry-timofeev:
> Isn't there a way to disable just this one test on JDK 10? Or is it too much work? I don't think we can say that we support JDK 9+, if we completely discard JDK 10 CI build results.

With a flaky test we probably can't either :-) 

Whilst fixing the test, we can build instead on JVM 9 and _not_ allow failures there. WDYT?

dmitry-timofeev:
Updated to the latest master, matrix is populated as expected: 
image

PR #347, opened 2018-07-08
Fix compiler warnings:

dmitry-timofeev:
## Overview

- Fix the compiler warning of unchecked operations
  in Cleaner.
- Fix deprecation compiler warnings in AbstractHasher
  and EqualsTest.
- Suppress unchecked operations warnings in tests.

Also, enable adding compiler arguments in the parent POM.
Default is "merge" which will discard any <arg> child elements in the parent.

---

### Definition of Done


exonum-java-proofs/src/main/java/com/exonum/binding/hash/AbstractHasher.java
```diff
@@ -133,4 +133,12 @@ public Hasher putChar(char c) {
     funnel.funnel(instance, this);
     return this;
   }
+
+  @Override
+  @Deprecated
+  public int hashCode() {
```
dmitry-timofeev: An alternative would be to suppress deprecation warnings for the whole class.

PR #349, opened 2018-07-09
Rollback Rust compiler version [ECR-1853]

vitvakatu:
## Overview

We're going to use Rust 1.26.2 until next release because of bug in Cargo.
**However, we're using latest stable Rust for rustfmt.**

---
See: https://jira.bf.local/browse/ECR-1839

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/run_ejb_app_tests.sh
+++ b/run_ejb_app_tests.sh
@@ -21,7 +21,7 @@ cd exonum-java-binding-core/rust
 
 ## Stable works well unless you want benchmarks.
 # TODO: stable does not work well until ECR-1839 is resolved
-RUST_COMPILER_VERSION="1.26.2"
+RUST_COMPILER_VERSION="${RUST_VERSION:-1.26.2}"
 
 cargo "+${RUST_COMPILER_VERSION}" test \
   --manifest-path ejb-app/Cargo.toml
--- a/run_ejb_app_tests.sh
+++ b/run_ejb_app_tests.sh
@@ -19,8 +19,9 @@ echo "LD_LIBRARY_PATH=${LD_LIBRARY_PATH}"
 
 cd exonum-java-binding-core/rust
 
-# Stable works well unless you want benchmarks.
-RUST_COMPILER_VERSION="stable"
+## Stable works well unless you want benchmarks.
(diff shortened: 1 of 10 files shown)


vitvakatu:
I don't think #336 is controversial. It looks nice and easy to use scripts by @alexander-irbis to specify Rust compiler version explicitly.
If you don't want #336 to be landed first, I can rework this PR.

dmitry-timofeev:
I’ll look at #336 to see if we can integrate it first.

vitvakatu:
I'll check if new 1.27.1 version fixes the problem.

vitvakatu:
Unfortunately not :(

vitvakatu:
Rebased, now it's on top of master.

run_ejb_app_tests.sh
```diff
@@ -19,8 +19,9 @@ echo "LD_LIBRARY_PATH=${LD_LIBRARY_PATH}"
 
 cd exonum-java-binding-core/rust
 
-# Stable works well unless you want benchmarks.
-RUST_COMPILER_VERSION="stable"
+## Stable works well unless you want benchmarks.
+# TODO: stable does not work well until ECR-1839 is resolved
+RUST_COMPILER_VERSION="1.26.2"
```
dmitry-timofeev: What if we make these scripts aware of the `RUST_VERSION` environment variable: if it's set use it, if it is not — use default?

vitvakatu: :thinking: but in this case it would be hard to run these scripts. You suggest to set `RUST_VERSION` in our dev envrironments (manually)?

dmitry-timofeev: No, 
>  if it is not [set] — use default

We have already done this with JAVA_HOME, it may be done easily with this feature: https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

vitvakatu: Well, what do you mean by `default`? Stable? Rustup default? 1.26.2?

dmitry-timofeev: Previously — stable, at this moment — 1.26.2.

PR #352, opened 2018-07-10
Fix type of the body of the transfer transaction message.

dmitry-timofeev:
## Overview

See:
exonum-java-binding-cryptocurrency-demo/src/main/proto/transactions.proto:6


boguslavsky:
will be fixed in #353

PR #358, opened 2018-07-17
ECR-1863 Deployment of parent and bom

MakarovS:
## Overview

Make some changes to configuration (deploy parent and BOM projects). The modules to be deployed are:

1. ejb-parent
2. ejb-bom
3. ejb-core
4. ejb-proofs
5. ejb-archetype.

---
See: https://jira.bf.local/browse/ECR-1863

### Definition of Done

### DIFF
--- a/exonum-java-binding-cryptocurrency-demo/pom.xml
+++ b/exonum-java-binding-cryptocurrency-demo/pom.xml
@@ -6,13 +6,13 @@
   <parent>
     <artifactId>exonum-java-binding-parent</artifactId>
     <groupId>com.exonum.binding</groupId>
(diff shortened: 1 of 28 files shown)


exonum-java-binding-cryptocurrency-demo/pom.xml
```diff
@@ -94,6 +94,14 @@
           </execution>
         </executions>
       </plugin>
+
+      <plugin>
```
dmitry-timofeev: Shan't we add a comment explaining that we skip deployment of internal modules, which is inherited from the parent?

PR #359, opened 2018-07-17
Fix start_qa_node script after adding JVM arguments.

vitvakatu:
## Overview

Debug flag is no longer used, we need to use ejb-jvm-args instead

---
See: https://jira.bf.local/browse/XYZ

### Definition of Done


exonum-java-binding-core/rust/ejb-app/start_qa_node.sh
```diff
@@ -80,7 +80,7 @@ do
      --ejb-classpath $EJB_CLASSPATH \
      --ejb-libpath $EJB_LIBPATH \
      --ejb-log-config-path $log_config_path \
-     --ejb-debug true \
+     --ejb-jvm-args Xdebug \
```
dmitry-timofeev: AFAIK, it's not useful by itself (without extra flags), therefore, better be removed.

PR #384, opened 2018-07-25
Set JAVA_HOME according to the information provided by java itself

atrykush:
## Overview
The script depends on the line containing "Java home" string in the output of the 'mvn --version' command, which doesn't have this line. This is fair for Maven 3.5.4 with JDK 1.8.0 and JDK 10.0.2.
Besides that, I prefer rather check whether $JAVA_HOME is defined in the user's environment, than extract it from the third-party command (which could change it's output somewhere in the future).
Anyway, this PR fixes build of the exonum-java-binding-core.
Thanks.

### DIFF
--- a/exonum-java-binding-core/rust/cargo.sh
+++ b/exonum-java-binding-core/rust/cargo.sh
@@ -13,7 +13,7 @@ set -eu -o pipefail
 #
 # Unfortunately, a simple `which java` will not work for some users (e.g., jenv),
 # hence this a bit complex thing.
-JAVA_HOME="$(mvn --version | grep 'Java home' | sed 's/.*: //')"
+JAVA_HOME="$(mvn --version | grep runtime | sed 's/.*: //')"
 echo "JAVA_HOME=${JAVA_HOME}"
 
 # Find the directory containing libjvm (the relative path has changed in Java 9)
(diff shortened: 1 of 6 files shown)


dmitry-timofeev:
Interesting that we specifically changed the script to *not* use `JAVA_HOME` from the environment: 33376bf4100eb454c1409456d19daf09029479ee

We did that just before release when were integrating native-proxies into master, so there is no detailed enough description in the commit :-) But I believe we did that because Maven (?) set the environment variable incorrectly (so it *was* set by *Maven*), but it did so incorrectly on some systems.

Does anyone remember the details (@vitvakatu , @alexander-irbis , @MakarovS )?

dmitry-timofeev:
I agree this script is absolutely not reliable, I remember there also was some code that printed the java system property 'java.home' by launching the jvm with some command line flags.

alexander-irbis:
In some cases system JAVA_HOME is set to directory in which we can't find libjvm.so.

dmitry-timofeev:
> In some cases system JAVA_HOME is set to directory in which we can't find libjvm.so.

Do you mean the case when supposedly Maven set it, which made us override it (https://github.com/exonum/exonum-java-binding/commit/33376bf4100eb454c1409456d19daf09029479ee)? 

Or does `java` incorrectly report it in some cases?

PR #392, opened 2018-07-30
Add TransactionExecutionException [ECR-1958]:

dmitry-timofeev:
## Overview
Add `TransactionExecutionException`. Such exception is thrown to indicate that the transaction cannot be executed and any changes must be rolled back. Its error code is included in the blockchain and can be requested by a client.

This exception corresponds to ExecutionError in the framework native API.

Also, add a method to create a transaction mock throwing `TransactionExecutionException` to be able to test the new behaviour in the native proxy code.

---
See: https://jira.bf.local/browse/ECR-1958

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+- `Transaction#execute` can throw `TransactionExecutionException` to roll back 
+  any changes to the database and pass an error code and an optional description 
+  to the framework which saves them to the storage.
+
 ### Removed
 - `Hashing#toHexString`. (#379)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 - `Transaction#execute` can throw `TransactionExecutionException` to roll back 
-  any changes to the database and pass an error code and an optional description 
-  to the framework which saves them to the storage.
+  any changes to the database. The exception includes an error code and an optional 
+  description which the framework saves to the storage for later retrieval. (#392)
 
 ### Removed
 - `Hashing#toHexString`. (#379)
--- a/exonum-java-binding-core/src/main/java/com/exonum/binding/messages/Transaction.java
+++ b/exonum-java-binding-core/src/main/java/com/exonum/binding/messages/Transaction.java
@@ -47,9 +47,14 @@ public interface Transaction {
    * Execute the transaction, possibly modifying the blockchain state.
    *
    * @param view a database view, which allows to modify the blockchain state
+   * @throws TransactionExecutionException if the transaction cannot be executed normally
+   * and has to be rolled back. The transaction will be committed as failed (status "error"),
+   * the error code and the optional description will be saved into the storage. The client
+   * can request the error code to know the reason of the failure.
+   * @throws RuntimeException if an unexpected error occurs. A correct transaction implementation
+   * must not throw such exceptions. The transaction will be committed as failed (status "panic").
    */
-  void execute(Fork view);
-
+  void execute(Fork view) throws TransactionExecutionException;
 
   /**
    * Returns some information about this transaction in JSON format.
--- a/exonum-java-binding-core/src/main/java/com/exonum/binding/messages/Transaction.java
+++ b/exonum-java-binding-core/src/main/java/com/exonum/binding/messages/Transaction.java
@@ -49,7 +49,7 @@ public interface Transaction {
    * @param view a database view, which allows to modify the blockchain state
    * @throws TransactionExecutionException if the transaction cannot be executed normally
    *     and has to be rolled back. The transaction will be committed as failed (status "error"),
-   *     the error code and the optional description will be saved into the storage. The client
+   *     the error code with the optional description will be saved into the storage. The client
    *     can request the error code to know the reason of the failure.
    * @throws RuntimeException if an unexpected error occurs. A correct transaction implementation
    *     must not throw such exceptions. The transaction will be committed as failed
--- a/exonum-java-binding-core/src/main/java/com/exonum/binding/messages/Transaction.java
+++ b/exonum-java-binding-core/src/main/java/com/exonum/binding/messages/Transaction.java
@@ -48,11 +48,12 @@ public interface Transaction {
    *
    * @param view a database view, which allows to modify the blockchain state
    * @throws TransactionExecutionException if the transaction cannot be executed normally
-   * and has to be rolled back. The transaction will be committed as failed (status "error"),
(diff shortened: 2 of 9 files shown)


exonum-java-binding-core/src/main/java/com/exonum/binding/messages/TransactionExecutionException.java
```diff
@@ -0,0 +1,103 @@
...
+ * are not saved in the database and are used for logging only.
+ *
+ * <p>An external client will get the error code and description when requests a transaction
+ * status of a failed transaction. See
+ * <a href="https://exonum.com/doc/advanced/node-management/#transaction">the API endpoint documentation</a>
+ * for more information.
+ */
+public class TransactionExecutionException extends Exception {
+
+  // TODO: Consider using enums and taking their ordinal as the error code?
```
dmitry-timofeev: WDYT? Won't be able to assign arbitrary values to enum elements, or skip some values, but it will be somewhat easier to work with.

vitvakatu: In core user defines his own enums and converts them into error codes.
https://github.com/exonum/exonum/blob/7ba9b1ed739c7baf0a13c6f70a8fca81ce6c5930/examples/cryptocurrency/src/lib.rs#L165-L199
I don't understand what you suggest to do in Java.

dmitry-timofeev: Passing any Enum and taking its position as the error code: https://docs.oracle.com/javase/10/docs/api/java/lang/Enum.html#ordinal()

We can require an Enum implementing a certain interface (like this with `short id()`: https://github.com/exonum/exonum-java-binding/blob/83010162593fb44f26677a7ed13fd4dcf36b5783/exonum-java-binding-qa-service/src/main/java/com/exonum/binding/qaservice/transactions/QaTransaction.java#L27), but it would be a bigger constraint.

vitvakatu: Ah, I see know, thanks. Actually, idea of enforcing enums usage seems good to me.

CHANGELOG.md
```diff
@@ -15,6 +15,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+- `Transaction#execute` can throw `TransactionExecutionException` to roll back 
+  any changes to the database and pass an error code and an optional description
```
bullet-tooth: ...and pass an error code **with** an optional description...

CHANGELOG.md
```diff
@@ -15,6 +15,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+- `Transaction#execute` can throw `TransactionExecutionException` to roll back 
+  any changes to the database and pass an error code and an optional description 
+  to the framework which saves them to the storage.
```
bullet-tooth: it's not clear what is `them`? Changes or exceptions?

exonum-java-binding-core/src/main/java/com/exonum/binding/messages/Transaction.java
```diff
@@ -47,9 +47,15 @@
    * Execute the transaction, possibly modifying the blockchain state.
    *
    * @param view a database view, which allows to modify the blockchain state
+   * @throws TransactionExecutionException if the transaction cannot be executed normally
+   *     and has to be rolled back. The transaction will be committed as failed (status "error"),
+   *     the error code and the optional description will be saved into the storage. The client
```
bullet-tooth: ...the error code **with** optional description...

exonum-java-binding-core/src/main/java/com/exonum/binding/messages/TransactionExecutionException.java
dmitry-timofeev: I am not sure we shall enforce that because if you have a single or a couple of error codes, defining an enum might seem too verbose (especially if that's an enum implementing a certain interface with `byte errorCode()` method). If we just use ordinals then it's OK in terms of verbosity, but the error codes are (a) implicit; (b) can no longer be arbitrary.

dmitry-timofeev: @bullet-tooth , @MakarovS , wdyt?

dmitry-timofeev: https://jira.bf.local/browse/ECR-2006

bullet-tooth: I also like to use enum here. As example we can handle user defined range of errors by some enum instance:
```
public enum ErrorEnum {
  SYSTEM_ERROR,
  CUSTOM_ERROR,
  DEFAULT_ERROR;
  
  public static ErrorEnum fromExonumErrorCode(int errorCode) {
    if(errorCode == -1) {
      return SYSTEM_ERROR;
    } else if(errorCode > 100 && errorCode <= 200) {
      return CUSTOM_ERROR;
    }
    return DEFAULT_ERROR;
  }
  
}
```

exonum-java-binding-fakes/src/test/java/com/exonum/binding/fakes/mocks/ThrowingTransactionsTest.java
```diff
@@ -60,5 +69,27 @@ public void createThrowingFailsIfUninstantiableThrowable() {
     ThrowingTransactions.createThrowing(UninstantiableException.class);
   }
 
+  @Test
+  @Ignore // Unfortunately, we do not perform such checks at the moment: ECR-1988
```
bullet-tooth: should `TODO` be added before tiket number?

dmitry-timofeev: Not necessarily :-)

exonum-java-binding-core/src/main/java/com/exonum/binding/messages/TransactionExecutionException.java
bullet-tooth: yes, but as far as I know there is a codes range for such errors. And my point was this range can be mapped to one enum value

dmitry-timofeev: I see, but how do you pass that value (= errorCode as a byte) to the native code, which does not understand enums? 

e.g.
```java
if (foo) {
  throw new ExecutionException(FOO);
} else if (bar) {
  throw new ExecutionException(BAR);
}
```

FOO & BAR may be:
* byte constants;
* simple enum constants (you translate them to a byte error code using their ordinal `int ordinal()`);
* enum constants implementing a certain interface (`byte toErrorCode()`).

PR #393, opened 2018-07-30
Fix scripts running tests [ECR-1993]:

dmitry-timofeev:
## Overview
The logic enabling tests has been moved to Maven build definitions,
therefore, there is no need to set the global user property
that will disable *all* tests (see 5e2055d).

Also, when compiling artefacts in run_native_integration_tests.sh,
there is no need to pass -DskipTests.

---

### Definition of Done

- [x] System libsodium is used by EJB-native
- [x] Libsodium is installed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,12 @@ sudo: false
 
 addons:
   apt:
+    sources:
+    - sourceline: 'ppa:chris-lea/libsodium'
     packages:
     - build-essential
+      # Install libsodium as it is required for the crypto package in the light client.
+    - libsodium-dev
     - libssl-dev
     - pkg-config
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,12 @@ sudo: false
 addons:
   apt:
     sources:
-    - sourceline: 'ppa:chris-lea/libsodium'
+      - sourceline: 'ppa:chris-lea/libsodium'
     packages:
-    - build-essential
-      # Install libsodium as it is required for the crypto package in the light client.
-    - libsodium-dev
-    - libssl-dev
-    - pkg-config
+      - build-essential
+      - libsodium-dev
+      - libssl-dev
+      - pkg-config
 
 env:
   global:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -46,9 +50,8 @@ before_install:
   - cargo-audit -V || cargo install cargo-audit --force
     # List all installed cargo packages.
   - cargo install --list
-    # libjava_bindings.so and native application both need to load common Rust libs (eg libstd.so).
-  - export LD_LIBRARY_PATH=$RUST_LIB_DIR:$LD_LIBRARY_PATH
-    # force building instead of using from apt.
+    # Force building Exonum dependencies instead of using the system ones, if any.
+    # Required because using system results in broken artefacts on older CPUs, see ECR-1658.
   - export SODIUM_BUILD=1
```
alexander-irbis: If we decided to install system libsodium.so for Java, should we reuse it in Rust part too?
`export SODIUM_LIB_DIR="/usr/lib"`

vitvakatu: We had crashes with libsodium from repos, remember?

dmitry-timofeev: I noticed that we build libsodium but couldn't remember _why_ we do. Still can't remember the crashes :man_shrugging: 

#399 might make system libsodium unnecessary

vitvakatu: afaik we had some problems (not exactly crashes), so @alexander-irbis found that building libsodium from source solves these problems.

dmitry-timofeev: I remember RocksDB & SIGILL on some older CPUs.

vitvakatu: :thinking: interesting, so why we build libsodium?

alexander-irbis: > We had crashes with libsodium from repos, remember?

Definitely that issue with SIGILL was discussed primarily in the context of RocksDB.
LIbsodium was switched to build at the same time with RocksDB but for another reason, as try to fix crashing issue on Linux. It did not help.

dmitry-timofeev: I see, thanks! So we could remove that line and have exonum pick up the system one?

dmitry-timofeev: Updated and cleared the caches on Travis to get a clean build, will see how it goes: https://travis-ci.org/exonum/exonum-java-binding/builds/411277584

PR #455, opened 2018-09-04
Update to Rust 1.28 [ECR-2027]

skletsun:
## Overview

All affected configuration files have been updated for using the current stable rust 1.28.0 toolchain.
Also, the EJB-core pom.xml has been adjusted in order to isolate the building of test artifacts from the main library.
---
See: https://jira.bf.local/browse/ECR-2027

### Definition of Done


CONTRIBUTING.md
```diff
@@ -15,8 +15,8 @@ You need to install the following dependencies:
   * Linux or macOS. Windows support is coming soon. <!-- TODO: Link Java roadmap when it is published -->
   * [JDK 1.8+](http://jdk.java.net/10/).
   * [Maven 3.5+](https://maven.apache.org/download.cgi).
-  * [Rust 1.26.2](https://www.rust-lang.org/).
-    To install specific Rust version, use `rustup install 1.26.2` command.
+  * [Rust 1.28.0](https://www.rust-lang.org/).
```
dmitry-timofeev: I'd use the previous version (the latest stable Rust) — see in the history.

.travis.yml
```diff
@@ -17,9 +17,9 @@ addons:
 
 env:
   global:
-    - RUST_COMPILER_VERSION=1.26.2
-    - RUST_NIGHTLY_VERSION=nightly-2018-06-29
-    - RUST_CLIPPY_VERSION=0.0.211
+    - RUST_COMPILER_VERSION=1.28.0
```
dmitry-timofeev: (here and elsewhere) Shall we still specify a particular version (1.28.0 instead of `stable`)?

exonum-java-binding-core/pom.xml
```diff
@@ -125,6 +128,8 @@
                 <argument>${rust.itSubCrate}</argument>
                 <argument>--color</argument>
                 <argument>always</argument>
+                <argument>--target-dir</argument>
```
dmitry-timofeev: These unit tests are run with the same features that the native library is built. As I understand, specifying `--target-dir` here will incur a full (?) rebuild even if we move "test-rust-lib-no-resource-manager" to ci-only profile. Shan't we specify `--target-dir` in "test-rust-lib-no-resource-manager" only?

.travis.yml
vitvakatu: Agreed, we should specify `stable`

vitvakatu:
We also need to fix `run_travis_job.sh` script - we're installing `stable` specially for rustfmt now.

exonum-java-binding-core/pom.xml
skletsun: If we specify `--target-dir` for `test-rust-lib-no-resource-manager` only it will build but won't override the `libjava_binding` shared library artifact built during the `build-rust-lib` phase and will crash with a dynamic linking error while executing the tests on the `test-rust-lib` phase. And yes - it will cause the full rebuild even if we move `test-rust-lib-no-resource-manager` to ci-only profile.

dmitry-timofeev: I am not sure I understand — you say that we specify --target-dir for test-rust-lib-no-resource-manager to isolate it from the rest of the builds, but isn't that enough? How a test _with_ a resource manager affects a test _without_ that uses _another_ target directory?

exonum-java-binding-core/pom.xml
```diff
@@ -22,10 +22,11 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <!-- TODO: stable does not work well until ECR-1839 is resolved. -->
-    <rust.compiler.version>1.26.2</rust.compiler.version>
+    <rust.compiler.version>1.28.0</rust.compiler.version>
```
dmitry-timofeev: I believe we shall default everywhere to stable, shan't we?

exonum-java-binding-core/pom.xml
skletsun: Tests _with_ and _without_ resource manager are sharing the same folder (with changes from this PR) specified with `--target-dir`. The main problem with the `test` phase when just upgrading to 1.28.0 step by step: (1) at the `build` step we build the _libjava_binding.dylib_ library, then the `test` phase comes and (2) tests _without_ resource-manager get built and executed against the library built at (1). The same is for (3) tests _with_ resource-manager that also require rebuilt (with cargo test) version of _libjava_binding.dylib_. My current solution is that I specify the same directory for _both_ of tests _with_ and _without_ resource-manager and they are executed one after another and share the resulting binaries from `--target-dir`. If I don't specify the `--target-dir` for tests _with_ resource-manager it will fail anyway because it will attempt to rebuild the _libjava_binding.dylib_ and _wont_ be able to replace the previous version from (1) and will fail with a dynamic linking error.

exonum-java-binding-core/pom.xml
vitvakatu: :thinking: We've had `stable` written here before.

exonum-java-binding-core/pom.xml
dmitry-timofeev: I see, thanks. This part seems strange:

>  If I don't specify the --target-dir for tests with resource-manager it will fail anyway because it will attempt to rebuild the libjava_binding.dylib and wont be able to replace the previous version from (1) and will fail with a dynamic linking error.

Why does it have to replace anything from (1) if it is built with the same flags (= features)? 

The problem with such solution (always separating targets for build and test) is that even if we don't run a build _without_ rm (2) locally — a build _with_ rm (3) will still increase the compilation time significantly.

exonum-java-binding-core/pom.xml
dmitry-timofeev: Exactly. And the paths were produced correctly as the command to do that uses rustup/rustc to determine the path.

vitvakatu:
Closing as heavily outdated and blocked. For now we're using version 1.27.2, we'll make a transition to 1.31 once it'll be released

PR #474, opened 2018-09-13
Ignore VS Code configuration

dmitry-timofeev:
## Overview

Ignore VS Code configuration

### Definition of Done


vitvakatu:
TBH I don't like such changes, because each developer should use their global `.gitignore` files

PR #479, opened 2018-09-17
Testing builds on OSX with Travis-CI

atrykush:
## Overview

Added OSX tests to the .travis.yml file.

see https://jira.bf.local/browse/ECR-2367

### Definition of Done

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ cache:
 
 before_install:
   # Skip the build if only md files were updated.
-  - source .travis/skip-ci.sh
+#  - source .travis/skip-ci.sh
   - export PATH="$PATH":"$HOME/.cargo/bin"
   # Install rustup if it's not already installed (i.e., not in CI cache).
   - which rustup > /dev/null || curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain "$RUST_COMPILER_VERSION"
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ cache:
 
 before_install:
   # Skip the build if only md files were updated.
-#  - source .travis/skip-ci.sh
+  - source .travis/skip-ci.sh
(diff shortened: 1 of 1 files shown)


atrykush:
In the light that I couldn't find a single file in the repo that checks $ROCKSDB_BUILD variable value, may I ask you to point to the script that actually build rocksdb and snappy libraries?
Thanks in advance.

dmitry-timofeev:
This variable was added to the travis build file in ba9c8ef50a1ca2ad296ea6511863e5be9e75f206 after its processing was added in https://github.com/exonum/exonum_rocksdb/pull/8

.travis.yml
```diff
@@ -25,8 +25,9 @@ env:
     - RUST_CLIPPY_VERSION=0.0.211
     - EJB_RUST_BUILD_DIR="$TRAVIS_BUILD_DIR/exonum-java-binding-core/rust/"
     # force building instead of using from apt.
-    - ROCKSDB_BUILD=1
-    - SNAPPY_BUILD=1
+    # let's see what would happen
+    # - ROCKSDB_BUILD=1
```
dmitry-timofeev: Please note that this is required on _some_ (older) Linux systems — it will most likely build fine on Linux this time.

atrykush: Trusty used by Travis is pretty old though, however, ROCKSDB_BUILD and SNAPPY_BUILD would be set only for linux system.

.travis.yml
```diff
@@ -53,6 +57,10 @@ before_install:
   - rustup default "$RUST_COMPILER_VERSION"
   # List all installed cargo packages.
   - cargo install --list
+  # Install OSX requirements
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install libsodium rocksdb pkg-config; fi
```
dmitry-timofeev: It's interesting that installing them from brew takes > 3 minutes.

PR #482, opened 2018-09-18
Update TransactionProxy#execute to use transaction status from Java [ECR-1956]

skletsun:
## Overview

The result of execution of a transaction is treated in a little different way now: any JNI error or Java throwable that _is not_ an instance or subclass of the `TransactionExecutionException` considered an unexpected exception and causes panic. `TransactionExecutionException` and its descendants are converted into `Error`s with their descriptions. Former behavior: JNI errors and anything other than `lang.java.Exception` and its descendants considered unexpected exception --> panic.

---
See: https://jira.bf.local/browse/ECR-1956

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding-core/rust/integration_tests/src/mock/transaction.rs
+++ b/exonum-java-binding-core/rust/integration_tests/src/mock/transaction.rs
@@ -1,5 +1,5 @@
 use java_bindings::exonum::messages::{MessageBuffer, RawMessage};
-use java_bindings::jni::objects::{GlobalRef, JObject, JString, JValue};
+use java_bindings::jni::objects::{GlobalRef, JObject, JValue};
 use java_bindings::serde_json::Value;
 use java_bindings::{JniExecutor, MainExecutor, TransactionProxy};
 
@@ -55,7 +55,7 @@ pub fn create_throwing_exec_exception_mock_transaction_proxy(
                     format!("(BLjava/lang/String;)L{};", TRANSACTION_ADAPTER_CLASS),
                     &[
                         JValue::from(error_code),
-                        JValue::from(JObject::from(JString::from(err_msg))),
+                        JValue::from(JObject::from(err_msg)),
                     ],
                 )?.l()?;
             let java_tx_mock = env.new_global_ref(java_tx_mock)?;
--- a/exonum-java-binding-core/rust/integration_tests/src/mock/transaction.rs
+++ b/exonum-java-binding-core/rust/integration_tests/src/mock/transaction.rs
@@ -1,5 +1,5 @@
 use java_bindings::exonum::messages::{MessageBuffer, RawMessage};
-use java_bindings::jni::objects::{GlobalRef, JObject, JValue};
+use java_bindings::jni::objects::{GlobalRef, JObject, JString, JValue};
 use java_bindings::serde_json::Value;
 use java_bindings::{JniExecutor, MainExecutor, TransactionProxy};
 
@@ -30,6 +30,33 @@ pub fn create_throwing_mock_transaction_proxy(
                     "createThrowingTransaction",
                     format!("(Ljava/lang/Class;)L{};", TRANSACTION_ADAPTER_CLASS),
                     &[JValue::from(JObject::from(exception.into_inner()))],
+                )?.l()?;
+            let java_tx_mock = env.new_global_ref(java_tx_mock)?;
+            let raw = RawMessage::new(MessageBuffer::from_vec(vec![]));
+            Ok((java_tx_mock, raw))
+        }).unwrap();
+
+    TransactionProxy::from_global_ref(executor, java_tx_mock, raw)
+}
+
+/// Creates `TransactionProxy` which throws TransactionExecutionException on the `execute` call.
+pub fn create_throwing_exec_exception_mock_transaction_proxy(
+    executor: MainExecutor,
+    error_code: i8,
+    error_message: &str,
+) -> TransactionProxy {
+    let (java_tx_mock, raw) = executor
+        .with_attached(|env| {
+            let err_msg = env.new_string(error_message)?;
+            let java_tx_mock = env
+                .call_static_method(
+                    NATIVE_FACADE_CLASS,
+                    "createThrowingExecutionExceptionTransaction",
+                    format!("(BLjava/lang/String;)L{};", TRANSACTION_ADAPTER_CLASS),
+                    &[
+                        JValue::from(error_code),
+                        JValue::from(JObject::from(JString::from(err_msg))),
+                    ],
                 )?.l()?;
             let java_tx_mock = env.new_global_ref(java_tx_mock)?;
             let raw = RawMessage::new(MessageBuffer::from_vec(vec![]));
--- a/exonum-java-binding-core/rust/integration_tests/src/mock/transaction.rs
+++ b/exonum-java-binding-core/rust/integration_tests/src/mock/transaction.rs
@@ -48,6 +48,13 @@ pub fn create_throwing_exec_exception_mock_transaction_proxy(
 ) -> TransactionProxy {
     let (java_tx_mock, raw) = executor
         .with_attached(|env| {
+            let msg = match error_message {
+                Some(err_msg) => {
+                    let msg = env.new_string(err_msg)?;
+                    JObject::from(msg)
+                }
+                None => JObject::null(),
+            };
             let java_tx_mock = env
                 .call_static_method(
                     NATIVE_FACADE_CLASS,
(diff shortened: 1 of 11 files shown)


exonum-java-binding-core/rust/src/utils/errors.rs
```diff
@@ -46,6 +48,35 @@ pub fn check_error_on_exception<T>(env: &JNIEnv, result: JniResult<T>) -> Result
...
 }
 
+/// Handles exceptions after executing transactions
+///
+/// The TransactionExecutionException and its descendants are converted into `Error`s with their
+/// descriptions. The rest (Java and JNI errors) are treated as unrecoverable and result in a panic.
+///
+/// Panics:
+/// - Panics if there is some JNI error.
+/// - If there is a pending Java throwable that IS NOT an instance or subclass of the
```
dmitry-timofeev: that IS NOT an instance of

(an instance of implies subclasses)

exonum-java-binding-core/rust/src/utils/errors.rs
```diff
@@ -46,6 +48,35 @@ pub fn check_error_on_exception<T>(env: &JNIEnv, result: JniResult<T>) -> Result
...
+/// `TransactionExecutionException`.
+pub fn check_transaction_execution_result<T>(
+    env: &JNIEnv,
+    result: JniResult<T>,
+) -> Result<T, String> {
+    result.map_err(|jni_error| match jni_error.0 {
+        JniErrorKind::JavaException => {
+            let exception = get_and_clear_java_exception(env);
+            let message = describe_java_exception(env, exception);
```
dmitry-timofeev: I think we shall get the exception message instead of using `describe_java_exception` which includes the class name and is vaguely defined.

exonum-java-binding-core/rust/src/utils/errors.rs
```diff
@@ -46,6 +48,35 @@ pub fn check_error_on_exception<T>(env: &JNIEnv, result: JniResult<T>) -> Result
...
+/// Handles exceptions after executing transactions
+///
+/// The TransactionExecutionException and its descendants are converted into `Error`s with their
+/// descriptions. The rest (Java and JNI errors) are treated as unrecoverable and result in a panic.
+///
+/// Panics:
+/// - Panics if there is some JNI error.
+/// - If there is a pending Java throwable that IS NOT an instance or subclass of the
+/// `TransactionExecutionException`.
+pub fn check_transaction_execution_result<T>(
```
dmitry-timofeev: I don't think this class belongs here — (in errors) — it is used in TransactionProxy only and must be a private method there (not a public in separate module) and tested through the `Transaction` public interface.

It must also include the error code from the exception — that’s the whole point of this exception type.

exonum-java-binding-core/rust/integration_tests/tests/transaction_proxy.rs
```diff
@@ -84,20 +91,51 @@ fn execute_should_panic_if_java_error_occurred() {
...
+    let mut fork = db.fork();
     let err = invalid_tx
         .execute(&mut fork)
         .map_err(TransactionError::from)
         .expect_err("This transaction should be executed with an error!");
-    assert!(
-        err.description()
-            .unwrap()
-            .starts_with("Java exception: java.lang.ArithmeticException",)
+    assert!(err.description().unwrap().starts_with(err_message));
```
dmitry-timofeev: We must assert that the err includes the code — that's the feature we are testing.

exonum-java-binding-core/rust/src/proxy/transaction.rs
```diff
@@ -119,3 +119,44 @@ impl Message for TransactionProxy {
...
+/// - If there is a pending Java throwable that IS NOT an instance of the
+/// `TransactionExecutionException`.
+fn check_transaction_execution_result<T>(
+    env: &JNIEnv,
+    result: JniResult<T>,
+) -> Result<T, ExecutionError> {
+    result.map_err(|jni_error| match jni_error.0 {
+        JniErrorKind::JavaException => {
+            let exception = get_and_clear_java_exception(env);
+            let message = unwrap_jni_verbose(env, get_exception_message(env, exception));
```
dmitry-timofeev: Do we have to use `unwrap_jni_verbose` in this method (asking as it adds some noise)? 

From the spec, "unwraps … with verbose error message **if** Java exception occurred", but in these cases exceptions seem unlikely (you just call two getters and a JNI API method).

exonum-java-binding-core/rust/src/utils/errors.rs
```diff
@@ -126,7 +95,7 @@ pub fn get_and_clear_java_exception<'e>(env: &'e JNIEnv) -> JObject<'e> {
     exception
 }
 
-fn describe_java_exception(env: &JNIEnv, exception: JObject) -> String {
+pub fn describe_java_exception(env: &JNIEnv, exception: JObject) -> String {
```
dmitry-timofeev: Would you please add a short doc comment to a newly public method?

exonum-java-binding-core/rust/integration_tests/tests/transaction_proxy.rs
```diff
@@ -84,20 +91,51 @@ fn execute_should_panic_if_java_error_occurred() {
...
+        err_message,
     );
+    let db = MemoryDB::new();
+    let mut fork = db.fork();
+    let err = invalid_tx
+        .execute(&mut fork)
+        .map_err(TransactionError::from)
+        .expect_err("This transaction should be executed with an error!");
+    assert!(err.description().unwrap().starts_with(err_message));
 }
```
dmitry-timofeev: Can we also test a case with no message (getMessage _may_ return `null`)?

exonum-java-binding-core/rust/integration_tests/tests/transaction_proxy.rs
dmitry-timofeev: one of the features :-)

exonum-java-binding-core/rust/integration_tests/tests/transaction_proxy.rs
dmitry-timofeev: It's up to user to specify the message.

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/NativeFacade.java
```diff
@@ -73,6 +75,24 @@ public static UserTransactionAdapter createThrowingTransaction(
...
+  /**
+   * Creates a UserTransactionAdapter that contains a transaction that throws
+   * {@link TransactionExecutionException} in its execute method.
+   *
+   * @param isSubclass whether method should produce a subclass of TransactionExecutionException
+   * @param errorCode an error code that will be included in the exception
+   * @param description a description; may be {@code null}
+   * @return a transaction mock throwing in execute
+   */
+  public static UserTransactionAdapter createThrowingExecutionExceptionTransaction(
```
dmitry-timofeev: :+1:

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/test/TestTxExecException.java
```diff
@@ -0,0 +1,22 @@
+package com.exonum.binding.fakes.test;
+
+import com.exonum.binding.transaction.TransactionExecutionException;
+
+import javax.annotation.Nullable;
+
+/**
+ * Used in tests that cover the cases of using subclass of {@link #TransactionExecutionException}.
+ */
+public class TestTxExecException extends TransactionExecutionException {
```
dmitry-timofeev: package-private if it's not used outside the package

exonum-java-binding-core/rust/integration_tests/src/mock/transaction.rs
```diff
@@ -57,7 +56,13 @@ pub fn create_throwing_exec_exception_mock_transaction_proxy(
                     &[
                         JValue::from(is_subclass),
                         JValue::from(error_code),
-                        JValue::from(JObject::from(err_msg)),
+                        match error_message {
+                            Some(err_msg) => {
+                                let msg = env.new_string(err_msg)?;
+                                JValue::from(JObject::from(msg))
+                            }
+                            None => JValue::Void,
```
dmitry-timofeev: Is `JValue::Void` the same as `JValue::from(JObject::null())`? Shan't we use sth like that:

```rust
let msg = match error_message {
  Some(err_msg) => {
    let msg = env.new_string(err_msg)?;
    JObject::from(msg)
  }
  None => JObject::null(),
}

// call the method, wrapping the JObject into JValue.
```

skletsun: Well, from Rust's point of view `JValue::from(JObject::null())` != `JValue::Void`, but it looks like for Java it is the same.

exonum-java-binding-core/rust/src/utils/errors.rs
```diff
@@ -95,7 +96,8 @@ pub fn get_and_clear_java_exception<'e>(env: &'e JNIEnv) -> JObject<'e> {
     exception
 }
 
-fn describe_java_exception(env: &JNIEnv, exception: JObject) -> String {
+/// Describes Java exception in a form of "Java ecception: EXCEPTION_NAME: EXCEPTION_DETAILS"
```
vitvakatu: > ecception

nit: typo

PR #487, opened 2018-09-20
Update Guice

dmitry-timofeev:
## Overview
Update Guice to 4.2.1, which has gained a proper Java 10
support in AOP-enabled version. This is likely to fix
the flaky service_runtime IT (will be verified separately).

See also: https://github.com/google/guice/issues/1181#issuecomment-423216468

---

### Definition of Done

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,11 +45,6 @@ matrix:
       env: CHECK_RUST=true
     - os: osx
       jdk: openjdk10
-  # See ECR-1734
-  allow_failures:
-    - jdk: openjdk10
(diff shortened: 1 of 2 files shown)


dmitry-timofeev:
Well, it didn't work — we got a failure of that service runtime test in the first build. However, the PR from dependabot succeeded — so it is still flaky.

I guess we'll need to reproduce the issue with sth simpler. Or go back to non-AOP flavour of Guice (I think that the only usage of it must be replaced anyway with hand-written logging because log levels must be different for each method).

dmitry-timofeev:
Already using a more recent version.

PR #494, opened 2018-09-24
 Fix the DbKey#hashCode [ECR-2416]

dmitry-timofeev:
## Overview
DbKey#hashCode turned out to be broken: it passes array references to
Objects.hashCode, instead of passing them to Arrays.hashCode.

---
See: https://jira.bf.local/browse/ECR-2416

### Definition of Done

- [x] Public API has Javadoc


dmitry-timofeev:
I think we’d require EqualsVerifier.verify for any class overriding e&h, wdyt?

exonum-java-binding-common/src/main/java/com/exonum/binding/common/proofs/map/DbKey.java
```diff
@@ -78,6 +77,11 @@ private static Type from(byte code) {
    */
   public static final int DB_KEY_SIZE = KEY_SIZE + 2;
 
+  /**
+   * Position of the user key in a database key.
+   */
+  private static final int KEY_START_POSITION = 1;
```
DENMROOT: USER_KEY_START_POSITION without comments

DENMROOT: as for me it will clarify, usage of comments for clarification means bad naming

PR #522, opened 2018-10-05
Tidy POMs [ECR-2406]

dmitry-timofeev:
## Overview

 - Apply tidy POM to make the order of all elements consistent.
 - Use two spaces everywhere
 - Use project.version to refer to the current version where applicable

Recommend to review commit-by-commit.

---
See: https://jira.bf.local/browse/ECR-2406

### Definition of Done


dmitry-timofeev:
We could also enable tidy:check so that it fails the build if the POM does not match the [convention](https://maven.apache.org/developers/conventions/code.html) but that might be an overkill.

MakarovS:
> We could also enable tidy:check so that it fails the build if the POM does not match the convention but that might be an overkill.

I think it'd be okay. I guess if we apply tidy pom now, it makes sense to keep it that way.

dmitry-timofeev:
> I think it'd be okay. I guess if we apply tidy pom now, it makes sense to keep it that way.

Makes sense. tidy:check is real fast so it won't increase the build time any significantly.

PR #525, opened 2018-10-08
Fix coverage reporting:

dmitry-timofeev:
## Overview

"report" goal bound to post-integration-test phase overwrote
the previous report from unit tests bound to "prepare-package"
phase. It happened because both execution of the *same* goal
used the same output directory.

---

### Definition of Done


dmitry-timofeev:
\+3.6% with a single line :-)

PR #534, opened 2018-10-17
Implemented Transaction Message. [ECR-2377]

bullet-tooth:
## Overview
Transaction Message builder implementation

---
See: https://jira.bf.local/browse/ECR-2377

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/BinaryTransactionMessage.java
```diff
@@ -0,0 +1,103 @@
...
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+/**
+ * Binary implementation of the {@link TransactionMessage} class.
+ */
+public final class BinaryTransactionMessage implements TransactionMessage {
+  private final ByteBuffer rawTransaction;
+
+  BinaryTransactionMessage(ByteBuffer rawTransaction) {
```
DENMROOT: should constructor be public as the whole class ?

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/BinaryTransactionMessage.java
```diff
@@ -0,0 +1,103 @@
...
+  public byte[] getPayload() {
+    int payloadSize = rawTransaction.limit() - (PAYLOAD_OFFSET + SIGNATURE_SIZE);
+    byte[] payload = new byte[payloadSize];
+    rawTransaction.get(payload, PAYLOAD_OFFSET, payloadSize);
+    return payload;
+  }
+
+  @Override
+  public HashCode hash() {
+    return defaultHashFunction().hashBytes(toBytes());
```
DENMROOT: We would duplicate transaction each time toBytes() would be invoked ?

exonum-java-binding-common/pom.xml
```diff
@@ -88,6 +88,10 @@
       <artifactId>equalsverifier</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
```
DENMROOT: Why do we need it ? 
if I understand correctly you add it to use assertThat() ? But we already use assertThat() from  org.hamcrest.MatcherAssert.assertThat  in common tests

exonum-java-binding-common/src/test/java/com/exonum/binding/common/message/TransactionMessageTest.java
```diff
@@ -0,0 +1,110 @@
...
+  private static final CryptoFunction CRYPTO = CryptoFunctions.ed25519();
+  private static final KeyPair KEYS = CRYPTO.generateKeyPair();
+
+  @Test
+  void noProperlyFilledMessagesTest() {
+    short serviceId = 1;
+    short transactionId = 1;
+    byte[] payload = Bytes.bytes();
+
+    assertThrows(NullPointerException.class,
```
DENMROOT: Seems like a good place for parametrized test, you assert same thing all the time ?

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,174 @@
...
+   * Creates Transaction Message from the given bytes buffer.
+   */
+  static TransactionMessage fromBuffer(ByteBuffer buffer) {
+    return BinaryTransactionMessage.fromBuffer(buffer);
+  }
+
+  /**
+   * Transaction Message Builder class.
+   */
+  class Builder {
```
DENMROOT: Why do you make it like inner classes in interface ? Looks confusing to me that I need to use interface static builder to get one of interface implementations .

bullet-tooth: as for me, it's ok. For example `java.util.stream.Stream#builder` has similar code. 
We can declare `Builder` as an interface and create an implementation class in a separate file. Or just move it to another file.

exonum-java-binding-common/src/test/java/com/exonum/binding/common/message/TransactionMessageTest.java
bullet-tooth: yes, I considered that. But then the parameter will be `Executable`, not  `TransactionMessage`. Because the exception occurs while`sign`.

CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `StandardSerializers` now supports `bool`, `fixed32`, `uint32`, `sint32`, 
   `fixed64`, `uint64`, `sint64`, `float` and `double` primitive types, 
   `PrivateKey`, `PublicKey` and `bytes` serialization. (#514, #523)
+- Multiproofs support in `ProofMapIndexProxy`. (#507)
```
dmitry-timofeev: No multiproof --- moved ^

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,174 @@
...
+  int TAG_OFFSET = 33;
+  int SERVICE_ID_OFFSET = 34;
+  int TRANSACTION_ID_OFFSET = 36;
+  int PAYLOAD_OFFSET = 38;
+
+  int AUTHOR_PUBLIC_KEY_SIZE = 32;
+  int SIGNATURE_SIZE = 64;
+
+  /**
+   * Returns a public key of the Transaction Message's author.
```
dmitry-timofeev: Here and elsewhere, I'd suggest to not capitalize the individual words that comprise the interface name when they refer not to the interface, but to the thing it represents (a transaction message).

I.e., "Returns a public key of the author of the transaction message".

Also, if I remember correctly, `’s` is used with people or organizations to specify possession.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,174 @@
...
+   */
+  PublicKey getAuthor();
+
+  /**
+   * Returns service id of the Transaction Message.
```
dmitry-timofeev:
```suggestion
   * Returns the identifier of the service this message belongs.
```

exonum-java-binding-common/src/test/java/com/exonum/binding/common/message/TransactionMessageTest.java
```diff
@@ -0,0 +1,111 @@
...
+import com.exonum.binding.test.Bytes;
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+class TransactionMessageTest {
+
+  private static final CryptoFunction CRYPTO = CryptoFunctions.ed25519();
```
dmitry-timofeev: Shan't we prefer a mock of cryptofunction to verify the behaviour of the builder?
```java
var message = builder.build()

assertThat(message.getSignature(), equalTo(expectedSignature));
verify(cryptoFunction).sign(eq(/* slice that must be signed */))
```

exonum-java-binding-common/src/test/java/com/exonum/binding/common/message/TransactionMessageTest.java
```diff
@@ -0,0 +1,111 @@
...
+  private static final KeyPair KEYS = CRYPTO.generateKeyPair();
+
+  @Test
+  void noProperlyFilledMessagesTest() {
```
dmitry-timofeev: I'd consider if we can reduce this test matrix :-) @dip56 WDYT?

exonum-java-binding-common/src/test/java/com/exonum/binding/common/message/TransactionMessageTest.java
```diff
@@ -0,0 +1,111 @@
...
+            .sign(keys, CRYPTO)
+    );
+  }
+
+  @ParameterizedTest
+  @MethodSource("source")
+  void roundTripTest(TransactionMessage message) {
+    byte[] bytes = message.toBytes();
+    TransactionMessage actualMessage = TransactionMessage.fromBytes(bytes);
+    assertThat(actualMessage, is(message));
```
dmitry-timofeev: It's OK to have this test that basically verifies that `ByteBuffer#equals` and our `equals`
work correctly, but it is not enough to verify that BinaryTransactionMessage properly takes out
parts of message and converts them to proper objects. Each method that parses (on the fly!)
a transaction buffer must be tested in BinaryTransactionMessageTest.

exonum-java-binding-common/src/test/java/com/exonum/binding/common/message/BinaryTransactionMessageTest.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ *
+ */
+
+package com.exonum.binding.common.message;
+
+import nl.jqno.equalsverifier.EqualsVerifier;
+import org.junit.jupiter.api.Test;
+
+class BinaryTransactionMessageTest {
```
dmitry-timofeev: Given the number of methods, I believe that's not enough to confirm it works
and ensure it does not break.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,174 @@
...
+   */
+  short getServiceId();
+
+  /**
+   * Returns transaction id of the Transaction Message.
```
dmitry-timofeev: What is transaction id? How is that useful to anyone? Please give these javadocs some love.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,174 @@
...
+   */
+  short getTransactionId();
+
+  /**
+   * Returns payload of the Transaction Message.
```
dmitry-timofeev: What is payload?

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,174 @@
...
+   */
+  byte[] getPayload();
+
+  /**
+   * Returns Transaction Message hash.
```
dmitry-timofeev: A hash of what exactly?

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,174 @@
...
+   */
+  HashCode hash();
+
+  /**
+   * Returns Transaction Message signature.
```
dmitry-timofeev: A signature of what? Is the format specified?

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,174 @@
...
+   */
+  byte[] getSignature();
+
+  /**
+   * Returns Transaction Message in binary format.
+   */
+  byte[] toBytes();
+
+  /**
+   * Returns the Builder for the Transaction Message.
```
dmitry-timofeev: Creates a new builder of the transaction message?

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,174 @@
...
+   */
+  static Builder builder() {
+    return new Builder();
+  }
+
+  /**
+   * Creates Transaction Message from the given bytes array.
```
dmitry-timofeev: I wonder if we shall validate the format somehow on instantiation (in `BTM`) :thinking:

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,174 @@
...
+     * Sets payload to the transaction message.
+     */
+    public Builder payload(ByteBuffer payload) {
+      this.payload = payload.duplicate().order(ByteOrder.LITTLE_ENDIAN);
+      return this;
+    }
+
+    /**
+     * Signs and creates an instance of the transaction message.
+     *
```
dmitry-timofeev: `@param` will be useful here

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,174 @@
...
+     * @throws NullPointerException if serviceId or transactionId or payload haven't set
+     * @throws IllegalArgumentException if public key has wrong size
+     */
+    public TransactionMessage sign(KeyPair keys, CryptoFunction crypto) {
```
dmitry-timofeev: Doc: Where do keys end up? Will it eat my secret key?

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,174 @@
...
+      checkNotNull(serviceId);
```
dmitry-timofeev: Something like `checkRequiredFieldSet(field: Object, fieldName: String)` so that clients do not get plain NPEs, but, say, IllegalStateException with message that explains things?

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/BinaryTransactionMessage.java
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Binary implementation of the {@link TransactionMessage} class.
+ */
+public final class BinaryTransactionMessage implements TransactionMessage {
+  private final ByteBuffer rawTransaction;
+
+  private BinaryTransactionMessage(ByteBuffer rawTransaction) {
+    this.rawTransaction = rawTransaction.duplicate().order(ByteOrder.LITTLE_ENDIAN);
+    this.rawTransaction.position(0);
```
dmitry-timofeev: It's not OK to do with a ByteBuffer *unless* you make it a precondition
that a buffer has zero position **and** enforce that. But that limits the applicability
of this class.

Therefore, I'd recommend an alternative that is to slice a buffer:
```java
this.rawTransaction = rawTransaction.slice().order(…);
```

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/BinaryTransactionMessage.java
```diff
@@ -0,0 +1,103 @@
...
+  public byte[] getPayload() {
+    int payloadSize = rawTransaction.limit() - (PAYLOAD_OFFSET + SIGNATURE_SIZE);
+    byte[] payload = new byte[payloadSize];
+    rawTransaction.get(payload, PAYLOAD_OFFSET, payloadSize);
+    return payload;
+  }
+
+  @Override
+  public HashCode hash() {
+    return defaultHashFunction().hashBytes(rawTransaction.array());
```
dmitry-timofeev: That is broken:
1. ByteBuffer#array won't work for a readonly BB
2. ByteBuffer#array won't work for a direct BB
3. When it works, it will return **the whole** array, i.e., not just bytes between `bb.position()` and `bb.limit()`.

Please add regression tests for these cases.

I think this will work:
```java
    Hashing.sha256().newHasher()
        .putBytes(rawTransaction.duplicate())
        .hash();
```

Please note that `BB.duplicate()` **does not** duplicate the underlying byte storage,
it duplicates the wrapper around that storage, i.e., BB, so that it can have independent
marks from the original object.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/BinaryTransactionMessage.java
```diff
@@ -0,0 +1,103 @@
...
+    byte[] signature = new byte[SIGNATURE_SIZE];
+    int payloadSize = rawTransaction.limit() - (PAYLOAD_OFFSET + SIGNATURE_SIZE);
+    int offset = PAYLOAD_OFFSET + payloadSize;
+    rawTransaction.get(signature, offset, SIGNATURE_SIZE);
+    return signature;
+  }
+
+  @Override
+  public byte[] toBytes() {
+    return rawTransaction.duplicate().array();
```
dmitry-timofeev: This code won't always work for the same reasons as above + `duplicate` here does not do anything useful.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/BinaryTransactionMessage.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  public static BinaryTransactionMessage fromBuffer(ByteBuffer buffer) {
```
dmitry-timofeev: I thought of adding a delegating `+fromBytes(byte[])`, but there is already one in `TransactionMessage`. Do you think we shall expose this class to the outside world?

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/MessageType.java
```diff
@@ -0,0 +1,69 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.exonum.binding.common.message;
+
+/**
+ * Message types used in Exonum framework.
```
dmitry-timofeev: I’d probably keep this although it is not needed at the moment in case we will ever need to access some of the system messages, but make it package private for now.

dmitry-timofeev:
@bullet-tooth , would you please include this patch:
```
Index: exonum-java-binding-common/src/main/java/com/exonum/binding/common/crypto/AbstractKey.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- exonum-java-binding-common/src/main/java/com/exonum/binding/common/crypto/AbstractKey.java	(date 1539866590000)
+++ exonum-java-binding-common/src/main/java/com/exonum/binding/common/crypto/AbstractKey.java	(date 1539869964000)
@@ -48,7 +48,7 @@
   }
 
   /**
-   * Returns the length of this key.
+   * Returns the length of this key in bytes.
    */
   public int size() {
     return rawKey.length;
```

CHANGELOG.md
bullet-tooth: crap. I used github ui for resolving conflicts for the first time :)

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/BinaryTransactionMessage.java
bullet-tooth: good question. How about to remove this method and make the constructor package-private?

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    /**
+     * Signs the message, creating a new signed binary transaction message.
+     *
+     * @param keys key pair with private and public keys. Public key is used as an author key of the
+     *        message and private key is used for signing the message.
+     * @param crypto a cryptographic function to use
+     * @return a new signed binary transaction message
+     * @throws NullPointerException if serviceId or transactionId or payload haven't set
```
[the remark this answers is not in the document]
MakarovS: Sorry, I reviewed this yesterday during Github outage and I got 405 when trying to submit this review and now some comments are duplicated. I'll delete those.

exonum-java-testing/src/main/java/com/exonum/binding/test/Bytes.java
```diff
@@ -100,5 +103,19 @@ public static String toHexString(byte[] bytes) {
...
 
-  private Bytes() {}
+  /**
+   * Creates a new array with random bytes of the given size.
+   *
+   * @param size a size of the new array
+   * @return a new array that contains random bytes
+   */
+  public static byte[] randomBytes(int size) {
+    checkArgument(0 <= size, "Size is negative: %s", size);
```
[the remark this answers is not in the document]
bullet-tooth: size can be 0. Because `nonpositive` is not appropriate here as for me.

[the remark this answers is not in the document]
vitvakatu: But operator here is `<=`, hence it cannot be zero?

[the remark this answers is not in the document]
dmitry-timofeev: checkArgument will throw if the condition is false, i.e. if size is less than zero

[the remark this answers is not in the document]
bullet-tooth: Added test for that

exonum-java-testing/src/test/java/com/exonum/binding/test/BytesTest.java
```diff
@@ -47,4 +51,16 @@ void toHexStringAllHexNumbersUpper() {
       assertThat(Bytes.toHexString(bytes(b)), equalTo(expected));
     }
   }
+
+  @ParameterizedTest
+  @ValueSource(ints = {-1024, -1, 0, 1, 1024})
```
dmitry-timofeev: Those are two different tests: for invalid sizes and for valid sizes.

exonum-java-testing/src/main/java/com/exonum/binding/test/Bytes.java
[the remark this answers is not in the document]
MakarovS: My mistake! Thanks for the test 👍

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,200 @@
...
+   */
+  PublicKey getAuthor();
+
+  /**
+   * Returns the identifier of the service this message belongs to.
+   */
+  short getServiceId();
+
+  /**
+   * Returns transaction identifier which is unique within the service.
```
dmitry-timofeev: Returns the transaction type identifier which is unique within the service.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,200 @@
...
+   */
+  short getTransactionId();
+
+  /**
+   * Returns the transaction message body.
```
dmitry-timofeev: Returns the payload containing the serialized transaction parameters.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,200 @@
...
+   */
+  byte[] getPayload();
+
+  /**
+   * Returns the transaction message hash.
```
dmitry-timofeev: Returns the SHA-256 hash of the binary message representation.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,200 @@
...
+   */
+  byte[] toBytes();
+
+  /**
+   * Returns the transaction message size in bytes.
+   */
+  int size();
+
+  /**
+   * Create a new builder for the transaction message.
```
dmitry-timofeev: Creates

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,200 @@
...
+   */
+  static TransactionMessage fromBytes(byte[] bytes) {
+    return new BinaryTransactionMessage(bytes);
+  }
+
+  /**
+   * Creates the transaction message from the given bytes buffer.
+   */
+  static TransactionMessage fromBuffer(ByteBuffer buffer) {
+    return new BinaryTransactionMessage(buffer.array());
```
dmitry-timofeev: This is broken for the reasons described earlier. Please pass the buffer to the message, so that it works with:
* direct,
* readonly
* with any marks set (e.g., pos=10, lim=128)

byte buffers.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,200 @@
...
+     */
+    public Builder payload(byte[] payload) {
+      return payload(ByteBuffer.wrap(payload));
+    }
+
+    /**
+     * Sets payload to the transaction message.
+     */
+    public Builder payload(ByteBuffer payload) {
+      this.payload = payload.duplicate().order(ByteOrder.LITTLE_ENDIAN);
```
dmitry-timofeev: Order is irrelevant here — it is just some bytes.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
```diff
@@ -0,0 +1,200 @@
...
+     * @throws NullPointerException if serviceId or transactionId or payload weren't set
+     * @throws IllegalArgumentException if public key has wrong size
+     */
+    public TransactionMessage sign(KeyPair keys, CryptoFunction crypto) {
+      checkRequiredFieldsSet();
+      PublicKey authorPublicKey = keys.getPublicKey();
+      checkArgument(authorPublicKey.size() == AUTHOR_PUBLIC_KEY_SIZE);
+
+      ByteBuffer buffer = ByteBuffer
+          .allocate(MIN_MESSAGE_SIZE + payload.limit())
```
dmitry-timofeev: This code is incorrect: payload.limit() is not always the number of bytes that would be transferred in
ByteBuffer#put(ByteBuffer). 

Please add a test for that.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/BinaryTransactionMessage.java
```diff
@@ -0,0 +1,119 @@
...
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+
+/**
+ * Binary implementation of the {@link TransactionMessage} class.
+ */
+public final class BinaryTransactionMessage implements TransactionMessage {
+  private final ByteBuffer rawTransaction;
+
+  BinaryTransactionMessage(byte[] bytes) {
```
dmitry-timofeev: Why don't it work with ByteBuffers, requiring a copy of byte array, or, if ByteBuffer is given, two copies?

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/BinaryTransactionMessage.java
```diff
@@ -0,0 +1,119 @@
...
+    checkArgument(MIN_MESSAGE_SIZE <= bytes.length,
+        "Expected an array of size at least %s, but was %s", MIN_MESSAGE_SIZE, bytes.length);
+    this.rawTransaction = ByteBuffer.wrap(copyOf(bytes, bytes.length)).order(LITTLE_ENDIAN);
+  }
+
+  @Override
+  public PublicKey getAuthor() {
+    byte[] key = new byte[AUTHOR_PUBLIC_KEY_SIZE];
+    rawTransaction.position(AUTHOR_PUBLIC_KEY_OFFSET);
```
dmitry-timofeev: Why are relative operations used, that involve two steps and modify the bytebuffer marks instead of absolute operations that used to be here?

exonum-java-binding-common/src/test/java/com/exonum/binding/common/message/BinaryTransactionMessageTest.java
```diff
@@ -0,0 +1,121 @@
...
+    // mutate input parameter
+    mutate(mutableIn);
+    assertThat(message.getPayload(), not(mutableIn));
+
+    // mutate output parameters
+    byte[] mutablePayload = message.getPayload();
+    mutate(mutablePayload);
+    assertThat(message.getPayload(), not(mutablePayload));
+
+    byte[] mutableAuthor = message.getAuthor().toBytes();
```
dmitry-timofeev: PublicKey is not mutable

exonum-java-binding-common/src/test/java/com/exonum/binding/common/message/BinaryTransactionMessageTest.java
```diff
@@ -0,0 +1,121 @@
...
+    mutate(mutableAuthor);
+    assertThat(message.getAuthor().toBytes(), not(mutableAuthor));
+
+    byte[] mutableSignature = message.getSignature();
+    mutate(mutableSignature);
+    assertThat(message.getSignature(), not(mutableSignature));
+
+    byte[] mutableHash = message.hash().asBytes();
```
dmitry-timofeev: HashCode is not mutable

exonum-java-binding-common/src/test/java/com/exonum/binding/common/message/BinaryTransactionMessageTest.java
```diff
@@ -0,0 +1,121 @@
...
+  @Test
+  void equalsTest() {
+    EqualsVerifier
+        .forClass(BinaryTransactionMessage.class)
+        .verify();
+  }
+
+  @Test
+  void immutabilityTest() {
+    byte[] mutableIn = Bytes.bytes(0x00, 0x01, 0x02);
```
dmitry-timofeev: mutablePayload

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/BinaryTransactionMessage.java
bullet-tooth: Because `ByteByffer` doesn't provide a method for reading byte array using absolute position.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/BinaryTransactionMessage.java
```diff
@@ -0,0 +1,128 @@
...
+    byte[] payload = new byte[payloadSize];
+    rawTransaction.position(PAYLOAD_OFFSET);
+    rawTransaction.get(payload);
+    return payload;
+  }
+
+  @Override
+  public HashCode hash() {
+    return sha256().newHasher()
+        .putBytes(rawTransaction.duplicate())
```
dmitry-timofeev: It is broken as the code that gets byte arrays modifies position of the original buffer, therefore, the bytes that get hashed depend on the invocations of `getPayload`/`getSignature`/`getAuthor`. 

Can we fix it so that the implementation is more reliable? E.g., as the bb is internally allocated and guaranteed to have an array corresponding to the full message, hash the array (with a comment clarifying this implementation)?

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/TransactionMessage.java
bullet-tooth: I see, thanks. Fixed

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/BinaryTransactionMessage.java
bullet-tooth: fixed

dmitry-timofeev: Please add tests for `hash` — there are currently none.

PR #540, opened 2018-10-24
Update versions

dmitry-timofeev:
## Overview
- Update project version to 0.3 to create a release branch
- Update project version to 0.4-SNAPSHOT for the next iteration

---

### Definition of Done


dmitry-timofeev:
@MakarovS , thanks, updated to `${version}` so that it is no longer needed in the future.

dmitry-timofeev:
I used the wrong property — `${version}` is the version of the project-to-be-generated, therefore, we have to specify a version in the pom template (as is) or pull it from a property file.

Rebased both commits again

PR #545, opened 2018-10-26
Reorder entries according to the end-user significance.

dmitry-timofeev:
### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,11 +18,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [0.3] - TBA <!-- somewhere next week -->
 
 ### Added
-- `Message.Builder#setBody(byte[])` to avoid `ByteBuffer.wrap` in the client code. (#401)
-- `MapIndex.isEmpty()` method to check if MapIndex is empty. (#445)
 - Flat map proofs support, including proofs of absence and multiproofs — proofs for several
   entries at once. (#250, #507, #532)
-- Wallet transactions history support to the cryptocurrency-demo. (#481)
 - `StandardSerializers` now supports `bool`, `fixed32`, `uint32`, `sint32`, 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -18,24 +18,23 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 - Static factory methods accepting protobuf messages to collections,
   allowing to pass Protocol Buffer messages directly instead of using
   `StandardSerializers#protobuf`. (#505)
+- `Message.Builder#setBody(byte[])` to avoid `ByteBuffer.wrap` in the client code. (#401)
+- `MapIndex.isEmpty()` method to check if MapIndex is empty. (#445)
+- Wallet transactions history support to the cryptocurrency-demo. (#481)
 
 ### Changed
-- `Transaction#execute` can throw `TransactionExecutionException` to roll back 
+-  `Transaction#execute` can throw `TransactionExecutionException` to roll back
```
MakarovS: Why this space?

dmitry-timofeev:
@exonum/enterprise ,

I believe the requirement of (reverse) chronological order applies to entries as sections corresponding to project versions. It is also reasonable to use (just) chronological ordering for each individual change. However, one must not follow the latter _recommendation_ blindly but remember the principles of this standard: "Changelogs are for humans, not machines.", "The same types of changes should be grouped.". Therefore, if one makes a change to something that was added in an unreleased revision, just update an already present description in the changlelog. If you add a similar feature (e.g., Serializer#bool & Serializer#protobuf) — group them together.

dmitry-timofeev:
On top of that, with my highlight:

> The purpose of a changelog entry is to document the noteworthy difference, often *across multiple commits*, to communicate them clearly to end users. 

If it takes us several commits (= PRs) to implement a feature, there is still a single feature, not several. Therefore, a single changelog entry.

PR #548, opened 2018-10-30
Implemented Core Schema skeleton [ECR-2541]

bullet-tooth:
## Overview

---
Impemended java part for core schema skeleton.

See: https://jira.bf.local/browse/ECR-2541

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -0,0 +1,71 @@
...
+package com.exonum.binding.blockchain;
+
+import com.exonum.binding.common.hash.HashCode;
+import com.exonum.binding.storage.database.View;
+import com.exonum.binding.storage.indices.ListIndex;
+import com.exonum.binding.storage.indices.ProofListIndexProxy;
+import com.google.common.annotations.VisibleForTesting;
+
+/**
+ * Provides read-only access to the blockchain state
```
dmitry-timofeev: Not "informational".
```
Provides read-only access to the blockchain state that is maintained by Exonum core:
blocks, transaction messages, execution results.
```

I'd also link https://docs.rs/exonum/0.9.4/exonum/blockchain/struct.Schema.html

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -0,0 +1,71 @@
...
+    CoreSchemaProxy coreSchema = CoreSchemaProxy.newInstance(view);
+    return new Blockchain(coreSchema);
+  }
+
+  /**
+   * Returns the height of the latest committed block.
+   *
+   * @throws RuntimeException if the "genesis block" was not created
+   */
+  public long getHeight() {
```
dmitry-timofeev: Which height does the genesis block have? The first block? — I'd clarify in the docs.

Is it equal to the number of blocks? — if not, I'd add a warning

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -0,0 +1,71 @@
...
+    return schema.getHeight();
+  }
+
+  /**
+   * Returns an list index containing a block hash for every block height
```
dmitry-timofeev:
```
Returns a list of all block hashes, indexed by the block height.
For example, the genesis block will be at index **?**, 
the block at height {@code h} — at index **?**.
```

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -0,0 +1,71 @@
...
+   * (represented by list index id).
+   */
+  public ListIndex<HashCode> getAllBlockHashes() {
+    return schema.getAllBlockHashes();
+  }
+
+  /**
+   * Returns an proof list index containing block hashes for the given height.
```
dmitry-timofeev: This javadoc is incorrect.

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -0,0 +1,71 @@
...
+   */
```
dmitry-timofeev: `@param` (what is a valid range of values one can pass), `@throws` — what if I don't get it right?

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/CoreSchemaProxy.java
```diff
@@ -0,0 +1,103 @@
...
+  private CoreSchemaProxy(NativeHandle nativeHandle, View dbView) {
+    this.nativeHandle = nativeHandle;
+    this.dbView = dbView;
+  }
+
+  /**
+   * Constructs a schema proxy for a given dbView.
+   */
+  static CoreSchemaProxy newInstance(View dbView) {
+    long nativeCreate = nativeCreate(dbView.getViewNativeHandle());
```
dmitry-timofeev: This var name is off.

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/CoreSchemaProxy.java
```diff
@@ -0,0 +1,103 @@
...
+        CoreIndex.BLOCK_TRANSACTIONS, id, dbView, StandardSerializers.hash());
+  }
+
+  private static native long nativeCreate(long viewNativeHandle);
+
+  private static native void nativeFree(long nativeHandle);
+
+  private static native long nativeGetHeight(long nativeHandle);
+
+  private static native byte[] nativeGetLastBlock(long nativeHandle);
```
dmitry-timofeev: `@SuppressWarnings("unused") // Will be done in the next task`

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/CoreSchemaProxy.java
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * Mapping for Exonum core indexes by name.
+   */
+  private static final class CoreIndex {
+    private static final String BLOCK_TRANSACTIONS = "block_transactions";
```
dmitry-timofeev: These names do not seem correct — see the Schema source (some prefix is missing)

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/CoreSchemaProxy.java
```diff
@@ -0,0 +1,103 @@
...
+  private static native void nativeFree(long nativeHandle);
+
+  private static native long nativeGetHeight(long nativeHandle);
+
+  private static native byte[] nativeGetLastBlock(long nativeHandle);
+
+  /**
+   * Mapping for Exonum core indexes by name.
+   */
+  private static final class CoreIndex {
```
dmitry-timofeev: Do these constants belong in a separate inner class?

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/BlockchainCollectionsTest.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  @Ignore
+  @Test(expected = UnsupportedOperationException.class)
+  public void getAllBlockHashesShouldBeReadOnly() {
+    when(mockSchema.getAllBlockHashes()).thenReturn(mock(ListIndexProxy.class));
+
+    ListIndex<HashCode> list = blockchain.getAllBlockHashes();
+
+    list.add(HashCode.fromInt(0x0));
```
dmitry-timofeev: It will do whatever the mock will do. I'd remove these tests, as they don't work.

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/CoreSchemaProxyIntegrationTest.java
```diff
@@ -0,0 +1,78 @@
...
+import org.junit.Ignore;
+import org.junit.Test;
+
+public class CoreSchemaProxyIntegrationTest {
+
+  static {
+    LibraryLoader.load();
+  }
+
+  @Ignore
```
dmitry-timofeev: I'd move to the class level too and explain *why* they are ignored.

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/CoreSchemaProxyIntegrationTest.java
```diff
@@ -0,0 +1,78 @@
...
+  @Test
+  @RequiresNativeLibrary
```
dmitry-timofeev: Please move the annotation to the class level.

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/CoreSchemaProxyIntegrationTest.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }
+
+  @Ignore
+  @Test
+  @RequiresNativeLibrary
+  public void getBlockTransactionsTest() throws CloseFailuresException {
+    try (MemoryDb db = MemoryDb.newInstance();
+        Cleaner cleaner = new Cleaner()) {
+      Snapshot view = db.createSnapshot(cleaner);
```
dmitry-timofeev: https://jira.bf.local/browse/ECR-2634

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
bullet-tooth: Which version should be used in a link `0.9.4` or `latest`? 🤔

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/CoreSchemaProxy.java
bullet-tooth: Will them be needed somewhere outside the Core Schema? If yes, i would move the class to separate file

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/BlockchainCollectionsTest.java
bullet-tooth: My suggestion was that we'll implement unmodifiable collections and use them in the `Blockchain`  class. And after that, just remove `@Ignore` and these tests will work correctly.

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
vitvakatu: I think `latest`. This API will be either supported by core in future versions, or we will change our API (together with this docstring) later

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
vitvakatu: I still cannot understand the docs. Is blockchain height equals to zero before and after genesis block was commited?

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/CoreSchemaProxy.java
vitvakatu: I'd better use `TODO` for easier searching

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -0,0 +1,79 @@
...
+   * @throws RuntimeException if the "genesis block" was not created
+   */
+  public long getHeight() {
+    return schema.getHeight();
+  }
+
+  /**
+   * Returns a list of all block hashes, indexed by the block height.
+   * For example, the "genesis block" will be at index 0,
+   * the block at height {@code h = 10} — at index **10**.
```
MakarovS: Why `**10**`?

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
bullet-tooth: No. Zero is just after genesis block. RuntimeException will be thrown before the genesis block.

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/CoreSchemaProxy.java
dmitry-timofeev: I mean why are they in a separate class, not in `CoreSchemaProxy`. 

I do not expect them to be used anywhere else.

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/BlockchainCollectionsTest.java
dmitry-timofeev: They will work correctly **iff** we implement that in a particular way (i.e., if we wrap whatever `CoreSchemaProxy` returns in an unmodifiable decorator). If it is implemented differently (e.g., `CoreSchemaProxy` returns unmodifiable things, or if the core treats the collections in this namespace as unmodifiable), these tests will not work.

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -0,0 +1,79 @@
...
+
+import com.exonum.binding.common.hash.HashCode;
+import com.exonum.binding.storage.database.View;
+import com.exonum.binding.storage.indices.ListIndex;
+import com.exonum.binding.storage.indices.ProofListIndexProxy;
+import com.google.common.annotations.VisibleForTesting;
+
+/**
+ * Provides read-only access to the blockchain state that is maintained by Exonum core:
+ * blocks, transaction messages, execution results. Please refer to the
```
dmitry-timofeev: I'd reword that "This class is equivalent to (or provides access to) the blockchain::Schema in the Rust API."
 because it's not OK to refer Java users to that documentation unconditionally — it won't make any sense to them.

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -0,0 +1,79 @@
...
+   * Constructs a new blockchain instance for the given database view.
+   */
+  public static Blockchain newInstance(View view) {
+    CoreSchemaProxy coreSchema = CoreSchemaProxy.newInstance(view);
+    return new Blockchain(coreSchema);
+  }
+
+  /**
+   * Returns the height of the latest committed block in the blockchain.
+   * Height is a number of blocks in the blockchain starting from {@code h = 0} zero block,
```
dmitry-timofeev: The wording is unclear — I do not understand "number of blocks starting from zero block".

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -0,0 +1,79 @@
...
+   * For example, the "genesis block" will be at index 0,
+   * the block at height {@code h = 10} — at index **10**.
+   * The last committed block will be at height {@code h = getAllBlockHashes().size() - 1}.
+   */
+  public ListIndex<HashCode> getAllBlockHashes() {
+    return schema.getAllBlockHashes();
+  }
+
+  /**
+   * Returns a proof list of transaction hashes within the block by the given height.
```
dmitry-timofeev: … committed in the block at the given height.

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -0,0 +1,79 @@
...
+   *
+   * @param height block height starting from 0
+   * @throws IllegalArgumentException if the height is negative
+   */
+  public ProofListIndexProxy<HashCode> getBlockTransactions(long height) {
```
dmitry-timofeev: What if I use a non-existent height? Will I get an empty list? Or an exception?

dmitry-timofeev:
This task also included updates to the QA service and its spec. Do you prefer to do that in a separate PR?

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/CoreSchemaProxy.java
bullet-tooth: @dmitry-timofeev, to make them self-documented with some concise name. If move them to `CoreSchemaProxy` then names should be like `INDEX_NAME_BLOCK_TRANSACTIONS` or something...

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
bullet-tooth: described in javadoc

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
MakarovS: Why don't we leave possibility for a user to go to the documentation of Rust `blockchain::Schema` if we refer to it anyway? Right now this javadoc doesn't seem clear to me. I think it's better to have at least some information than none.
Or maybe we should refer to `CoreSchemaProxy` here?

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - Support of new transaction message format added in Exonum 0.10. (#534)
+- Support of core schema API (#548, #549)
```
MakarovS: A dot `.` 🙀

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
dmitry-timofeev: I might not have been clear: it's good to have a link to the core schema, but the wording of the second sentence is better be improved.

>  Right now this javadoc doesn't seem clear to me.

Please request improving it (ideally, with suggestions) — it must be clear :upside_down_face:

MakarovS: I think now it's okay, but I'd still add a link to it:
```
 * Provides read-only access to the subset of <a href="https://docs.rs/exonum/latest/exonum/blockchain/struct.Schema.html">blockchain::Schema</a> features in the Core API:
 * blocks, transaction messages, execution results.
```
Something like that?

Issue #446, opened 2018-08-31
Service::after_commit support


vitvakatu:

MakarovS:
Implemented in https://github.com/exonum/exonum-java-binding/pull/550 and https://github.com/exonum/exonum-java-binding/pull/572

PR #550, opened 2018-10-31
Add afterCommit [ECR-2537]

MakarovS:
## Overview
Add afterCommit method to services and BlockCommittedEvent structure.

---
See: 
- #446 
- ECR-2537

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommitedEventImpl.java
```diff
@@ -0,0 +1,55 @@
...
+public class BlockCommitedEventImpl implements BlockCommittedEvent {
+
+  private int validatorId;
+  private long height;
+  private Snapshot snapshot;
+
+  /**
+   * Creates a new block commited event.
+   * @param snapshot a snapshot of the blockchain state
+   * @param validatorId a validator id. Negative if this node is not a validator
```
bullet-tooth: what is 0 for `validatorId`?

exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommitedEventImpl.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  @Override
+  public Snapshot getSnapshot() {
+    return snapshot;
+  }
+
+  @Override
+  public Optional<Integer> getValidatorId() {
+    return validatorId > 0 ? Optional.of(validatorId) : Optional.empty();
```
bullet-tooth:
```suggestion
    return Optional.of(validatorId).filter(validatorId > 0);
```

exonum-java-binding-core/src/main/java/com/exonum/binding/service/Service.java
```diff
@@ -118,4 +118,12 @@
    * @param router a router responsible for handling requests to this service
    */
   void createPublicApiHandlers(Node node, Router router);
+
+  /**
+   * Handles block commit event. This handler is invoked for each service after commit of the block.
```
bullet-tooth: ...This handler is a callback method which is invoked by the blockchain? after each block commit.

exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommitedEventImpl.java
MakarovS: 0 seems like a valid `validatorId`, in core it has `u16` type.
Why do you think `Integer` is better?

vitvakatu: 0 is definitely valid validatorID

bullet-tooth: I supposed that because `Integer` can be null. The logic is a little bit messy here, 0 is valid `validatorId` but we filter it in `getValidatorId` method...

exonum-java-binding-core/src/main/java/com/exonum/binding/service/adapters/UserServiceAdapter.java
```diff
@@ -139,6 +141,24 @@ public void mountPublicApiHandler(long nodeNativeHandle) {
     server.mountSubRouter(serviceApiPath(), router);
   }
 
+  /**
+   * Handles block commit event. This handler is invoked after commit of the block.
```
dmitry-timofeev: Tests?

exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommitedEventImpl.java
```diff
@@ -0,0 +1,55 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.exonum.binding.service;
+
+import com.exonum.binding.storage.database.Snapshot;
+import java.util.Optional;
+
+public class BlockCommitedEventImpl implements BlockCommittedEvent {
```
dmitry-timofeev: final

exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommitedEventImpl.java
```diff
@@ -0,0 +1,55 @@
...
+
+  private int validatorId;
```
dmitry-timofeev: final

exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommitedEventImpl.java
dmitry-timofeev: I'd replace with `@AutoValue` implementation (this particular implementation is a value) — so that it gets `#toString`.

exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommitedEventImpl.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  @Override
+  public Snapshot getSnapshot() {
+    return snapshot;
+  }
+
+  @Override
+  public Optional<Integer> getValidatorId() {
+    return Optional.of(validatorId).filter(v -> v > 0);
```
dmitry-timofeev: This logic does not belong here, neither in this method nor in this class.
The fact that validator id is negative is just an implementation detail of `UserServiceAdapter`. It's its responsibility to decode any internal types used to transfer values from native to Java to appropriate Java types.

A side question is why does the framework communicate the type of node through this event :thinking: (@DarkEld3r, @alekseysidorov )?

exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommitedEventImpl.java
```diff
@@ -0,0 +1,55 @@
...
+    this.snapshot = snapshot;
+  }
+
+  @Override
+  public Snapshot getSnapshot() {
+    return snapshot;
+  }
+
+  @Override
+  public Optional<Integer> getValidatorId() {
```
dmitry-timofeev: OptionalInt

exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommittedEvent.java
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+package com.exonum.binding.service;
+
+import com.exonum.binding.storage.database.Snapshot;
+import java.util.Optional;
+
+/**
+ * The current node state on which the blockchain is running.
```
dmitry-timofeev: The blockchain state just after the corresponding block is committed? or the block at the corresponding height?

exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommittedEvent.java
```diff
@@ -0,0 +1,45 @@
...
+ * This structure is passed to the `afterCommit` method of the `Service` interface and is used
+ * for the interaction between service business logic and the current node state.
```
dmitry-timofeev: "and is used for the interaction between service business logic and the current node state."

I don't understand that.

exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommittedEvent.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package com.exonum.binding.service;
+
+import com.exonum.binding.storage.database.Snapshot;
+import java.util.Optional;
+
+/**
+ * The current node state on which the blockchain is running.
+ * This structure is passed to the `afterCommit` method of the `Service` interface and is used
```
dmitry-timofeev: backticks don't work here. Use {@link} or {@linkplane} or {@code} — here it makes sense to link the notification method.

exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommittedEvent.java
```diff
@@ -0,0 +1,45 @@
...
+ * for the interaction between service business logic and the current node state.
+ */
+public interface BlockCommittedEvent {
+
+  /**
+   * If the current node is a validator, returns its identifier.
```
dmitry-timofeev: I'd say "this node". "current" is less precise.

exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommittedEvent.java
```diff
@@ -0,0 +1,45 @@
...
+   * For other nodes return {@code Optional.empty()}.
```
dmitry-timofeev: there are no "other nodes". 

If this node is a node-auditor (link to the docs), it will return Optional.empty().

exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommittedEvent.java
```diff
@@ -0,0 +1,45 @@
...
+   */
+  Optional<Integer> getValidatorId();
+
+  /**
+   * Returns the current blockchain height. This height is "height of the last committed block".
```
dmitry-timofeev: the current blockchain height, which is the height of the last committed block.

exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommittedEvent.java
```diff
@@ -0,0 +1,45 @@
...
+   */
+  long getHeight();
+
+  /**
+   * Returns the current database snapshot. This snapshot is used to retrieve schema information
```
dmitry-timofeev: I'd replace "This snapshot is used to retrieve schema information from the database."
with "It is immutable and represents the database state as of the block
at the current height. (link getHeight)"

exonum-java-binding-core/src/main/java/com/exonum/binding/service/Service.java
```diff
@@ -118,4 +118,13 @@
    * @param router a router responsible for handling requests to this service
    */
   void createPublicApiHandlers(Node node, Router router);
+
+  /**
+   * Handles block commit event. This handler is a callback method which is invoked by
+   * the blockchain after each block commit.
+   * For example, a service can create one or more transactions if a specific condition
+   * has occurred.
+   * @param event the current node state on which the blockchain is running
```
dmitry-timofeev: the context allowing to access the blockchain state as of that committed block?

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/services/service/TestService.java
```diff
@@ -42,12 +43,24 @@
...
+
+  private int validatorId;
+
   @Inject
   public TestService(SchemaFactory<TestSchema> schemaFactory) {
     super(ID, NAME, (rawTx) -> PutValueTransaction.from(rawTx, schemaFactory));
     this.schemaFactory = schemaFactory;
   }
 
+  public long getBlockchainHeight() {
```
dmitry-timofeev: @vitvakatu, @skletsun Are such methods accessible in native (can you implement tests for afterCommit with these)?

If not, please suggest what will help.

exonum-java-binding-core/src/main/java/com/exonum/binding/service/Service.java
```diff
@@ -118,4 +118,13 @@
...
+   */
+  default void afterCommit(BlockCommittedEvent event) {}
```
DENMROOT: Is default without implementation to allow optional afterCommit()  implementation by the service ?

MakarovS: Yes! Some services might not need this, so I thought default method would be better rather than forcing the user to override it with empty implementation

exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommittedEvent.java
MakarovS: Ok, it's just the way it was written in core (as most of the comments here), but I agree that it's better to specify what "other nodes" are (as we only have two categories of full nodes  🙂 )

exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommittedEvent.java
MakarovS: This is basically a copy of `ServiceContext` doc from core. I've replaced `current node state` with `blockchain state` though, hope it's more clear now

exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommittedEvent.java
dmitry-timofeev: Yes, "other nodes" might be confused with "other validators".

exonum-java-binding-core/src/main/java/com/exonum/binding/service/Service.java
```diff
@@ -118,4 +118,13 @@
...
   void createPublicApiHandlers(Node node, Router router);
+
+  /**
+   * Handles block commit event. This handler is a callback method which is invoked by
+   * the blockchain after each block commit.
+   * For example, a service can create one or more transactions if a specific condition
+   * has occurred.
+   * @param event the context allowing to access the blockchain state as of that committed block
+   */
+  default void afterCommit(BlockCommittedEvent event) {}
```
dmitry-timofeev: As this method may be used for anything (it does not have a narrow purpose, as #convertTx or Tx#execute), I'd specify what will happen (or reassure that nothing bad will) if it throws some exception.

exonum-java-binding-core/src/main/java/com/exonum/binding/service/BlockCommittedEvent.java
```diff
@@ -0,0 +1,47 @@
...
+  OptionalInt getValidatorId();
+
+  /**
+   * Returns the current blockchain height, which is the height of the last committed block.
+   */
+  long getHeight();
+
+  /**
+   * Returns the current database snapshot. It is immutable and represents the database state
+   * as of the block at the current height {@link #getHeight()}.
```
dmitry-timofeev: `{@linkplain #getHeight() height}`

exonum-java-binding-core/src/main/java/com/exonum/binding/service/Service.java
```diff
@@ -118,4 +118,13 @@
    * @param router a router responsible for handling requests to this service
    */
   void createPublicApiHandlers(Node node, Router router);
+
+  /**
+   * Handles block commit event. This handler is a callback method which is invoked by
+   * the blockchain after each block commit.
+   * For example, a service can create one or more transactions if a specific condition
+   * has occurred.
```
dmitry-timofeev: Please add as a separate paragraph a warning that this method is invoked synchronously from the thread that commits the block, therefore, implementations of this method must not perform any blocking or long-running operations.

exonum-java-binding-core/src/main/java/com/exonum/binding/service/Service.java
MakarovS: Do you think we should specify it for this method? I've added `read-only` and `optional` to this javadoc, I think it should imply that it's okay to throw here. If that's not enough, could you please propose the correct wording?

bullet-tooth: Mb something like that: 
`... it's safe to throw any exceptions inside the method and it won't cause transaction failing. `

MakarovS: Doesn't seem correct to me - what transaction are we speaking about? Maybe we should refer to `block commit` rather than `transaction`?

dmitry-timofeev: As it is invoked during block processing (afaik, in the same thread), I'd reassure that any exceptions will be swallowed and will not affect the processing of transactions/blocks.

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/services/service/TestService.java
skletsun: As for me - looks like it's enough for testing from the native side.

exonum-java-binding-core/src/test/java/com/exonum/binding/service/adapters/UserServiceAdapterTest.java
```diff
@@ -223,4 +224,53 @@ public void mountPublicApiHandler_FailsOnSubsequentCalls() {
...
+        .thenReturn(snapshot);
+    serviceAdapter.afterCommit(SNAPSHOT_HANDLE, VALIDATOR_ID, HEIGHT);
+
+    ArgumentCaptor<BlockCommittedEvent> ac = ArgumentCaptor.forClass(BlockCommittedEvent.class);
+    verify(service).afterCommit(ac.capture());
+
+    BlockCommittedEvent event = ac.getValue();
+
+    assertThat(event.getHeight(), equalTo(HEIGHT));
+    assertThat(event.getValidatorId(), equalTo(OptionalInt.of(VALIDATOR_ID)));
```
bullet-tooth: assert snapshot?

exonum-java-binding-core/src/test/java/com/exonum/binding/service/adapters/UserServiceAdapterTest.java
```diff
@@ -223,4 +224,53 @@ public void mountPublicApiHandler_FailsOnSubsequentCalls() {
     expectedException.expect(IllegalStateException.class);
     serviceAdapter.mountPublicApiHandler(0x0B);
   }
+
+  @Test
+  public void afterCommit_ValidatorNode() {
+    Snapshot snapshot = mock(Snapshot.class);
```
bullet-tooth: Just for your taste: how about to refactor it to class field `@Mock  Snapshot snapshot`?

PR #567, opened 2018-11-06
Add afterCommit to QaService [ECR-2552]

MakarovS:
## Overview
Add afterCommit to QaService.
Now it creates `after_commit_counter` on service initialization. This counter is incremented on every `afterCommit` invocation.

---
See: https://jira.bf.local/browse/ECR-2552

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding-qa-service/src/main/java/com/exonum/binding/qaservice/QaServiceImpl.java
```diff
@@ -166,6 +171,17 @@ public HashCode submitUnknownTx() {
...
 
+  /**
+   * Increments the initial counter so the number of times this method was invoked is stored in it.
+   */
+  @Override
+  public void afterCommit(BlockCommittedEvent event) {
+    long seed = 1L;
+    HashCode counterId = Hashing.sha256()
+        .hashString(INITIAL_COUNTER_NAME, StandardCharsets.UTF_8);
+    submitIncrementCounter(seed, counterId);
```
bullet-tooth: Won't it work forever? `Initialize -> new block -> after commit -> new block -> after commit -> new block`. Or mb I missunderstood something?

bullet-tooth: it'll spam the network and create the new transaction for each block.

exonum-java-binding-qa-service/src/main/java/com/exonum/binding/qaservice/QaServiceImpl.java
```diff
@@ -88,9 +94,8 @@ protected Schema createDataSchema(View view) {
 
   @Override
   public Optional<String> initialize(Fork fork) {
-    // Add a default counter to the blockchain.
-    String defaultCounterName = "default";
-    new CreateCounterTx(defaultCounterName)
+    // Add an initial counter to the blockchain.
+    new CreateCounterTx(INITIAL_COUNTER_NAME)
```
DENMROOT: What was the purpose of previously existed counter ? Can you create new separate counter for your purposes ? Just to avoid potential mess.

exonum-java-binding-qa-service/src/main/java/com/exonum/binding/qaservice/QaServiceImpl.java
MakarovS: Is new block created on every transaction?

bullet-tooth: Exonum produces blocks with parameterized constant frequency. If no transaction received then empty block generates. So, therefore, this method will create the transaction for each block 🤔

MakarovS: Sorry, I don't see why is that a bad thing. We allow services to submit transactions in `afterCommit` method. If they want to submit a transaction on every block commit - why shouldn't they?

vitvakatu: Every blockchain would work forever, won't it? :)

exonum-java-binding-qa-service/src/test/java/com/exonum/binding/qaservice/QaServiceImplIntegrationTest.java
```diff
@@ -272,6 +276,28 @@ void getValueBeforeInit() {
...
+      Fork view = db.createFork(cleaner);
+      service.initialize(view);
+
+      BlockCommittedEvent event = mock(BlockCommittedEvent.class);
+      service.afterCommit(event);
+
+      // Check that the initial counter was updated.
+      QaSchema schema = new QaSchema(view);
+      MapIndex<HashCode, Long> counters = schema.counters();
+      MapIndex<HashCode, String> counterNames = schema.counterNames();
```
vitvakatu: What is the reason behind having counter names?

MakarovS: Below we check that our `after_commit_counter` counter has correct name.
Or do you mean why we don't have HashCodes instead of names in QA service as a whole? That I don't know, probably it's just easier to create counters that way

exonum-java-binding-qa-service/src/main/java/com/exonum/binding/qaservice/QaServiceImpl.java
dmitry-timofeev: @dip56 :arrow_up: :warning: 

@bullet-tooth Are you saying that the system will never be idle, i.e., there will always be a transaction, triggering the next block as soon as `min_propose_timeout`? 

@MakarovS is there a setting that defines how many transactions there must be to use `min_propose_timeout`?

exonum-java-binding-qa-service/src/main/java/com/exonum/binding/qaservice/QaServiceImpl.java
```diff
@@ -89,8 +98,11 @@ protected Schema createDataSchema(View view) {
   @Override
   public Optional<String> initialize(Fork fork) {
     // Add a default counter to the blockchain.
-    String defaultCounterName = "default";
-    new CreateCounterTx(defaultCounterName)
+    new CreateCounterTx(DEFAULT_COUNTER_NAME)
```
dmitry-timofeev: I propose to extract that in `createCounter(name: String)` method to hide irrelevant implementation details — the fact that we create transaction is a little misleading because it is not a regular transaction, we just want to reuse the code in its `#execute`.

exonum-java-binding-qa-service/src/main/java/com/exonum/binding/qaservice/QaServiceImpl.java
```diff
@@ -166,6 +178,18 @@ public HashCode submitUnknownTx() {
     });
   }
 
+  /**
+   * Increments the afterCommit counter so the number of times this method was invoked is stored
+   * in it.
+   */
+  @Override
+  public void afterCommit(BlockCommittedEvent event) {
```
dmitry-timofeev: I'd consider moving it up to other methods defined in `Service`, and then have methods from `CryptocurrencyService`.

exonum-java-binding-qa-service/src/test/java/com/exonum/binding/qaservice/QaServiceImplIntegrationTest.java
```diff
@@ -272,6 +276,29 @@ void getValueBeforeInit() {
...
 
+  @Test
+  void afterCommit() throws CloseFailuresException {
+    try (MemoryDb db = MemoryDb.newInstance();
+        Cleaner cleaner = new Cleaner()) {
+      Fork view = db.createFork(cleaner);
+      setServiceNode(node);
+      service.initialize(view);
+
+      BlockCommittedEvent event = mock(BlockCommittedEvent.class);
```
dmitry-timofeev: Why do you mock a value and not just create it? You don't seem to verify any interactions with it

MakarovS: But why can't we use this mock as a stub? We don't use it inside `afterCommit` method, so we can avoid creating it. What are the disadvantages of using a mock here (or advantages of creating an object)? Mock seems simpler to me.

dmitry-timofeev: As a mocking framework is designed to verify interactions between objects, using a mock usually suggests that you will do that — verify the SuT interaction with this mock (or help it get through to the interesting condition). As one does not verify interactions with values, using mocks to create an object gives a misleading message.

exonum-java-binding-qa-service/src/main/java/com/exonum/binding/qaservice/QaServiceImpl.java
```diff
@@ -108,6 +118,18 @@ public void createPublicApiHandlers(Node node, Router router) {
     controller.mountApi(router);
   }
 
+  /**
+   * Increments the afterCommit counter so the number of times this method was invoked is stored
+   * in it.
+   */
+  @Override
+  public void afterCommit(BlockCommittedEvent event) {
+    long seed = 1L;
```
dmitry-timofeev: I don't believe it will work as it will submit duplicate transactions.

exonum-java-binding-qa-service/src/main/java/com/exonum/binding/qaservice/QaServiceImpl.java
```diff
@@ -215,4 +237,8 @@ private HashCode submitTransaction(Transaction tx) {
   private void checkBlockchainInitialized() {
     checkState(node != null, "Service has not been fully initialized yet");
   }
+
+  private CreateCounterTx createCounter(String name, Fork fork) {
```
dmitry-timofeev: I'd keep the private methods in some logical order too — usually, in the call order (after the callee).

exonum-java-binding-qa-service/src/test/java/com/exonum/binding/qaservice/QaServiceImplIntegrationTest.java
```diff
@@ -308,6 +312,30 @@ void getBlockTransactions() {
...
+      service.afterCommit(event);
+
+      // Check that the afterCommit counter was updated.
```
dmitry-timofeev: This test does not seem correct — after commit cannot synchronously make any changes to the database.

exonum-java-binding-qa-service/src/main/java/com/exonum/binding/qaservice/QaServiceImpl.java
```diff
@@ -108,6 +121,18 @@ public void createPublicApiHandlers(Node node, Router router) {
     controller.mountApi(router);
   }
 
+  /**
+   * Increments the afterCommit counter so the number of times this method was invoked is stored
+   * in it.
+   */
+  @Override
+  public void afterCommit(BlockCommittedEvent event) {
+    long seed = RANDOM.nextLong();
```
dmitry-timofeev: Shan't we just use the blockchain height from the event — it is guaranteed to be unique? It will be simpler in terms of code and the reader won't have to think about the probability of a certain value appearing again in the sequence after _N_ invocations. It must be small given `N << 2^64` though

exonum-java-binding-qa-service/src/test/java/com/exonum/binding/qaservice/QaServiceImplIntegrationTest.java
```diff
@@ -308,6 +315,22 @@ void getBlockTransactions() {
...
+         Cleaner cleaner = new Cleaner()) {
+      Fork fork = db.createFork(cleaner);
+      setServiceNode(node);
+      service.initialize(fork);
+
+      Snapshot snapshot = db.createSnapshot(cleaner);
+      BlockCommittedEvent event = BlockCommittedEventImpl.valueOf(snapshot, OptionalInt.of(1), 1L);
+      service.afterCommit(event);
+
+      verify(service).submitIncrementCounter(anyLong(), any(HashCode.class));
```
dmitry-timofeev: Such use of Mockito won't work — it can only verify invocations on spies and mocks. Shan;t you verify invocation of a `Node#submitTransaction` with a certain transaction?

exonum-java-binding-qa-service/src/main/java/com/exonum/binding/qaservice/QaServiceImpl.java
MakarovS: Yes, as you mentioned it's `propose_timeout_threshold` - https://github.com/exonum/exonum-java-binding/pull/571/files/99e7f9e644726b119e992b242e25e609d4b91278..31ec5c477ea24654538c922f91fef77d875a73ee#diff-53ec26b2abc32dea308e36d7e4efdc4aR79
Also there is `txs_block_limit` - maximum number of transactions per block

exonum-java-binding-qa-service/src/test/java/com/exonum/binding/qaservice/QaServiceImplIntegrationTest.java
```diff
@@ -313,26 +315,24 @@ void getBlockTransactions() {
   }
 
   @Test
```
dmitry-timofeev: `@RequiresNativeLibrary`

PR #570, opened 2018-11-09
Implemented native method for CoreSchema#getActualConfiguration [ECR-2646].

skletsun:
## Overview
The method returns the configuration for the latest height of the blockchain serialized to JSON string.

---
See: https://jira.bf.local/browse/ECR-2646

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding-core/rust/src/storage/core_schema.rs
+++ b/exonum-java-binding-core/rust/src/storage/core_schema.rs
@@ -18,7 +18,7 @@ use exonum::{
 };
 use jni::{
     objects::JClass,
-    sys::{jbyteArray, jlong},
+    sys::{jbyteArray, jlong, jstring},
     JNIEnv,
 };
 use std::{panic, ptr};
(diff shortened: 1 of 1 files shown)


exonum-java-binding-core/rust/src/storage/core_schema.rs
```diff
@@ -94,3 +94,25 @@ pub extern "system" fn Java_com_exonum_binding_blockchain_CoreSchemaProxy_native
     });
     utils::unwrap_exc_or(&env, res, ptr::null_mut())
 }
+
+/// Returns the configuration for the latest height of the blockchain. Throws
+/// `java.lang.RuntimeException` if the "genesis block" has not been created yet.
```
DENMROOT: Can we throw custom exception here?  RuntimeException is too general.

exonum-java-binding-core/rust/src/storage/core_schema.rs
```diff
@@ -94,3 +94,25 @@ pub extern "system" fn Java_com_exonum_binding_blockchain_CoreSchemaProxy_native
...
+    env: JNIEnv,
+    _: JClass,
+    schema_handle: Handle,
+) -> jstring {
+    let res = panic::catch_unwind(|| {
+        let val = match utils::cast_handle::<SchemaType>(schema_handle) {
+            SchemaType::SnapshotSchema(schema) => schema.actual_configuration(),
+            SchemaType::ForkSchema(schema) => schema.actual_configuration(),
+        };
+        serde_json::to_value(val)
```
vitvakatu: This is redundant, you should be able to use `serde_json::to_string` directly, without conversion to `serde_json::Value`

exonum-java-binding-core/rust/src/storage/core_schema.rs
skletsun: I agree that it's rather general, but for the moment it's our convention. We can consider it as an area for further improvement.

exonum-java-binding-core/rust/src/storage/core_schema.rs
skletsun: Thanks, corrected

PR #572, opened 2018-11-19
Implement block commited event translation from Rust to Java [ECR-2454]

skletsun:
### Implemented translation of the `after_commit` callback from Rust to Java.
---
See: https://jira.bf.local/browse/ECR-2454

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding-core/rust/integration_tests/src/mock/service.rs
+++ b/exonum-java-binding-core/rust/integration_tests/src/mock/service.rs
@@ -162,6 +162,19 @@ impl ServiceMockBuilder {
         self
     }
 
+    pub fn get_mock_interaction_after_commit(self) -> (Self, GlobalRef) {
+        let obj = unwrap_jni(self.exec.with_attached(|env| {
+            let mock_interaction: JObject = env.call_method(
+                self.builder.as_obj(),
+                "getMockInteractionAfterCommit",
+                "()Lcom/exonum/binding/fakes/mocks/MockInteraction;",
+                &[],
+            )?.l()?;
+            Ok(env.new_global_ref(mock_interaction)?)
+        }));
+        (self, obj)
+    }
+
     pub fn build(self) -> ServiceProxy {
         let (executor, service) = unwrap_jni(self.exec.clone().with_attached(|env| {
             let value = env.call_method(
--- a/exonum-java-binding-core/rust/integration_tests/src/mock/service.rs
+++ b/exonum-java-binding-core/rust/integration_tests/src/mock/service.rs
@@ -1,4 +1,4 @@
-use java_bindings::exonum::crypto::{Hash, hash};
+use java_bindings::exonum::crypto::{hash, Hash};
 use java_bindings::jni::objects::{GlobalRef, JObject, JValue};
 use java_bindings::jni::strings::JNIString;
 use java_bindings::jni::sys::jsize;
--- a/exonum-java-binding-core/rust/integration_tests/src/mock/service.rs
+++ b/exonum-java-binding-core/rust/integration_tests/src/mock/service.rs
@@ -164,12 +164,13 @@ impl ServiceMockBuilder {
 
     pub fn get_mock_interaction_after_commit(self) -> (Self, GlobalRef) {
         let obj = unwrap_jni(self.exec.with_attached(|env| {
-            let mock_interaction: JObject = env.call_method(
-                self.builder.as_obj(),
-                "getMockInteractionAfterCommit",
-                "()Lcom/exonum/binding/fakes/mocks/MockInteraction;",
-                &[],
-            )?.l()?;
+            let mock_interaction: JObject = env
+                .call_method(
+                    self.builder.as_obj(),
+                    "getMockInteractionAfterCommit",
+                    "()Lcom/exonum/binding/fakes/mocks/MockInteraction;",
+                    &[],
+                )?.l()?;
             Ok(env.new_global_ref(mock_interaction)?)
         }));
         (self, obj)
--- a/exonum-java-binding-core/rust/integration_tests/src/mock/service.rs
+++ b/exonum-java-binding-core/rust/integration_tests/src/mock/service.rs
@@ -148,6 +148,20 @@ impl ServiceMockBuilder {
         self
     }
 
+    pub fn after_commit_throwing(self, exception_class: &str) -> Self {
+        unwrap_jni(self.exec.with_attached(|env| {
+            let exception = env.find_class(exception_class)?;
+            env.call_method(
+                self.builder.as_obj(),
+                "afterCommitHandlerThrowing",
+                "(Ljava/lang/Class;)V",
+                &[JValue::from(JObject::from(exception.into_inner()))],
+            )?;
+            Ok(())
+        }));
+        self
+    }
+
     pub fn build(self) -> ServiceProxy {
         let (executor, service) = unwrap_jni(self.exec.clone().with_attached(|env| {
             let value = env.call_method(
--- a/exonum-java-binding-core/rust/integration_tests/src/mock/service.rs
+++ b/exonum-java-binding-core/rust/integration_tests/src/mock/service.rs
@@ -12,9 +12,6 @@ pub const SERVICE_ADAPTER_CLASS: &str = "com/exonum/binding/service/adapters/Use
 pub const SERVICE_MOCK_BUILDER_CLASS: &str =
     "com/exonum/binding/fakes/mocks/UserServiceAdapterMockBuilder";
 
-pub const SERVICE_DEFAULT_ID: u16 = 42;
-pub const SERVICE_DEFAULT_NAME: &str = "service 42";
-
 pub struct ServiceMockBuilder {
     exec: MainExecutor,
     builder: GlobalRef,
@@ -32,8 +29,6 @@ impl ServiceMockBuilder {
             env.new_global_ref(value.l()?)
         }));
         ServiceMockBuilder { exec, builder }
-            .id(SERVICE_DEFAULT_ID)
-            .name(SERVICE_DEFAULT_NAME)
     }
 
     pub fn id(self, id: u16) -> Self {
--- a/exonum-java-binding-core/rust/integration_tests/src/mock/service.rs
+++ b/exonum-java-binding-core/rust/integration_tests/src/mock/service.rs
@@ -1,4 +1,4 @@
-use java_bindings::exonum::crypto::Hash;
+use java_bindings::exonum::crypto::{Hash, hash};
 use java_bindings::jni::objects::{GlobalRef, JObject, JValue};
 use java_bindings::jni::strings::JNIString;
 use java_bindings::jni::sys::jsize;
@@ -11,6 +11,9 @@ use super::NATIVE_FACADE_CLASS;
 pub const SERVICE_ADAPTER_CLASS: &str = "com/exonum/binding/service/adapters/UserServiceAdapter";
 pub const SERVICE_MOCK_BUILDER_CLASS: &str =
     "com/exonum/binding/fakes/mocks/UserServiceAdapterMockBuilder";
+
+pub const SERVICE_DEFAULT_ID: u16 = 42;
+pub const SERVICE_DEFAULT_NAME: &str = "service 42";
 
 pub struct ServiceMockBuilder {
     exec: MainExecutor,
@@ -29,6 +32,9 @@ impl ServiceMockBuilder {
             env.new_global_ref(value.l()?)
         }));
         ServiceMockBuilder { exec, builder }
+            .id(SERVICE_DEFAULT_ID)
+            .name(SERVICE_DEFAULT_NAME)
+            .state_hashes(&[hash(&[1, 2, 3])])
     }
 
     pub fn id(self, id: u16) -> Self {
--- a/exonum-java-binding-core/rust/integration_tests/tests/service_proxy.rs
+++ b/exonum-java-binding-core/rust/integration_tests/tests/service_proxy.rs
@@ -1,8 +1,10 @@
+extern crate exonum_testkit;
 extern crate integration_tests;
 extern crate java_bindings;
 #[macro_use]
 extern crate lazy_static;
 
+use exonum_testkit::TestKitBuilder;
 use integration_tests::mock::service::ServiceMockBuilder;
 use integration_tests::mock::transaction::{create_mock_transaction, INFO_VALUE};
 use integration_tests::test_service::{
@@ -193,4 +195,34 @@ fn service_can_modify_db_on_initialize() {
         .get(&key)
         .expect("Failed to find the entry created in the test service");
     assert_eq!(INITIAL_ENTRY_VALUE, value);
+}
+
+#[test]
+#[should_panic(expected = "Java exception: java.lang.RuntimeException")]
+fn after_commit_throwing() {
+
+    let service = ServiceMockBuilder::new(EXECUTOR.clone())
+        .id(1)
+        .name("test")
+        .state_hashes(&[hash(&[1, 2, 3])])
+        .after_commit_throwing(EXCEPTION_CLASS) // TODO: think about changing to some custom exception that is not called from regular code
+        .build();
+
+    // It turned out that it is MUCH easier to use testkit in order to trigger the after_commit()
+    // callback than calling it by hands providing manually constructed ServiceContext entity.
+    let mut testkit = TestKitBuilder::validator()
+        .with_service(service.clone())
+        .create();
+
+    testkit.create_block();
+}
+
+#[test]
+fn after_commit_validator() {
+    unimplemented!()
+}
+
+#[test]
+fn after_commit_auditor() {
+    unimplemented!()
 }
--- a/exonum-java-binding-core/rust/integration_tests/tests/service_proxy.rs
+++ b/exonum-java-binding-core/rust/integration_tests/tests/service_proxy.rs
@@ -30,7 +30,7 @@ use java_bindings::{
         storage::{Database, MemoryDB},
     },
     jni::{objects::JObject, JavaVM},
-    serde_json::{from_str, Value},
+    serde_json,
     utils::{any_to_string, convert_to_string, unwrap_jni},
     JniExecutor, MainExecutor,
 };
@@ -48,7 +48,8 @@ const TEST_CONFIG_JSON: &str = r#""test config""#;
 const TEST_CONFIG_NOT_JSON: &str = r#"()"#;
 
 lazy_static! {
-    static ref TEST_CONFIG_VALUE: Value = Value::String("test config".to_string());
+    static ref TEST_CONFIG_VALUE: serde_json::Value =
+        serde_json::Value::String("test config".to_string());
 }
 
 #[test]
@@ -154,7 +155,7 @@ fn initialize_config_null() {
         .build();
 
     let config = service.initialize(&mut fork);
-    assert_eq!(config, Value::Null);
+    assert_eq!(config, serde_json::Value::Null);
 }
 
 #[test]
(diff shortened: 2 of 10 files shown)


exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/mocks/MockInteraction.java
```diff
@@ -0,0 +1,91 @@
...
+import org.mockito.stubbing.Answer;
+
+/**
+ * Helper class to make testing callbacks from native code easier. It allows simple tracking of
+ * interaction with a specific method of a mocked object and then return it (by calling
+ * getInteractions() method from native side) as a JSON string.
+ */
+public class MockInteraction {
+  private final String[] arguments;
+  private final List<String> interactions;
```
MakarovS: Initialize `interactions` at the same line where you declare it?
```suggestion
  private final List<String> interactions = new ArrayList();
```

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/mocks/MockInteraction.java
```diff
@@ -0,0 +1,91 @@
...
+  private MockInteraction(String[] arguments) {
+    this.arguments = arguments;
+    interactions = new ArrayList<>();
+  }
+
+  /**
+   * Creates new instange of {@link MockInteraction} configured with expected names of arguments for
+   * a mocked method.
+   *
+   * @param expectedArguments Names of expected arguments for a mocked method in their right order.
```
MakarovS: Nit: here and in other javadocs we don't use capital letters in the beginning and dots in the end in places like these.
```
* @param expectedArguments names of expected arguments for a mocked method in their right order
* @return new and configured instance of {@link MockInteraction}
```

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/mocks/MockInteraction.java
```diff
@@ -0,0 +1,91 @@
...
+   * Returns list of all interactions (arguments values) with particular method of mocked object
+   * in form of JSON string.
+   *
+   * <p>Example: "[{"handle":4635874800,"height":1},{"handle":4635875424,"height":2}]"</p>
+   *
+   * @return Result of interactions with mocked object.
+   */
+  public String getInteractions() {
+    // We don't use Gson here because we need custom serialization for strings and it is simpler
+    // to do it by hands
```
MakarovS: Replace `by hands` with `manually`?

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/mocks/MockInteraction.java
```diff
@@ -0,0 +1,91 @@
...
+   * @return New and configured instance of {@link MockInteraction}
+   */
+  public static MockInteraction createInteraction(String[] expectedArguments) {
+    return new MockInteraction(expectedArguments);
+  }
+
+  /**
+   * Custom Answer that is used to track interactions with a particular method of mocked object.
+   */
+  private class InteractionAnswer implements Answer {
```
MakarovS: Please add tests for this class.

exonum-java-binding-core/rust/integration_tests/tests/service_proxy.rs
```diff
@@ -1,32 +1,47 @@
...
+use java_bindings::{
+    exonum::{
+        blockchain::Service,
+        crypto::hash,
+        encoding::Error as MessageError,
+        messages::RawTransaction,
+        storage::{Database, MemoryDB},
+    },
+    jni::{objects::JObject, JavaVM},
+    serde_json::{from_str, Value},
```
vitvakatu: Perhaps a personal taste, but I don't like imports from `serde_json`. That's because when you find `let smth = from_str(smth_else)` - it's hard to understand why you can actually convert `smth_else` to `smth`

```suggestion
    serde_json,
```

exonum-java-binding-core/rust/integration_tests/tests/service_proxy.rs
```diff
@@ -194,3 +209,91 @@ fn service_can_modify_db_on_initialize() {
...
+    let service = builder.build();
+    let mut testkit = TestKitBuilder::validator()
+        .with_service(service.clone())
+        .create();
+
+    testkit.create_block();
+    testkit.create_block();
+
+    let result = get_mock_interaction_result(&EXECUTOR, interactor.as_obj());
+    let res_iter: Vec<AfterCommitArgs> = from_str(&result).unwrap();
```
vitvakatu: I'd rather call it `after_commit_args` 🙃

exonum-java-binding-core/rust/integration_tests/tests/service_proxy.rs
```diff
@@ -194,3 +209,91 @@ fn service_can_modify_db_on_initialize() {
...
+
+    assert_eq!(res_iter.len(), 2);
+
+    let item: &AfterCommitArgs = res_iter.get(0).unwrap();
```
vitvakatu: Not completely sure, but I remember clippy lint which forces `array[0]` instead of `array.get(0).unwrap()`. Or, actually: https://rust-lang-nursery.github.io/rust-clippy/master/index.html#get_unwrap

exonum-java-binding-core/rust/integration_tests/tests/service_proxy.rs
```diff
@@ -194,3 +209,91 @@ fn service_can_modify_db_on_initialize() {
...
+    assert!(item.handle > 0);
```
vitvakatu: I wonder if handle can be negative 🤔 If it's pointer representation, it can, I suppose

exonum-java-binding-core/rust/integration_tests/tests/service_proxy.rs
```diff
@@ -194,3 +209,91 @@ fn service_can_modify_db_on_initialize() {
...
+
+    assert_eq!(res_iter.len(), 1);
+
+    let item: &AfterCommitArgs = res_iter.get(0).unwrap();
+    assert!(item.handle > 0);
+    assert_eq!(item.validator, -1);
+    assert_eq!(item.height, 1);
+}
+
+// Helper methods that gets the JSON representation of interaction with mock
```
vitvakatu:
```suggestion
// Helper method. Gets the JSON representation of interaction with mock
```

exonum-java-binding-core/rust/integration_tests/tests/service_proxy.rs
```diff
@@ -1,32 +1,47 @@
...
-use std::panic::{catch_unwind, AssertUnwindSafe};
-use std::sync::Arc;
 
 lazy_static! {
     static ref VM: Arc<JavaVM> = create_vm_for_tests_with_fake_classes();
     pub static ref EXECUTOR: MainExecutor = MainExecutor::new(VM.clone());
 }
 
 const EXCEPTION_CLASS: &str = "java/lang/RuntimeException";
+const TEST_EXCEPTION_CLASS: &str = "com/exonum/binding/fakes/mocks/TestException";
```
vitvakatu: Unfortunately this constant is almost useless - it is used only once, and it duplicates line 214 (with `should_panic`) - but we can't reuse it there

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/mocks/MockInteraction.java
```diff
@@ -0,0 +1,91 @@
...
+  private final String[] arguments;
+  private final List<String> interactions;
+
+  private MockInteraction(String[] arguments) {
+    this.arguments = arguments;
+    interactions = new ArrayList<>();
+  }
+
+  /**
+   * Creates new instange of {@link MockInteraction} configured with expected names of arguments for
```
vitvakatu:
```suggestion
   * Creates new instance of {@link MockInteraction} configured with expected names of arguments for
```

exonum-java-binding-core/rust/integration_tests/tests/service_proxy.rs
skletsun: Most probably you're right... Then it makes sense to check for equality to zero only.

exonum-java-binding-core/rust/integration_tests/tests/service_proxy.rs
skletsun: Yes, but anyways to me it looks more consistent than specifying it once in that place. And yes, we definitely cannot reuse it in `should_panic`, unfortunately.

exonum-java-binding-core/rust/integration_tests/tests/service_proxy.rs
skletsun: 😁

exonum-java-binding-core/rust/integration_tests/tests/service_proxy.rs
skletsun: You're right, thanks.

exonum-java-binding-core/rust/integration_tests/tests/service_proxy.rs
skletsun: Well, mee too ) But I accepted this for tests 😁. Ok, I agree - rewriting it.

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/mocks/MockInteraction.java
skletsun: Doesn't it look like overhead to test our internal class created for testing our code? 🙃

MakarovS: It's up to you, but StringBuilder construction logic seems error-prone. However we'd get an error in tests where this class is used, so maybe it's indeed excessive. I'd wait for someone else's opinion.

vitvakatu: I don't think tests are necessary here

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/mocks/MockInteraction.java
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  /**
+   * Creates new instance of {@link MockInteraction} configured with expected names of arguments for
+   * a mocked method.
+   *
+   * @param expectedArguments names of expected arguments for a mocked method in their right order
+   * @return New and configured instance of {@link MockInteraction}
+   */
+  public static MockInteraction createInteraction(String[] expectedArguments) {
```
dmitry-timofeev: Is there a specific reason to use static factory method?

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/mocks/MockInteraction.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Override
+    public Object answer(InvocationOnMock invocationOnMock) throws Throwable {
+      Object[] args = invocationOnMock.getArguments();
+      int numArgs = args.length;
+
+      assert numArgs == arguments.length;
+
+      Gson gson = new Gson();
+      StringBuilder sb = new StringBuilder();
```
dmitry-timofeev: No need for this: you can convert a simple `Map<?, ?>` with Gson. 

However please do not change this at the moment because this class seems redundant: https://static.javadoc.io/org.mockito/mockito-core/2.23.4/org/mockito/Mockito.html#mocking_details

exonum-java-binding-core/rust/integration_tests/tests/service_proxy.rs
```diff
@@ -194,3 +210,90 @@ fn service_can_modify_db_on_initialize() {
         .expect("Failed to find the entry created in the test service");
     assert_eq!(INITIAL_ENTRY_VALUE, value);
 }
+
+#[test]
+#[should_panic(expected = "Java exception: com.exonum.binding.fakes.mocks.TestException")]
+fn after_commit_throwing() {
+    let service = ServiceMockBuilder::new(EXECUTOR.clone())
+        .state_hashes(&[hash(&[1, 2, 3])])
```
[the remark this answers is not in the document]
dmitry-timofeev: I wouldn't consider that a drawback :-) 

The usual approach with the test builders is to make the builder provide the defaults, so that the client code does not have to specify parameters that are not relevant to that test case. Can we make the ServiceMockBuilder always set some 32-byte hash?

FYI, more on that in this blog post: https://testing.googleblog.com/2018/02/testing-on-toilet-cleanly-create-test.html
and the book, which I highly recommend.

exonum-java-binding-core/rust/integration_tests/src/mock/service.rs
```diff
@@ -148,6 +148,34 @@ impl ServiceMockBuilder {
         self
     }
 
+    pub fn after_commit_throwing(self, exception_class: &str) -> Self {
```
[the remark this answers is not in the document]
skletsun: UserServiceAdapter#afterCommit() re-throws caught CloseFailuresException as RuntimeException if any.

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/mocks/MockInteraction.java
skletsun: No, just wanted to provide a user with straightforward method that creates required object. Perhaps, within the current implementation it's not important.

exonum-java-binding-core/rust/integration_tests/src/mock/service.rs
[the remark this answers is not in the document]
dmitry-timofeev: The doc for `Service` says 
> Any exceptions in this method will be swallowed and will not affect the processing of transactions or blocks.

Therefore the impl of the adapter is not correct — it must log and swallow _user_ exceptions (CloseFailures is not one of them and it *is* handled correctly).

https://jira.bf.local/browse/ECR-2692

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/mocks/UserServiceAdapterMockBuilder.java
```diff
@@ -36,7 +36,29 @@
...
+    init_mocks();
+  }
+
+  private void init_mocks() {
+    byte[] hash = new byte[32];
+    for (int i = 0; i < 32; i++) {
+      hash[i] = (byte) i;
+    }
+
+    when(service.getId()).thenReturn((short) 42);
```
dmitry-timofeev: Please reuse the methods that are already there (id, name, stateHashes) for clarity.

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/mocks/UserServiceAdapterMockBuilder.java
```diff
@@ -36,7 +36,29 @@
 @SuppressWarnings({"unused", "WeakerAccess"}) // Used in native code
 public final class UserServiceAdapterMockBuilder {
 
-  private final UserServiceAdapter service = mock(UserServiceAdapter.class);
+  private final UserServiceAdapter service;
+
+  /**
+   * Default constructor.
```
dmitry-timofeev: Please remove this comment — it has negative value. Could probably make the class and ctor package-private to avoid having to document it.

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/mocks/UserServiceAdapterMockBuilder.java
```diff
@@ -36,7 +36,29 @@
...
+   */
+  public UserServiceAdapterMockBuilder() {
+    this.service = mock(UserServiceAdapter.class);
+
+    // Some default mocking that could be overridden later
+    init_mocks();
+  }
+
+  private void init_mocks() {
```
dmitry-timofeev: underscore is not allowed in method names

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/mocks/UserServiceAdapterMockBuilder.java
```diff
@@ -36,7 +36,29 @@
...
+    byte[] hash = new byte[32];
```
dmitry-timofeev: `s|32|Hashing.<the constant>|`

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/mocks/UserServiceAdapterMockBuilder.java
```diff
@@ -36,7 +36,29 @@
...
+  private final UserServiceAdapter service;
+
+  /**
+   * Default constructor.
+   */
+  public UserServiceAdapterMockBuilder() {
+    this.service = mock(UserServiceAdapter.class);
+
+    // Some default mocking that could be overridden later
+    init_mocks();
```
dmitry-timofeev: It is good to set defaults here — but if we do that, then this code and the constants
```rust
        ServiceMockBuilder { exec, builder }
            .id(SERVICE_DEFAULT_ID) <- no longer needed
            .name(SERVICE_DEFAULT_NAME) <- no longer needed
```
must be removed.

exonum-java-binding-fakes/pom.xml
```diff
@@ -77,7 +77,7 @@
         <groupId>com.google.code.gson</groupId>
         <artifactId>gson</artifactId>
         <version>2.8.5</version>
-        <scope>compile</scope>
+        <scope>test</scope>
```
dmitry-timofeev: It is not correct — it is required in the compile scope (which is the default).

exonum-java-binding-core/rust/integration_tests/src/mock/service.rs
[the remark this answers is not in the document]
dmitry-timofeev: Therefore, I'd make this method hard-code RuntimeException.

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/mocks/UserServiceAdapterMockBuilder.java
skletsun: Ok, I added it just because of Checkstyle warnings. Going to make it package-private.

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/mocks/UserServiceAdapterMockBuilder.java
skletsun: Well, I just missed that place... Now I'm wondering what is better - just add a record with setting defaults for hashes and rollback the changes for Java part OR remove this and finish the Java part? @dmitry-timofeev WDYT - is the first option is acceptable?

exonum-java-binding-fakes/pom.xml
skletsun: well, could you please then guide me here what exactly should I do since I'm not so experienced with maven?

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/mocks/UserServiceAdapterMockBuilder.java
dmitry-timofeev: I would set defaults in a single place only: either in Java constructor or in Rust ctor. Both options are good.

> just add a record with setting defaults for hashes here and rollback the changes for Java part

If by 'here' you mean in the Rust ctor — it sounds perfectly good to me.

exonum-java-binding-fakes/pom.xml
dmitry-timofeev: Just remove the scope because it is the default = compile.

exonum-java-binding-fakes/src/main/java/com/exonum/binding/fakes/mocks/UserServiceAdapterMockBuilder.java
dmitry-timofeev: If you go with the first — I believe some comments on the Java part may be no longer relevant.

dmitry-timofeev:
Re keeping track of the invocations to verify them — improvements are to be done separately by Java guys, so that we can integrate this PR earlier: https://jira.bf.local/browse/ECR-2694

PR #573, opened 2018-11-19
CoreSchema accessors P2 [ECR-2542]

MakarovS:
## Overview
Add CoreSchema accessors P2.
Changes to QA service will be in separate PR.

---
See: https://jira.bf.local/browse/ECR-2542

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Block.java
+++ b/exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Block.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2018 The Exonum Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.exonum.binding.blockchain;
+
+import com.exonum.binding.common.hash.HashCode;
+import com.google.common.base.Objects;
+
+/**
+ * Exonum block header data structure.
+ *
+ * A block is essentially a list of transactions, which is a result of the consensus algorithm
+ * (thus authenticated by the supermajority of validators) and is applied atomically to the
+ * blockchain state.
+ *
+ * This structure only contains the amount of transactions and the transactions root hash as well as
+ * other information, but not the transactions themselves.
+ */
+class Block {
+
+  private short proposerId;
+  private long height;
+  private int numTransactions;
+  private HashCode previousBlockHash;
+  private HashCode txRootHash;
+  private HashCode stateHash;
+
+  Block(short proposerId, long height, int numTransactions, HashCode previousBlockHash,
+      HashCode txRootHash, HashCode stateHash) {
+    this.proposerId = proposerId;
+    this.height = height;
+    this.numTransactions = numTransactions;
+    this.previousBlockHash = previousBlockHash;
+    this.txRootHash = txRootHash;
+    this.stateHash = stateHash;
+  }
+
+  /**
+   * Identifier of the leader node which has proposed the block.
+   */
+  public short getProposerId() {
+    return proposerId;
+  }
+
+  /**
+   * Height of the block, which is also the number of this particular block in the blockchain.
+   */
+  public long getHeight() {
+    return height;
+  }
+
+  /**
+   * Number of transactions in this block.
+   */
+  public int getNumTransactions() {
+    return numTransactions;
+  }
+
+  /**
+   * Hash link to the previous block in the blockchain.
+   */
+  public HashCode getPreviousBlockHash() {
+    return previousBlockHash;
+  }
+
+  /**
+   * Root hash of the Merkle tree of transactions in this block.
+   */
+  public HashCode getTxRootHash() {
+    return txRootHash;
+  }
+
+  /**
+   * Hash of the blockchain state after applying transactions in the block.
+   */
+  public HashCode getStateHash() {
+    return stateHash;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    Block block = (Block) o;
+    return proposerId == block.proposerId &&
+        height == block.height &&
+        numTransactions == block.numTransactions &&
+        Objects.equal(previousBlockHash, block.previousBlockHash) &&
+        Objects.equal(txRootHash, block.txRootHash) &&
+        Objects.equal(stateHash, block.stateHash);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects
+        .hashCode(proposerId, height, numTransactions, previousBlockHash, txRootHash, stateHash);
+  }
+
+  @Override
+  public String toString() {
+    return "Block{" +
+        "proposerId=" + proposerId +
+        ", height=" + height +
+        ", numTransactions=" + numTransactions +
+        ", previousBlockHash=" + previousBlockHash +
+        ", txRootHash=" + txRootHash +
+        ", stateHash=" + stateHash +
+        '}';
+  }
+
+}
--- a/exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
+++ b/exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
@@ -18,10 +18,14 @@
 package com.exonum.binding.blockchain;
 
 import com.exonum.binding.common.hash.HashCode;
+import com.exonum.binding.common.message.TransactionMessage;
 import com.exonum.binding.storage.database.View;
 import com.exonum.binding.storage.indices.ListIndex;
+import com.exonum.binding.storage.indices.MapIndex;
 import com.exonum.binding.storage.indices.ProofListIndexProxy;
+import com.exonum.binding.storage.indices.ProofMapIndexProxy;
 import com.google.common.annotations.VisibleForTesting;
+import sun.reflect.generics.reflectiveObjects.NotImplementedException;
 
 /**
  * Provides read-only access to the subset of
@@ -79,6 +83,107 @@ public final class Blockchain {
    */
   public ProofListIndexProxy<HashCode> getBlockTransactions(long height) {
     return schema.getBlockTransactions(height);
+  }
+
+  /**
+   * Returns a proof list of transaction hashes committed in the block with given id.
+   *
+   * @param blockId id of the block
+   * @return a proof list of transaction hashes committed in the block.
+   *        Or an empty list if the block with given id doesn't exist
+   */
+   public ProofListIndexProxy<HashCode> getBlockTransactions(HashCode blockId) {
+     MapIndex<HashCode, Block> blocks = schema.getBlocks();
+     Block block = blocks.get(blockId);
+     return getBlockTransactions(block.getHeight());
+   }
+
+  /**
+   * Returns a proof list of transaction hashes committed in the given block.
+   *
+   * @param block block of which list of transaction hashes should be returned
+   * @return a proof list of transaction hashes committed in the block.
+   *        Or an empty list if the block with given id doesn't exist
+   */
+  public ProofListIndexProxy<HashCode> getBlockTransactions(Block block) {
+    return getBlockTransactions(block.getHeight());
+  }
+
+  /**
+   * Returns a table that represents a map with a key-value pair of a
+   * transaction hash and transaction message.
+   * @return a map with a key-value pair of a transaction hash and transaction message
+   */
+   public MapIndex<HashCode, TransactionMessage> getTxMessages() {
+     return schema.getTxMessages();
+   }
+
+  /**
+   * Returns a table that represents a map with a key-value pair of a transaction
+   * hash and execution result.
+   * @return a map with a key-value pair of a transaction hash and execution result
+   */
+  public ProofMapIndexProxy<HashCode, TransactionResult> getTxResults() {
+    return schema.getTxResults();
+  }
+
+  /**
+   * Returns a transaction execution result for given message hash.
+   * @param messageHash a message hash
+   * @return a transaction execution result
+   */
+  public TransactionResult getTxResult(HashCode messageHash) {
+    ProofMapIndexProxy<HashCode, TransactionResult> txResults = getTxResults();
+    return txResults.get(messageHash);
+  }
+
+  /**
+   * Returns a table that keeps the transaction position inside the block for every transaction
+   * hash.
+   * @return a map with transaction position for every transaction hash
+   */
+  public MapIndex<HashCode, TransactionLocation> getTxLocations() {
+    return schema.getTxLocations();
+  }
+
+  /**
+   * Return transaction position inside the block for given message hash.
+   * @param messageHash message hash
+   * @return transaction position inside the block
+   */
+  public TransactionLocation getTxLocation(HashCode messageHash) {
+    MapIndex<HashCode, TransactionLocation> txLocations = getTxLocations();
+    return txLocations.get(messageHash);
+  }
+
+  /**
+   * Returns a table that stores a block object for every block height.
+   * @return a map with block object for every block hash
+   */
+  public MapIndex<HashCode, Block> getBlocks() {
+    return schema.getBlocks();
+  }
+
+  // TODO: javadoc and implementation
+  public ListIndex<Block> getBlocksByHeight() {
+    throw new NotImplementedException();
+  }
+
+  /**
+   * Returns a block object for given block hash.
+   * @return a block object
+   */
+  public Block getBlock(HashCode blockHash) {
+    MapIndex<HashCode, Block> blocks = getBlocks();
+    return blocks.get(blockHash);
+  }
+
+  /**
+   * Returns the latest committed block.
+   * @return a block object
+   */
+  public Block getLastBlock() {
+    return schema.getLastBlock();
   }
 
 }
--- a/exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/CoreSchemaProxy.java
+++ b/exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/CoreSchemaProxy.java
@@ -21,6 +21,9 @@ import static com.exonum.binding.common.serialization.StandardSerializers.fixed6
 import static com.google.common.base.Preconditions.checkArgument;
 
 import com.exonum.binding.common.hash.HashCode;
+import com.exonum.binding.common.message.TransactionMessage;
+import com.exonum.binding.common.proofs.map.UncheckedMapProof;
+import com.exonum.binding.common.serialization.Serializer;
 import com.exonum.binding.common.serialization.StandardSerializers;
 import com.exonum.binding.proxy.Cleaner;
 import com.exonum.binding.proxy.NativeHandle;
@@ -28,7 +31,11 @@ import com.exonum.binding.proxy.ProxyDestructor;
 import com.exonum.binding.storage.database.View;
 import com.exonum.binding.storage.indices.ListIndex;
 import com.exonum.binding.storage.indices.ListIndexProxy;
+import com.exonum.binding.storage.indices.MapIndex;
+import com.exonum.binding.storage.indices.MapIndexProxy;
 import com.exonum.binding.storage.indices.ProofListIndexProxy;
+import com.exonum.binding.storage.indices.ProofMapIndexProxy;
+import sun.reflect.generics.reflectiveObjects.NotImplementedException;
 
 /**
  * A proxy class for the blockchain::Schema struct maintained by Exonum core.
@@ -39,6 +46,9 @@ final class CoreSchemaProxy {
 
   private final NativeHandle nativeHandle;
   private final View dbView;
+  private final Serializer<Block> blockSerializer = BlockSerializer.INSTANCE;
+  private final Serializer<TransactionLocation> transactionLocationSerializer =
+      TransactionLocationSerializer.INSTANCE;
 
   private CoreSchemaProxy(NativeHandle nativeHandle, View dbView) {
     this.nativeHandle = nativeHandle;
@@ -87,6 +97,60 @@ final class CoreSchemaProxy {
         CoreIndex.BLOCK_TRANSACTIONS, id, dbView, StandardSerializers.hash());
   }
 
+  /**
+   * Returns a table that stores a block object for every block height.
+   * @return a map with block object for every block hash
+   */
+  MapIndex<HashCode, Block> getBlocks() {
+    return MapIndexProxy.newInstance(
+        CoreIndex.BLOCKS, dbView, StandardSerializers.hash(), blockSerializer);
+  }
+
+  /**
+   * Returns the latest committed block.
+   * @return the latest committed block object
+   */
+  Block getLastBlock() {
+    return blockSerializer.fromBytes(nativeGetLastBlock(nativeHandle.get()));
+  }
+
+  /**
+   * Returns a table that represents a map with a key-value pair of a
+   * transaction hash and transaction message.
+   * @return a map with a key-value pair of a transaction hash and transaction message
+   */
+  MapIndex<HashCode, TransactionMessage> getTxMessages() {
+    // TODO: serializer
+    return MapIndexProxy.newInstance(CoreIndex.TRANSACTIONS, dbView, StandardSerializers.hash(),
+        null);
+  }
+
+  /**
+   * Returns a table that represents a map with a key-value pair of a transaction
+   * hash and execution result.
+   * @return a map with a key-value pair of a transaction hash and execution result
+   */
+  ProofMapIndexProxy<HashCode, TransactionResult> getTxResults() {
+    // TODO: serializer
+    return ProofMapIndexProxy.newInstance(CoreIndex.TRANSACTIONS_RESULTS, dbView,
+        StandardSerializers.hash(), null);
+  }
+
+  /**
+   * Returns a table that keeps the block height and transaction position inside the block for every
+   * transaction hash.
+   * @return a map with transaction position for every transaction hash
+   */
+  MapIndex<HashCode, TransactionLocation> getTxLocations() {
+    return MapIndexProxy.newInstance(CoreIndex.TRANSACTIONS_LOCATIONS, dbView,
+        StandardSerializers.hash(), transactionLocationSerializer);
+  }
+
+  @SuppressWarnings("unused") // TODO: should be implemented later
+  UncheckedMapProof getProofToServiceCollection(short serviceId, int collectionIndex) {
+    throw new NotImplementedException();
+  }
+
   private static native long nativeCreate(long viewNativeHandle);
 
   private static native void nativeFree(long nativeHandle);
@@ -98,7 +162,6 @@ final class CoreSchemaProxy {
    *
    * @throws RuntimeException if the "genesis block" was not created
    */
-  @SuppressWarnings("unused") //TODO: Will be done in the next task
   private static native byte[] nativeGetLastBlock(long nativeHandle);
 
   /**
@@ -108,6 +171,10 @@ final class CoreSchemaProxy {
     private static final String PREFIX = "core.";
     private static final String BLOCK_TRANSACTIONS = PREFIX + "block_transactions";
     private static final String ALL_BLOCK_HASHES = PREFIX + "block_hashes_by_height";
+    private static final String TRANSACTIONS = PREFIX + "transactions";
+    private static final String BLOCKS = PREFIX + "blocks";
+    private static final String TRANSACTIONS_RESULTS = PREFIX + "transaction_results";
+    private static final String TRANSACTIONS_LOCATIONS = PREFIX + "transactions_locations";
   }
 
 }
--- a/exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/TransactionResult.java
+++ b/exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/TransactionResult.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2018 The Exonum Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.exonum.binding.blockchain;
+
+import java.util.Optional;
+
+/**
+ * Returns a result of transaction execution. This result may be either a success, or an error,
+ * if execution has failed. Errors consist of an error code and an optional description.
+ */
+class TransactionResult {
+
+  private Type type;
+  private TransactionError error;
+
+  @SuppressWarnings("unused")  // native API
+  TransactionResult(Type type, TransactionError error) {
+    this.type = type;
+    this.error = error;
+  }
+
+  boolean isSuccessful() {
+    return type == Type.SUCCESS;
+  }
+
+  Type getType() {
+    return type;
+  }
+
+  Optional<TransactionError> getError() {
+    return Optional.ofNullable(error);
+  }
+
+  enum Type {
+    SUCCESS,
+    ERROR,
+    UNEXPECTED_ERROR
+  }
+
+}
--- a/exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/BlockchainTest.java
+++ b/exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/BlockchainTest.java
@@ -18,12 +18,14 @@
 package com.exonum.binding.blockchain;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.ArgumentMatchers.anyLong;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+import com.exonum.binding.common.hash.HashCode;
 import com.exonum.binding.storage.indices.ListIndexProxy;
+import com.exonum.binding.storage.indices.MapIndex;
 import com.exonum.binding.storage.indices.ProofListIndexProxy;
+import com.exonum.binding.storage.indices.ProofMapIndexProxy;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
@@ -33,7 +35,10 @@ import org.mockito.junit.jupiter.MockitoExtension;
 @ExtendWith(MockitoExtension.class)
 class BlockchainTest {
 
+  private static final long HEIGHT = 10L;
+
   private Blockchain blockchain;
+
   @Mock
   private CoreSchemaProxy mockSchema;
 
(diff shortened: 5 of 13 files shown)


exonum-java-binding-common/src/main/proto/block.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+
+option java_package = "com.exonum.binding.common.serialization";
```
DENMROOT: Not sure about package location. `serialization` looks strange for model classes What do you think about `com.exonum.binding.common.model `or `com.exonum.binding.common.schema` ?

exonum-java-binding-common/src/main/proto/transactionLocation.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+
+option java_package = "com.exonum.binding.common.serialization";
```
DENMROOT: Same about package name here

exonum-java-binding-common/src/test/java/com/exonum/binding/common/serialization/TransactionLocationSerializerTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+import com.exonum.binding.common.blockchain.TransactionLocation;
+import java.util.stream.Stream;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+class TransactionLocationSerializerTest {
+  private Serializer<TransactionLocation> serializer = TransactionLocationSerializer.INSTANCE;
+
+  @ParameterizedTest
```
DENMROOT: Whats the purpose of Parametrized test with 1 test case ? :)

exonum-java-binding-common/src/test/java/com/exonum/binding/common/serialization/BlockSerializerTest.java
```diff
@@ -0,0 +1,47 @@
...
+import com.exonum.binding.common.blockchain.Block;
+import com.exonum.binding.common.hash.HashCode;
+import java.util.stream.Stream;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+class BlockSerializerTest {
+  private Serializer<Block> serializer = BlockSerializer.INSTANCE;
+
+  @ParameterizedTest
```
DENMROOT: 1 test case ?

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -81,4 +88,105 @@ public long getHeight() {
...
+
+  /**
+   * Returns a table that stores a block object for every block height.
+   * @return a map with block object for every block hash
+   */
+  public MapIndex<HashCode, Block> getBlocks() {
+    return schema.getBlocks();
+  }
+
+  // TODO: javadoc and implementation
```
DENMROOT: Can you add Jira ticket link for this TODO

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -81,4 +88,105 @@ public long getHeight() {
...
+   * @param messageHash message hash
+   * @return transaction position inside the block
+   */
+  public TransactionLocation getTxLocation(HashCode messageHash) {
+    MapIndex<HashCode, TransactionLocation> txLocations = getTxLocations();
+    return txLocations.get(messageHash);
+  }
+
+  /**
+   * Returns a table that stores a block object for every block height.
```
DENMROOT: Is this Javadoc correct ?  block for height but I see block for hashCode

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/CoreSchemaProxy.java
```diff
@@ -87,6 +105,59 @@ long getHeight() {
...
+    return blockSerializer.fromBytes(nativeGetLastBlock(nativeHandle.get()));
+  }
+
+  /**
+   * Returns a table that represents a map with a key-value pair of a
+   * transaction hash and transaction message.
+   * @return a map with a key-value pair of a transaction hash and transaction message
+   */
+  MapIndex<HashCode, TransactionMessage> getTxMessages() {
+    // TODO: serializer
```
DENMROOT: TODO ?

DENMROOT:
Any plans for QA service updates according to this PR ?

exonum-java-binding-common/src/main/java/com/exonum/binding/common/serialization/BlockSerializer.java
```diff
@@ -0,0 +1,65 @@
...
+    try {
+      BlockProtos.Block copiedBlockProtos = BlockProtos.Block.parseFrom(binaryBlock);
+      block = new Block((short) copiedBlockProtos.getProposerId(),
+          copiedBlockProtos.getHeight(), copiedBlockProtos.getNumTransactions(),
+          toByteString(copiedBlockProtos.getPreviousBlockHash()),
+          toByteString(copiedBlockProtos.getTxRootHash()),
+          toByteString(copiedBlockProtos.getStateHash()));
+    } catch (InvalidProtocolBufferException e) {
+      throw new IllegalArgumentException(
+          "Unable to instantiate BlockProtos.Block instance from provided binary data", e);
```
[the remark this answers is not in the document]
MakarovS: We have this phrasing with `instantiate` in all our serializers, it seems okay to me.

exonum-java-binding-common/src/main/proto/block.proto
MakarovS: Seems ok to me. We have the same package location for `TestProtos` and for other protos we don't make a separate package as well

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -81,4 +89,114 @@ public long getHeight() {
...
+  public ProofListIndexProxy<HashCode> getBlockTransactions(Block block) {
+    return getBlockTransactions(block.getHeight());
+  }
+
+  /**
+   * Returns a table that represents a map with a key-value pair of a
+   * transaction hash and transaction message.
+   * @return a map with a key-value pair of a transaction hash and transaction message
+   */
+   public MapIndex<HashCode, TransactionMessage> getTxMessages() {
```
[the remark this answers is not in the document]
vitvakatu: Only committed transactions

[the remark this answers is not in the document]
vitvakatu: committed + in-pool

exonum-java-binding-common/src/main/proto/block.proto
```diff
@@ -0,0 +1,13 @@
...
+option java_outer_classname = "BlockProtos";
+
+message Block {
+  int32 proposerId = 1;
+  int64 height = 2;
```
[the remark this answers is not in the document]
dvush: We use uint64. [blockchain.proto](https://github.com/exonum/exonum/blob/master/exonum/src/encoding/protobuf/proto/blockchain.proto)

I thought that if protobuf specification allows this type than we should use it. But as I understood from [protocol-buffers/docs](https://developers.google.com/protocol-buffers/docs/proto#scalar) it may cause some inconvenience for Java developers when dealing with large unsigned numbers.

Also I wonder why do you write your own incompatible .proto files?
Maybe we can plan on how to sync them in future. It is also important for service/light client developers since they need .proto files as a part of exonum interface.

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -81,4 +89,114 @@ public long getHeight() {
...
+   * @param blockId id of the block
+   * @return a proof list of transaction hashes committed in the block.
+   *        Or null if the block with given id doesn't exist
+   */
+  @Nullable
+  public ProofListIndexProxy<HashCode> getBlockTransactions(HashCode blockId) {
+    MapIndex<HashCode, Block> blocks = schema.getBlocks();
+    Block block = blocks.get(blockId);
+    if (block == null) {
+      return null;
```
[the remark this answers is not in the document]
MakarovS: Yes, I agree that null seems bad in this case. I think we should use checked exceptions here then? Empty list seems not okay, as we can probably have a block without transactions?

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -81,4 +89,114 @@ public long getHeight() {
...
+  }
+
+  /**
+   * Returns a transaction execution result for given message hash.
+   * @param messageHash a message hash
+   * @return a transaction execution result
+   */
+  public TransactionResult getTxResult(HashCode messageHash) {
+    ProofMapIndexProxy<HashCode, TransactionResult> txResults = getTxResults();
+    return txResults.get(messageHash);
```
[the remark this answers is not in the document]
MakarovS: I'd go with `Optional` then

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/CoreSchemaProxy.java
```diff
@@ -87,6 +105,60 @@ long getHeight() {
         CoreIndex.BLOCK_TRANSACTIONS, id, dbView, StandardSerializers.hash());
   }
 
+  /**
+   * Returns a table that stores a block object for every block hash.
```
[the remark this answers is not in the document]
MakarovS: I wanted to stay consistent as existing methods in this class had javadocs. I think it's okay to leave them, even though they often duplicate javadocs from `Blockchain` class.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/blockchain/TransactionResult.java
```diff
@@ -0,0 +1,94 @@
...
+   */
+  public boolean isSuccessful() {
+    return type == Type.SUCCESS;
+  }
+
+  /**
+   * Return type of the transaction.
+   * @return {@code Type.SUCCESS} if transaction was successful
+   *         {@code Type.ERROR} if there was an error during transaction execution
+   *         {@code Type.UNEXPECTED_ERROR} if there was an error during transaction execution
```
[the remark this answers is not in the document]
MakarovS: I feel that if I change it to `{@code Type.UNEXPECTED_ERROR} if there was an unexpected error during transaction execution` it won't be sufficient, should I add something else?

exonum-java-binding-common/src/main/java/com/exonum/binding/common/blockchain/Block.java
```diff
@@ -0,0 +1,76 @@
...
+ * (thus authenticated by the supermajority of validators) and is applied atomically to the
+ * blockchain state.
+ *
+ * This structure only contains the amount of transactions and the transactions root hash as well as
+ * other information, but not the transactions themselves.
+ */
+@AutoValue
+public abstract class Block {
+
+  public static Block valueOf(
```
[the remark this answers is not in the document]
MakarovS: We don't have that in core.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/blockchain/TransactionError.java
```diff
@@ -0,0 +1,46 @@
...
+import java.io.Serializable;
+import java.util.Optional;
+
+/**
+ * Result of unsuccessful transaction execution.
+ */
+@AutoValue
+public abstract class TransactionError implements Serializable {
+
+  public static TransactionError valueOf(byte errorCode, String description) {
```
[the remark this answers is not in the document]
MakarovS: I believe `@Nullable` should be good enough.

exonum-java-binding-common/src/main/proto/block.proto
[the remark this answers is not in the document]
MakarovS: Yes, I wanted to reuse proto files from core, but I wasn't sure about how to sync. Current block.proto and transactionLocation.proto are temporary, we'll definitely have to use the same proto files

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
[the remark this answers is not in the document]
dmitry-timofeev: It might make sense to delay these changes (optionals + exceptions) and ponder about various options for a while. We may discuss it during the grooming.

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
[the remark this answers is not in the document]
dmitry-timofeev: Thanks, that is interesting, as some collections contain committed-only (results, locations), but this — committed + in pool. It (kinda) makes sense to me, but, I suspect, not so much to a user.

exonum-java-binding-common/src/main/proto/block.proto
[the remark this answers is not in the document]
dmitry-timofeev: Thanks, that's a good question if we can pull these files automatically/have an integration or system test verifying they are up-to-date. @dip56 :question:

> inconvenience … with large unsigned numbers.

The question is how likely are any of the corresponding objects that use uint64 can exceed 2^63-1 (9.223372036854775807 × 10^18). If they can't — it is not a problem.

@MakarovS , if those are temporary, please mark them as such and add an issue to this epic. Or, if they are ready in the core, we may pull them from here.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/blockchain/TransactionResult.java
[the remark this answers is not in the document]
dmitry-timofeev: we may contrast a "user-definied error" (or, better, "service-defined") with an "unexpected error".

exonum-java-binding-common/src/main/java/com/exonum/binding/common/blockchain/Block.java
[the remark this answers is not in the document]
dmitry-timofeev: Yes, but not exactly. The core doesn't have a hash as a stored *property*, but it gives you an instance method `hash` to compute it, because `Block` used to be (can't say about the protobuf implementation) a wrapper around a binary buffer. As `Block` is identified by its hash, I'd include that in the Java `Block`.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/blockchain/TransactionError.java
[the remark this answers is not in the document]
dmitry-timofeev: Yes, it's good here

exonum-java-binding-common/src/main/proto/block.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+
+option java_package = "com.exonum.binding.common.serialization";
+option java_outer_classname = "BlockProtos";
```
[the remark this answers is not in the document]
MakarovS: If we remove `java_outer_classname`, it still will be generated by converting the file name to camel case.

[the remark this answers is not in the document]
dmitry-timofeev: Not necessarily if `option java_multiple_files = true;`:  https://developers.google.com/protocol-buffers/docs/proto3#options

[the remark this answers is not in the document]
MakarovS: Thanks!
But if we will use the same proto messages as in core (unless they rename those), then we'd have same names for proto messages (`Block`) and our classes (`Block`), so I'd propose to leave outer class. Or we could move all proto messages into one class and have something like `Protos.Block`?

exonum-java-binding-common/src/main/java/com/exonum/binding/common/serialization/TransactionLocationSerializer.java
```diff
@@ -0,0 +1,47 @@
...
+        TransactionLocationProtos.TransactionLocation.newBuilder()
+            .setHeight(value.getHeight())
+            .setIndexInBlock(value.getIndexInBlock())
+            .build();
+    return txLocation.toByteArray();
+  }
+
+  @Override
+  public TransactionLocation fromBytes(byte[] binaryTransactionLocation) {
+    try {
```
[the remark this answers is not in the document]
MakarovS: I would do it for all our protobuf serializers in a separate PR

exonum-java-binding-common/src/main/proto/block.proto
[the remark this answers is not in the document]
MakarovS: I refactored those so we have similar protos now

exonum-java-binding-common/src/main/java/com/exonum/binding/common/blockchain/TransactionError.java
```diff
@@ -0,0 +1,46 @@
...
+
+package com.exonum.binding.common.blockchain;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.base.Objects;
+import java.io.Serializable;
+import java.util.Optional;
+
+/**
+ * Result of unsuccessful transaction execution.
```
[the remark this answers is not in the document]
MakarovS: Added `@see TransactionExecutionException` in `TransactionResult` javadoc

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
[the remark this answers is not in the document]
MakarovS: > txResult(notYetExecutedButSubmittedTxHash) -> Optional.empty()

Is there anything wrong with that? We don't know whether this transaction was submitted, but not executed yet, or not submitted at all. Optional.empty() seems appropriate for both cases

exonum-java-binding-common/src/main/proto/block.proto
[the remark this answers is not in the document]
dmitry-timofeev: If I understand correctly, there are a couple of approaches:
1. Name the proto messages as such, e.g., `BlockProto`
2. Add an outer class for each
3. Add a single outer class for all (i.e., define the messages in a single `.proto` file).

Both (2) and (3) give you an opportunity to `import` a simple name. With (1) you are stuck with `Proto` suffix, but, in cases when you need to use both a message and an application class, that suffix might be more concise than a qualified name (e.g., `BlockProto` vs `BlockProto.Block` vs `CoreProtos.Block`).

Actually, any seems fine to me. If we need to use both proto messages and our classes in the same context, I'd strive to make the code concise.

If there aren't any apparent drawbacks, I'd suggest making (3) change in core. However, it will not be desirable if someone needs only a subset of these proto messages (e.g,. a `Block`, but not a dozen of other proto messages).

exonum-java-binding-common/src/main/java/com/exonum/binding/common/serialization/TransactionLocationSerializer.java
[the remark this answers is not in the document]
dmitry-timofeev: :ok_hand:, a task please.

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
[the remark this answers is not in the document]
dmitry-timofeev: > Is there anything wrong with that? We don't know whether this transaction was submitted, but not executed yet, or not submitted at all.

This collection indeed does not contain such information; other collections — do. The thing is, a user is _likely_ to know that a transaction was submitted if they have the hash of its message.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/serialization/TransactionLocationSerializer.java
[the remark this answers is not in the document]
MakarovS: https://jira.bf.local/browse/ECR-2712

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Block.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  /**
+   * Identifier of the leader node which has proposed the block.
+   */
+  public abstract int getProposerId();
+
+  /**
+   * Height of the block, which also identifies the number of this particular block in the
+   * blockchain.
```
bullet-tooth: ...blockchain starting from 0 ("genesis" block)

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Block.java
```diff
@@ -0,0 +1,88 @@
...
+   */
+  public abstract HashCode getPreviousBlockHash();
+
+  /**
+   * Root hash of the Merkle tree of transactions in this block.
+   */
+  public abstract HashCode getTxRootHash();
+
+  /**
+   * Hash of the blockchain state after applying transactions in the block.
```
bullet-tooth: Does it equal to the blockchain root hash at the moment of applying this block?

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Block.java
```diff
@@ -0,0 +1,88 @@
...
+   */
+  public abstract HashCode getStateHash();
+
+  /**
+   * Calculates a hash of this instance.
```
bullet-tooth: I would reword this javadoc to "Returns the SHA-256 hash of this block".

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Block.java
```diff
@@ -0,0 +1,88 @@
...
+   */
+  public HashCode getBlockHash() {
+    return HashCode.fromBytes(blockSerializer.toBytes(this));
```
bullet-tooth: Look like it's a bug here. We need to use `HashFunction` instead of just copying block bytes.

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -71,17 +75,102 @@ public long getHeight() {
...
-   * @return a proof list of transaction hashes committed in the block.
-   *        Or an empty list if the block at the given height doesn't exist
    * @throws IllegalArgumentException if the height is negative
    */
   public ProofListIndexProxy<HashCode> getBlockTransactions(long height) {
     return schema.getBlockTransactions(height);
   }
 
+  /**
+   * Returns a proof list of transaction hashes committed in the block with given id or null if
```
bullet-tooth: It's quite weird. There are 3 overloaded methods and in the same case one returns `null` and others return 'empty list'

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Block.java
MakarovS: Yes

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
MakarovS: Thanks, I changed it to `empty list` for now, but we'd need to decide whether we want to use nulls, empty lists or exceptions (either checked or unchecked).

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Block.java
```diff
@@ -0,0 +1,90 @@
...
+ * (thus authenticated by the supermajority of validators) and is applied atomically to the
+ * blockchain state.
+ *
+ * <p>This structure only contains the amount of transactions and the transactions root hash as well
+ * as other information, but not the transactions themselves.
+ */
+@AutoValue
+public abstract class Block {
+
+  private final Serializer<Block> blockSerializer = BlockSerializer.INSTANCE;
```
bullet-tooth: static?

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Block.java
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Hash of the blockchain state after applying transactions in the block.
+   */
+  public abstract HashCode getStateHash();
+
+  /**
+   * Returns the SHA-256 hash of this block.
+   */
+  public HashCode getBlockHash() {
+    return Hashing.defaultHashFunction()
```
bullet-tooth: probably we need to use `Hashing.sha256()` or to change the javadoc.
Theoretically, may happen such a situation when `defaultHashFunction` changes and javadoc becomes wrong.

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -71,17 +74,101 @@ public long getHeight() {
...
+   */
+  public ProofListIndexProxy<HashCode> getBlockTransactions(HashCode blockId) {
+    MapIndex<HashCode, Block> blocks = schema.getBlocks();
+    Block block = blocks.get(blockId);
+    return getBlockTransactions(block.getHeight());
+  }
+
+  /**
+   * Returns a proof list of transaction hashes committed in the given block or an empty list if
+   * the block with given id doesn't exist.
```
bullet-tooth:
```suggestion
   * the block doesn't exist.
```

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/TransactionResult.java
```diff
@@ -0,0 +1,76 @@
...
+
+package com.exonum.binding.blockchain;
+
+import com.exonum.binding.transaction.TransactionExecutionException;
+import com.google.auto.value.AutoValue;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+/**
+ * Returns a result of transaction execution. This result may be either a success, or an error,
```
bullet-tooth:
```suggestion
 * Returns a result of the transaction execution. This result may be either a success or an error
```

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/TransactionResult.java
```diff
@@ -0,0 +1,76 @@
...
+ * if execution has failed.
+ * Errors might be either service-defined or unexpected. Service-defined errors consist of an error
+ * code and an optional description. Unexpected errors consist of an optional description.
```
bullet-tooth:
```suggestion
 * code and an optional description. Unexpected errors consist of an optional description only.
```

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/CoreSchemaProxyIntegrationTest.java
```diff
@@ -57,8 +58,32 @@ void getBlockTransactionsTest() {
...
   @Test
   void getActiveConfigurationBeforeGenesisBlock() {
-    assertSchema((schema) ->
-        assertThrows(RuntimeException.class, schema::getActualConfiguration));
+    assertSchema((schema) -> assertThrows(RuntimeException.class, schema::getActualConfiguration));
+  }
+
+  @Test
+  void getBlocksTest() {
+    assertSchema((schema) -> assertTrue(schema.getBlocks().isEmpty()));
```
bullet-tooth:
```suggestion
    assertSchema((schema) -> assertThat(schema.getBlocks()).isEmpty());
```

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/CoreSchemaProxyIntegrationTest.java
```diff
@@ -57,8 +58,32 @@ void getBlockTransactionsTest() {
...
+  }
+
+  @Test
+  void getLastBlockBeforeGenesisBlockTest() {
+    assertSchema((schema) -> assertThrows(RuntimeException.class, schema::getLastBlock));
+  }
+
+  @Test
+  void getTxMessagesTest() {
+    assertSchema((schema) -> assertTrue(schema.getTxMessages().isEmpty()));
```
bullet-tooth:
```suggestion
    assertSchema((schema) -> assertThat(schema.getTxMessages()).isEmpty());
```

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/CoreSchemaProxyIntegrationTest.java
```diff
@@ -57,8 +58,32 @@ void getBlockTransactionsTest() {
...
+  }
+
+  @Test
+  void getTxResultsTest() {
+    assertSchema((schema) -> assertTrue(schema.getTxResults().isEmpty()));
```
bullet-tooth:
```suggestion
    assertSchema((schema) -> assertThat(schema.getTxResults()).isEmpty());
```

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/CoreSchemaProxyIntegrationTest.java
```diff
@@ -57,8 +58,32 @@ void getBlockTransactionsTest() {
...
+  }
+
+  @Test
+  void getTxLocationsTest() {
+    assertSchema((schema) -> assertTrue(schema.getTxLocations().isEmpty()));
```
bullet-tooth:
```suggestion
    assertSchema((schema) -> assertThat(schema.getTxLocations()).isEmpty());
```

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/serialization/BlockSerializerTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+  private Serializer<Block> serializer = BlockSerializer.INSTANCE;
+
+  @ParameterizedTest
+  @MethodSource("testSource")
+  void roundTrip(Block expected) {
+    byte[] bytes = serializer.toBytes(expected);
+    Block actual = serializer.fromBytes(bytes);
+
+    assertThat(actual, equalTo(expected));
```
bullet-tooth: can we use `com.exonum.binding.common.serialization.StandardSerializersTest#roundTripTest` here?

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/serialization/TransactionLocationSerializerTest.java
```diff
@@ -0,0 +1,47 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+
+class TransactionLocationSerializerTest {
+
+  private Serializer<TransactionLocation> serializer = TransactionLocationSerializer.INSTANCE;
+
+  @ParameterizedTest
+  @MethodSource("testSource")
+  void roundTrip(TransactionLocation expected) {
+    byte[] bytes = serializer.toBytes(expected);
```
bullet-tooth: how about to use `com.exonum.binding.common.serialization.StandardSerializersTest#roundTripTest`?

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/serialization/TransactionResultSerializerTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+
+class TransactionResultSerializerTest {
+
+  private Serializer<TransactionResult> serializer = TransactionResultSerializer.INSTANCE;
+
+  @ParameterizedTest
+  @MethodSource("testSource")
+  void roundTrip(TransactionResult expected) {
+    byte[] bytes = serializer.toBytes(expected);
```
bullet-tooth: the same here

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/CoreSchemaProxyIntegrationTest.java
dmitry-timofeev: It's not a `j.u.List`

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/serialization/BlockSerializerTest.java
MakarovS: I don't think it's worth the hassle of using `common` test classes here.
@dmitry-timofeev what do you think?

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/CoreSchemaProxy.java
```diff
@@ -89,6 +104,47 @@ long getHeight() {
         CoreIndex.BLOCK_TRANSACTIONS, id, dbView, StandardSerializers.hash());
   }
 
+  /**
```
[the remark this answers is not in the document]
dmitry-timofeev: That's good, thanks, missed that.

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Block.java
```diff
@@ -0,0 +1,84 @@
...
+ * (thus authenticated by the supermajority of validators) and is applied atomically to the
+ * blockchain state.
+ *
+ * <p>This structure only contains the amount of transactions and the transactions root hash as well
+ * as other information, but not the transactions themselves.
+ */
+@AutoValue
+public abstract class Block {
+
+  public static Block valueOf(
```
dmitry-timofeev: 7 parameters of only two types (3 integers and 4 hashcodes) beg for a builder (may be done separately, but add a TODO with a jira ref)

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/TransactionResult.java
```diff
@@ -0,0 +1,76 @@
...
+   *         execution
+   */
+  public abstract Type getType();
+
+  /**
+   * Returns an error code of a transaction if its execution resulted in a service-defined error.
+   * @return a transaction error code in case of a service-defined error, or
+   * {@code Optional.empty()} otherwise
+   */
+  public abstract Optional<Integer> getErrorCode();
```
dmitry-timofeev: `OptionalInt` or `Optional<Short>`?

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/TransactionResult.java
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * Returns an optional description of a transaction if its execution resulted in an error.
+   * @return a description of an error, or {@code Optional.empty()} if transaction was successful
```
dmitry-timofeev: if there is no description

(it is perfectly possible in case of a user-defined error)

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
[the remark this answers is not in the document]
dmitry-timofeev: Please document the methods that return `null`/throw exceptions in Jira + TODOs.

Such methods will have to be fixed and have proper documentation on their behaviour.

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Block.java
MakarovS: Will do!

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/TransactionResult.java
MakarovS: Although in core we can guarantee that possible value will be less than Short.MAX_VALUE, in proto file we have `uint32`, so I wouldn't use `Optional<Short>` here.
I've added `errorCode == null ? OptionalInt.empty() : OptionalInt.of(errorCode)` but it seems kind of ugly, as `OptionalInt` doesn't have `ofNullable` method.

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/BlockchainTest.java
```diff
@@ -128,7 +129,18 @@ void getTxResult() {
...
+
+  @Test
+  void getNonexistentTxResult() {
+    ProofMapIndexProxy mockMapIndex = mock(ProofMapIndexProxy.class);
+    HashCode messageHash = HashCode.fromString("ab");
+
+    when(mockMapIndex.get(messageHash)).thenReturn(null);
+    when(mockSchema.getTxResults()).thenReturn(mockMapIndex);
+
+    assertThat(blockchain.getTxResult(messageHash)).isEqualTo(Optional.empty());
```
dmitry-timofeev: You can use (here and elsewhere) https://joel-costigliola.github.io/assertj/core-8/api/org/assertj/core/api/AbstractOptionalAssert.html#isEmpty--

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/CoreSchemaProxy.java
```diff
@@ -42,6 +50,13 @@
 
   private final NativeHandle nativeHandle;
   private final View dbView;
+  private static final Serializer<Block> blockSerializer = BlockSerializer.INSTANCE;
```
bullet-tooth: it's a constant. Why not it the upper case?

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/serialization/BlockSerializerTest.java
```diff
@@ -0,0 +1,79 @@
...
+import com.exonum.binding.common.hash.HashCode;
+import com.exonum.binding.common.hash.Hashing;
+import com.exonum.binding.common.serialization.Serializer;
+import java.util.stream.Stream;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+class BlockSerializerTest {
+
+  private static final Serializer<Block> serializer = BlockSerializer.INSTANCE;
```
bullet-tooth: upper case?

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/serialization/BlockSerializerTest.java
bullet-tooth: Ohh, I see. I missed that this test in the core module.

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/CoreSchemaProxy.java
dmitry-timofeev: Yep, it falls under our definition: https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

dmitry-timofeev:
Fixed the Checkstyle (I added a little more than Javadocs, but specialized factory methods), hopefully, the build too: b4a9d92657136a0adbb1afe39b0c8362869fa03e 

@bullet-tooth , @MakarovS , please review

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/serialization/TransactionResultSerializerTest.java
```diff
@@ -42,9 +41,10 @@ void roundTrip(TransactionResult expected) {
 
   private static Stream<TransactionResult> testSource() {
     return Stream.of(
-        TransactionResult.valueOf(Type.SUCCESS, null, null),
-        TransactionResult.valueOf(Type.ERROR, 1, "Error description"),
-        TransactionResult.valueOf(Type.UNEXPECTED_ERROR, null, ""));
+        TransactionResult.successful(),
+        TransactionResult.error(1, "Error description"),
+        TransactionResult.error(255, /* No description: */ null),
```
MakarovS: `null` description turns into `Optional.empty()` inside the class and then in serializer#toBytes it turns into an empty string, so this test fails :(

dmitry-timofeev: Thanks, fixed

dmitry-timofeev:
@MakarovS , that's a good question, I've put together `nullOrEmptyToNone` method to accept both an empty string and null ref as no description.

I've also submitted an issue to discuss what _return_ type makes sense here: we might use an empty string as "no description". Same with error codes: 
- https://jira.bf.local/browse/ECR-2748
- (0 ec) https://jira.bf.local/browse/ECR-2749

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/serialization/TransactionResultSerializer.java
```diff
@@ -16,13 +16,18 @@
...
+
 import com.exonum.binding.blockchain.TransactionResult;
 import com.exonum.binding.common.serialization.Serializer;
 import com.google.protobuf.InvalidProtocolBufferException;
 
 public enum TransactionResultSerializer implements Serializer<TransactionResult> {
   INSTANCE;
 
+  private static final int SUCCESSFUL_RESULT_STATUS_CODE = 256;
+  private static final int UNEXPECTED_ERROR_CODE = 257;
```
MakarovS: The same - `private static final int UNEXPECTED_ERROR_CODE = SUCCESSFUL_RESULT_STATUS_CODE + 1;`?

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/serialization/TransactionResultSerializer.java
```diff
@@ -16,13 +16,18 @@
...
+import static com.exonum.binding.blockchain.TransactionResult.MAX_USER_DEFINED_ERROR_CODE;
+
 import com.exonum.binding.blockchain.TransactionResult;
 import com.exonum.binding.common.serialization.Serializer;
 import com.google.protobuf.InvalidProtocolBufferException;
 
 public enum TransactionResultSerializer implements Serializer<TransactionResult> {
   INSTANCE;
 
+  private static final int SUCCESSFUL_RESULT_STATUS_CODE = 256;
```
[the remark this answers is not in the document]
dmitry-timofeev: I'd probably keep as is, because arithmetic operations on error codes don't make much sense.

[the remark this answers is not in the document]
MakarovS: I don't like arithmetic operations on error codes either (and some other code style features in our core module :) ). But the code as it is right now seems a bit messy, as we defined one constant in one class and two others in another. Maybe it would make sense to define all of them in `TransactionResult` class? That way we could avoid arithmetic operations, but they would be stored in the same place and relationship between them would be more clear.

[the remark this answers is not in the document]
dmitry-timofeev: Putting in one place is definitely better, `TxResult` is a good place)

MakarovS:
👍

PR #577, opened 2018-11-26
Improve NativeHandle debuggability [ECR-2528]

DENMROOT:
## Overview
NativeHandle preserve the original value of the handle after invalidation.

---
See: https://jira.bf.local/browse/ECR-2528

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding-core/src/main/java/com/exonum/binding/proxy/NativeHandle.java
+++ b/exonum-java-binding-core/src/main/java/com/exonum/binding/proxy/NativeHandle.java
@@ -40,7 +40,7 @@ public final class NativeHandle implements AutoCloseable {
    * @throws IllegalStateException if this native handle is invalid (nullptr)
    */
   public NativeHandle(long nativeHandle) {
-    checkState(nativeHandle != INVALID_NATIVE_HANDLE, "This handle is not valid: %s", this);
+    checkState(nativeHandle != INVALID_NATIVE_HANDLE, "This handle is not valid: %s", nativeHandle);
 
     this.nativeHandle = nativeHandle;
     this.isValid = true;
@@ -63,7 +63,7 @@ public final class NativeHandle implements AutoCloseable {
 
   @Override
   public void close() {
-    if (isValid) {
+    if (isValid()) {
       invalidate();
     }
   }
--- a/exonum-java-binding-core/src/main/java/com/exonum/binding/proxy/NativeHandle.java
+++ b/exonum-java-binding-core/src/main/java/com/exonum/binding/proxy/NativeHandle.java
@@ -21,7 +21,7 @@ import static com.google.common.base.Preconditions.checkState;
 import com.google.common.base.MoreObjects;
 
 /**
- * An implementation-specific handle to the native object.
+ * An implementation-specific handle to the native object. Once closed, can no longer be accessed.
  */
 public final class NativeHandle implements AutoCloseable {
 
@@ -40,10 +40,10 @@ public final class NativeHandle implements AutoCloseable {
    * @throws IllegalStateException if this native handle is invalid (nullptr)
    */
   public NativeHandle(long nativeHandle) {
+    checkState(nativeHandle != INVALID_NATIVE_HANDLE, "This handle is not valid: %s", this);
+
     this.nativeHandle = nativeHandle;
     this.isValid = true;
-
-    checkValid();
   }
 
   /**
@@ -63,20 +63,17 @@ public final class NativeHandle implements AutoCloseable {
 
   @Override
   public void close() {
-    if (isValid()) {
+    if (isValid) {
       invalidate();
     }
   }
 
   private void checkValid() {
-    checkState(isValid(), "This handle is not valid: %s", this);
+    checkState(isValid, "This handle is not valid: %s", this);
   }
 
-  /**
-   * Returns true if this native handle is valid.
-   */
   final boolean isValid() {
-    return nativeHandle != INVALID_NATIVE_HANDLE && isValid;
+    return isValid;
   }
 
   private void invalidate() {
--- a/exonum-java-binding-core/src/main/java/com/exonum/binding/proxy/NativeHandle.java
+++ b/exonum-java-binding-core/src/main/java/com/exonum/binding/proxy/NativeHandle.java
@@ -30,9 +30,12 @@ public final class NativeHandle implements AutoCloseable {
    */
   public static final long INVALID_NATIVE_HANDLE = 0L;
 
-  private long nativeHandle;
+  private final long nativeHandle;
+
+  private boolean isValid;
 
   public NativeHandle(long nativeHandle) {
+    this.isValid = true;
     this.nativeHandle = nativeHandle;
   }
 
@@ -66,11 +69,11 @@ public final class NativeHandle implements AutoCloseable {
    * Returns true if this native handle is valid.
    */
   final boolean isValid() {
-    return nativeHandle != INVALID_NATIVE_HANDLE;
+    return nativeHandle != INVALID_NATIVE_HANDLE && isValid;
   }
 
   private void invalidate() {
-    nativeHandle = INVALID_NATIVE_HANDLE;
+    isValid = false;
   }
 
   @Override
--- a/exonum-java-binding-core/src/main/java/com/exonum/binding/proxy/NativeHandle.java
+++ b/exonum-java-binding-core/src/main/java/com/exonum/binding/proxy/NativeHandle.java
@@ -21,7 +21,7 @@ import static com.google.common.base.Preconditions.checkState;
 import com.google.common.base.MoreObjects;
 
 /**
- * An implementation-specific handle to the native object. Once closed, can no longer be accessed.
+ * An implementation-specific handle to the native object.
  */
 public final class NativeHandle implements AutoCloseable {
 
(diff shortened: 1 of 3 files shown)


MakarovS:
@DENMROOT You probably wanted to appoint reviewers, not assignees?

vitvakatu:
Funny, I'm 100% sure Github did not support multiple assignees before.

exonum-java-binding-core/src/main/java/com/exonum/binding/proxy/NativeHandle.java
```diff
@@ -66,11 +69,11 @@ private void checkValid() {
    * Returns true if this native handle is valid.
    */
   final boolean isValid() {
-    return nativeHandle != INVALID_NATIVE_HANDLE;
+    return nativeHandle != INVALID_NATIVE_HANDLE && isValid;
```
dmitry-timofeev: I wonder if we shall _allow_ accessing `nullptr` == 0, just to simplify the condition here? @vitvakatu , @DarkEld3r , we can't have `nullptr` as a handle anywhere, can we?

exonum-java-binding-core/src/test/java/com/exonum/binding/proxy/NativeHandleTest.java
```diff
@@ -53,6 +51,7 @@ void close() {
 
     nativeHandle.close();
     assertFalse(nativeHandle.isValid());
+    assertThat(nativeHandle.toString()).contains(convertHandle(handle));
```
dmitry-timofeev: Nit: I'd suggest using a string so that the expected format is apparent and this code is more readable: "11" or "0x11" — whatever it is.

dmitry-timofeev:
> Funny, I'm 100% sure Github did not support multiple assignees before.

I saw it work a couple of months ago, I believe.

exonum-java-binding-core/src/main/java/com/exonum/binding/proxy/NativeHandle.java
vitvakatu: I believe we can safely use `return isValid` here, yep. We will be able to test debugging capabilities in case of any... hmmm.... troubles :D

DENMROOT: And what about creating incorrect handles from constructor ? 
new NativeHandel(INVALID_NATIVE_HANDLE) - do we allow this ?

exonum-java-binding-core/src/test/java/com/exonum/binding/proxy/NativeHandleTest.java
DENMROOT: I'll create String constant for this.

exonum-java-binding-core/src/test/java/com/exonum/binding/proxy/NativeHandleTest.java
```diff
@@ -23,53 +23,53 @@
...
 import org.junit.jupiter.api.Test;
 
-
 class NativeHandleTest {
 
   private NativeHandle nativeHandle;
 
+  private static long HANDLE = 0x11L;
+
+  private static final String NATIVE_HANDLE = "11";
```
dmitry-timofeev: I can't say it is better, it is not the same constant that you reuse in code because it *must* have the same value. It just _happens_ that tests use the same value, therefore, extracting these constants does not make them clearer, on contrary — it makes them harder to understand, because you can't immediately see that the second is another representation of the first.

On top of that, the naming does not help either.

exonum-java-binding-core/src/main/java/com/exonum/binding/proxy/NativeHandle.java
vitvakatu: I'm not completely sure, but we can allow it

DENMROOT: @dmitry-timofeev Allow creation of a handle from INVALID_NATIVE_HANDLE ? Should it be VALID until we invalidate it manually ?

dmitry-timofeev: I'd go with a stricter rule first and relax later if needed: forbid instantiation, if it *ever* (which we currently doubt, if I understand correctly) need that, allow.

exonum-java-binding-core/src/test/java/com/exonum/binding/proxy/NativeHandleTest.java
```diff
@@ -23,53 +23,53 @@
...
 
-
 class NativeHandleTest {
 
   private NativeHandle nativeHandle;
 
+  private static long HANDLE = 0x11L;
+
+  private static final String HANDLE_STRING_REPRESENTATION =
+      Long.toHexString(HANDLE).toUpperCase();
```
dmitry-timofeev: "11";

exonum-java-binding-core/src/main/java/com/exonum/binding/proxy/NativeHandle.java
dmitry-timofeev: The idea was to simplify this method — i.e., remove `nativeHandle != INVALID_NATIVE_HANDLE` since this condition is not possible because it is enforced on the instantiation.

exonum-java-binding-core/src/main/java/com/exonum/binding/proxy/NativeHandle.java
```diff
@@ -21,7 +21,7 @@
 import com.google.common.base.MoreObjects;
 
 /**
- * An implementation-specific handle to the native object. Once closed, can no longer be accessed.
+ * An implementation-specific handle to the native object.
```
dmitry-timofeev: Why this change if this is still true?

exonum-java-binding-core/src/test/java/com/exonum/binding/proxy/AbstractCloseableNativeProxyTest.java
```diff
@@ -126,9 +113,8 @@ void getNativeHandle_ShallFailIfProxyIsClosed() {
   @Test
   void getNativeHandle_ShallFailIfInvalid() {
```
dmitry-timofeev: The test name is no longer correct.

exonum-java-binding-core/src/test/java/com/exonum/binding/proxy/NativeHandleTest.java
DENMROOT: > I can't say it is better, it is not the same constant that you reuse in code because it must have the same value. It just happens that tests use the same value, therefore, extracting these constants does not make them clearer, on contrary — it makes them harder to understand, because you can't immediately see that the second is another representation of the first.
- No ?

exonum-java-binding-core/src/main/java/com/exonum/binding/proxy/NativeHandle.java
```diff
@@ -40,10 +40,10 @@
    * @throws IllegalStateException if this native handle is invalid (nullptr)
    */
   public NativeHandle(long nativeHandle) {
+    checkState(nativeHandle != INVALID_NATIVE_HANDLE, "This handle is not valid: %s", this);
```
dmitry-timofeev: Do not invoke toString on a partially constructed object

exonum-java-binding-core/src/test/java/com/exonum/binding/proxy/AbstractCloseableNativeProxyTest.java
```diff
@@ -111,7 +110,7 @@ void getNativeHandle_ShallFailIfProxyIsClosed() {
   }
 
   @Test
-  void getNativeHandle_ShallFailIfInvalid() {
+  void createNativeHandle_ShallFailIfInvalid() {
```
dmitry-timofeev: NativeProxy, it is not a test of a handle

exonum-java-binding-core/src/test/java/com/exonum/binding/proxy/NativeHandleTest.java
dmitry-timofeev: I am not sure I understand what you mean. No to what and why?

exonum-java-binding-core/src/main/java/com/exonum/binding/proxy/NativeHandle.java
```diff
@@ -53,24 +63,21 @@ public long get() {
 
   @Override
   public void close() {
-    if (isValid()) {
+    if (isValid) {
```
dmitry-timofeev: I'd consider keeping the method calls as more refactoring-friendly (if we change what isValid means)

exonum-java-binding-core/src/main/java/com/exonum/binding/proxy/NativeHandle.java
DENMROOT: Good catch , thanks

exonum-java-binding-core/src/test/java/com/exonum/binding/proxy/NativeHandleTest.java
DENMROOT: Trying to make HANDLE_STRING_REPRESENTATION more clear.

PR #579, opened 2018-11-28
Implemented initial cache for references of Java classes and method ids [ECR-2645]

skletsun:
## Implemented cache for repeatedly called classes and methods.
For the moment it affects the most frequently called methods of UserTransactionAdapter, UserServiceAdapter and some classes used while checking for errors.

---
See: https://jira.bf.local/browse/ECR-2645

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Removed
 - `com.exonum.binding.common.proofs.map.MapEntry` — moved to package
   `com.exonum.binding.common.collect`.
+  
+### Fixed
+- A bug in the cryptocurrency demo frontend that sometimes resulted in rejected transactions and/or
+wrong response code (#621).
 
 ## [0.3] - 2018-10-30
 
--- a/exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/CoreSchemaProxy.java
+++ b/exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/CoreSchemaProxy.java
@@ -100,15 +100,15 @@ final class CoreSchemaProxy {
    *
    * @throws IllegalArgumentException if the height is negative or there is no block at given height
    */
-  ProofListIndexProxy<HashCode> getBlockTransactions(long height) {
-    checkArgument(height >= 0, "Height shouldn't be negative, but was %s", height);
-    long actualHeight = getHeight();
+  ProofListIndexProxy<HashCode> getBlockTransactions(long blockHeight) {
+    checkArgument(blockHeight >= 0, "Height shouldn't be negative, but was %s", blockHeight);
+    long blockchainHeight = getHeight();
     checkArgument(
-        height > actualHeight,
+        blockchainHeight >= blockHeight,
         "Height should be less or equal compared to blockchain height %s, but was %s",
-        actualHeight,
-        height);
-    byte[] id = toCoreStorageKey(height);
+        blockchainHeight,
+        blockHeight);
+    byte[] id = toCoreStorageKey(blockHeight);
     return ProofListIndexProxy.newInGroupUnsafe(
         CoreIndex.BLOCK_TRANSACTIONS, id, dbView, StandardSerializers.hash());
   }
--- a/exonum-java-binding-cryptocurrency-demo/src/main/java/com/exonum/binding/cryptocurrency/ApiController.java
+++ b/exonum-java-binding-cryptocurrency-demo/src/main/java/com/exonum/binding/cryptocurrency/ApiController.java
@@ -82,6 +82,9 @@ final class ApiController {
   }
 
   private void submitTransaction(RoutingContext rc) {
+    checkRequestContentType(rc, "application/octet-stream");
+
+    // Read the transaction message
     Buffer buffer = rc.getBody();
     BinaryMessage message = BinaryMessage.fromBytes(buffer.getBytes());
 
@@ -94,6 +97,14 @@ final class ApiController {
     rc.response()
         .putHeader("Content-Type", "text/plain")
         .end(String.valueOf(txHash));
+  }
+
+  private void checkRequestContentType(RoutingContext rc,
+      String expectedContentType) {
+    String contentType = rc.request().getHeader("Content-Type");
+
+    checkArgument(expectedContentType.equals(contentType), "Content-Type must be %s, but was %s",
+        expectedContentType, contentType);
   }
 
   private void getWallet(RoutingContext rc) {
--- a/exonum-java-binding-cryptocurrency-demo/src/test/java/com/exonum/binding/cryptocurrency/ApiControllerTest.java
+++ b/exonum-java-binding-cryptocurrency-demo/src/test/java/com/exonum/binding/cryptocurrency/ApiControllerTest.java
@@ -122,7 +122,7 @@ class ApiControllerTest {
 
     String expectedResponse = messageHash;
     // Send a request to submitTransaction
-    post(ApiController.SUBMIT_TRANSACTION_PATH)
+    postTransaction()
         .sendBuffer(
             Buffer.buffer(message.getSignedMessage().array()),
             context.succeeding(
@@ -147,11 +147,30 @@ class ApiControllerTest {
   }
 
   @Test
+  void submitValidTransactionWrongContentType(VertxTestContext context) {
+    BinaryMessage message = createTestBinaryMessage(CREATE_WALLET_TX_ID);
+
+    // Send a request to submitTransaction
+    post(ApiController.SUBMIT_TRANSACTION_PATH)
+        .putHeader("Content-Type", "application/x-www-form-urlencoded")
+        .sendBuffer(
+            Buffer.buffer(message.getSignedMessage().array()),
+            context.succeeding(
+                response -> context.verify(() -> {
+                  // Check the response status
+                  int statusCode = response.statusCode();
+                  assertEquals(HTTP_BAD_REQUEST, statusCode);
+
+                  context.completeNow();
+                })));
+  }
+
+  @Test
   void submitTransactionOfIncorrectMessageSize(VertxTestContext context) {
     BinaryMessage message = createTestBinaryMessage(CREATE_WALLET_TX_ID);
     byte errorByte = 1;
 
-    post(ApiController.SUBMIT_TRANSACTION_PATH)
+    postTransaction()
         .sendBuffer(
             Buffer.buffer(message.getSignedMessage().array()).appendByte(errorByte),
             context.succeeding(response -> context.verify(() -> {
@@ -176,7 +195,7 @@ class ApiControllerTest {
     when(service.submitTransaction(transaction))
         .thenThrow(error);
 
-    post(ApiController.SUBMIT_TRANSACTION_PATH)
+    postTransaction()
         .sendBuffer(
             Buffer.buffer(message.getSignedMessage().array()),
             context.succeeding(response -> context.verify(() -> {
@@ -317,6 +336,11 @@ class ApiControllerTest {
 
   private HttpRequest<Buffer> get(String requestPath) {
     return webClient.get(port, HOST, requestPath);
+  }
+
+  private HttpRequest<Buffer> postTransaction() {
+    return post(ApiController.SUBMIT_TRANSACTION_PATH)
+        .putHeader("Content-Type", "application/octet-stream");
   }
 
   private HttpRequest<Buffer> post(String requestPath) {


exonum-java-binding-core/rust/integration_tests/tests/utils_errors.rs
```diff
@@ -3,12 +3,12 @@ extern crate java_bindings;
 #[macro_use]
 extern crate lazy_static;
 
-use integration_tests::vm::create_vm_for_tests;
+use integration_tests::vm::create_vm_for_tests_with_fake_classes;
```
skletsun: I switched to this because with just `create_vm_for_tests` it was unable to find `TransactionExecutioException` during initial caching.

exonum-java-binding-core/rust/integration_tests/tests/utils_errors.rs
```diff
@@ -22,8 +22,16 @@ const ARITHMETIC_EXCEPTION_CLASS_FQN: &str = "java.lang.ArithmeticException";
...
 
 lazy_static! {
-    static ref VM: Arc<JavaVM> = create_vm_for_tests();
-    pub static ref EXECUTOR: MainExecutor = MainExecutor::new(VM.clone());
+    static ref VM: Arc<JavaVM> = create_vm_for_tests_with_fake_classes();
+    pub static ref EXECUTOR: MainExecutor = {
+        let ex = MainExecutor::new(VM.clone());
+        // JNI_OnLoad() is not called for these tests
+        ex.with_attached(|env|{
+            jni_cache::cache_methods(env);
```
skletsun: It seems `JNI_OnLoad()` is not called for `utils_errors` family of tests, so I still see no better solution here.

exonum-java-binding-core/rust/src/utils/jni_cache.rs
```diff
@@ -0,0 +1,158 @@
...
+static mut CLASS_CACHED_ERROR: Option<GlobalRef> = None;
+static mut CLASS_CACHED_TEE: Option<GlobalRef> = None;
+
+#[allow(non_snake_case)]
+#[no_mangle]
+pub extern "system" fn JNI_OnLoad(vm: JavaVM, _: *mut c_void) -> jint {
+    let env = vm.get_env().expect("Cannot get reference to the JNIEnv");
+    cache_methods(&env);
+
+    JNI_VERSION_1_8
```
skletsun: I decided to return the 1.8 here. According to JNI docs: `If the native library does not export a JNI_OnLoad function, the VM assumes that the library only requires JNI version JNI_VERSION_1_1`. We didn't have this function exported before. Does it worth investigation of differences between JNI versions and/or noticing this in Changelog?

exonum-java-binding-core/rust/src/utils/jni_cache.rs
```diff
@@ -0,0 +1,158 @@
...
+static mut UTA_CACHED_VERIFY: Option<JMethodID> = None;
+
+static mut USA_CACHED_STATE_HASHES: Option<JMethodID> = None;
+static mut USA_CACHED_CONVERT_TRANSACTION: Option<JMethodID> = None;
+
+static mut CLASS_CACHED_ERROR: Option<GlobalRef> = None;
+static mut CLASS_CACHED_TEE: Option<GlobalRef> = None;
+
+#[allow(non_snake_case)]
+#[no_mangle]
```
skletsun: 1. Probably this function could be moved to separate file since it's not semantically linked to JNI cache.
2. Not sure whether we need to implement JNI_On_Unload() to do any cleanup of cached data - everything should be destroyed with JVN shutdown.

exonum-java-binding-core/rust/src/utils/mod.rs
```diff
@@ -20,6 +20,7 @@ mod errors;
 mod exception;
 mod handle;
 mod jni;
+pub mod jni_cache;
```
skletsun: I declared the module this way because the intention was to use cached entities in other modules as:
```
use jni_cache;
...
jni_cache::get_some_cached_entity()
```
@vitvakatu WDYT - is it a good approach?

exonum-java-binding-core/rust/integration_tests/tests/utils_errors.rs
vitvakatu: So are we using `create_vm_for_tests` now?

exonum-java-binding-core/rust/src/utils/errors.rs
```diff
@@ -3,11 +3,9 @@ use jni::JNIEnv;
 
 use std::cell::Cell;
 
-use utils::{get_class_name, get_exception_message};
+use utils::{get_class_name, get_exception_message, jni_cache};
```
vitvakatu: What about importing `get_error_class` here?

exonum-java-binding-core/rust/src/utils/jni_cache.rs
vitvakatu: As we use this function only for cache - I can't see any reason to move it from here. We can always do it later.

vitvakatu: JNI_On_Unload() is unnecessary according to requirements

exonum-java-binding-core/rust/src/utils/jni_cache.rs
```diff
@@ -0,0 +1,158 @@
...
+
+static mut OBJECT_CACHED_GET_CLASS: Option<JMethodID> = None;
+static mut CLASS_CACHED_GET_NAME: Option<JMethodID> = None;
+static mut THROWABLE_CACHED_GET_MESSAGE: Option<JMethodID> = None;
+
+static mut UTA_CACHED_EXECUTE: Option<JMethodID> = None;
+static mut UTA_CACHED_INFO: Option<JMethodID> = None;
+static mut UTA_CACHED_VERIFY: Option<JMethodID> = None;
+
+static mut USA_CACHED_STATE_HASHES: Option<JMethodID> = None;
```
vitvakatu: Cached USA states - sounds quite funny. Although names of these statics and method names are not much readable, unfortunately. We are not using such abbreviations commonly, but I don't know how we can avoid them without blowing method names lengths.

exonum-java-binding-core/rust/src/utils/jni_cache.rs
```diff
@@ -0,0 +1,158 @@
...
+        );
+        UTA_CACHED_INFO = get_method_id(
+            &env,
+            "com/exonum/binding/service/adapters/UserTransactionAdapter",
+            "info",
+            "()Ljava/lang/String;",
+        );
+        UTA_CACHED_VERIFY = get_method_id(
+            &env,
+            "com/exonum/binding/service/adapters/UserTransactionAdapter",
```
vitvakatu: I'm in doubt whether we should move this line (and the one with UserServiceAdapter to the constant value. Technically it will prevent from bigger diffs in case of any refactoring

exonum-java-binding-core/rust/src/utils/jni_cache.rs
```diff
@@ -0,0 +1,158 @@
...
+            "([B)Lcom/exonum/binding/service/adapters/UserTransactionAdapter;",
+        );
+        CLASS_CACHED_ERROR = env
+            .new_global_ref(env.find_class("java/lang/Error").unwrap().into())
+            .ok();
+        CLASS_CACHED_TEE = env
+            .new_global_ref(
+                env.find_class("com/exonum/binding/transaction/TransactionExecutionException")
+                    .unwrap()
+                    .into(),
```
vitvakatu: Formatting here is like a complete disaster 😕

exonum-java-binding-core/rust/src/utils/jni_cache.rs
```diff
@@ -0,0 +1,158 @@
...
+                && CLASS_CACHED_ERROR.is_some()
+                && CLASS_CACHED_TEE.is_some(),
+            "Error caching Java entities"
+        );
+
+        info!("Done caching references to Java classes and methods.");
+    }
+}
+
+// Helper method. Produces JMethodID for a particular class dealing with lifetime.
```
vitvakatu: It should be doc comment started with `///`

vitvakatu: Also I encourage to use backticks around every type in docstring and comments - `JMethodID`, not JMethodID

exonum-java-binding-core/rust/src/utils/jni_cache.rs
```diff
@@ -0,0 +1,158 @@
...
+fn get_method_id(env: &JNIEnv, class: &str, name: &str, sig: &str) -> Option<JMethodID<'static>> {
+    env.get_method_id(class, name, sig)
+        // we need this line to deal with lifetime
```
vitvakatu: To be more precise, `to erase lifetime in order to save underlying raw pointer in static`

exonum-java-binding-core/rust/src/utils/jni_cache.rs
```diff
@@ -0,0 +1,158 @@
...
+        .map(|mid| mid.into_inner().into())
+        .ok()
+}
+
+/// Returns cached `JClass` for "java/lang/Error" as a GlobalRef
```
vitvakatu: Here and below:
```suggestion
/// Returns cached `JClass` for `java/lang/Error` as a `GlobalRef`
```

exonum-java-binding-core/rust/src/utils/jni_cache.rs
skletsun: Statics are not exposed to the public so I believe it's OK. And I make method names short on purpose because of initial intention of using them prefixed with corresponding module's part. Like `jni_cache::get_error_class()`. This way it's kinda more clear that cached `error_class` comes from `jni_cache` entity. And this from my POV gives us an ability to just import the `jni_cache` in client classes and write prefixed methods instead of importing a bunch of explicit methods like `cache_get_uta_cached_execute`. So this was a reason I asked you if this approach is better or this is more clear to me only.

exonum-java-binding-core/rust/src/utils/errors.rs
skletsun: I explained my general idea of imports and methods names below in comments.

exonum-java-binding-core/rust/src/utils/jni_cache.rs
skletsun: I also in doubt in such cases (when there is only one line)

exonum-java-binding-core/rust/src/utils/jni_cache.rs
skletsun: But it comes from the regular format command :)

exonum-java-binding-core/rust/src/utils/jni_cache.rs
skletsun: Yeap I agree about backticks (just missed them in this case), but about doc comments - don't we do this for public methods only, do we?

exonum-java-binding-core/rust/integration_tests/tests/utils_errors.rs
skletsun: No, for the moment we're using `create_vm_for_tests_with_fake_classes`.

exonum-java-binding-core/rust/src/utils/jni_cache.rs
vitvakatu: Doc comments are invisible for private methods, but they must be supported by IDE, for example

exonum-java-binding-core/rust/src/utils/jni_cache.rs
```diff
@@ -0,0 +1,158 @@
...
+        .ok()
+}
+
+/// Returns cached `JClass` for `java/lang/Error` as a `GlobalRef`
+pub fn get_error_class() -> GlobalRef {
+    unsafe { CLASS_CACHED_ERROR.clone().unwrap() }
+}
+
+/// Returns cached `JClass` for `TransactionExecutionException` as a `GlobalRef`
+pub fn get_tee_class() -> GlobalRef {
```
vitvakatu: I believe `get_` prefixes are redundant in case you always import `jni_cache`

exonum-java-binding-core/rust/src/utils/jni_cache.rs
skletsun: Well, looking at our code we didn't follow this rule before, but if the general approach is like "not required, but nice to have" then it's better to add a spare slash ;-)

exonum-java-binding-core/rust/src/utils/jni_cache.rs
skletsun: Thanks, I also feel like that, let's also wait for anyone else's opinion.

exonum-java-binding-core/rust/integration_tests/tests/utils_errors.rs
dmitry-timofeev: I think what Ilya asks is if that method is used at all now.

exonum-java-binding-core/rust/integration_tests/tests/utils_errors.rs
```diff
@@ -22,8 +22,16 @@ const ARITHMETIC_EXCEPTION_CLASS_FQN: &str = "java.lang.ArithmeticException";
 const CUSTOM_EXCEPTION_MESSAGE: &str = "Test exception message";
 
 lazy_static! {
-    static ref VM: Arc<JavaVM> = create_vm_for_tests();
-    pub static ref EXECUTOR: MainExecutor = MainExecutor::new(VM.clone());
+    static ref VM: Arc<JavaVM> = create_vm_for_tests_with_fake_classes();
+    pub static ref EXECUTOR: MainExecutor = {
+        let ex = MainExecutor::new(VM.clone());
+        // JNI_OnLoad() is not called for these tests
```
dmitry-timofeev: Is it a bug or expected behaviour? It's not apparent from this comment/code.

exonum-java-binding-core/rust/src/proxy/service.rs
```diff
@@ -75,15 +75,16 @@ impl Service for ServiceProxy {
...
-            let java_service_hashes = panic_on_exception(
-                env,
-                env.call_method(
+            let java_service_hashes = panic_on_exception(env, unsafe {
+                env.call_method_unsafe(
                     self.service.as_obj(),
-                    "getStateHashes",
-                    "(J)[[B",
+                    jni_cache::get_usa_state_hashes(),
+                    JavaType::Array(Box::new(JavaType::Array(Box::new(JavaType::Primitive(
```
dmitry-timofeev: Can anyone say why jni-rs requires Box for Array enum variant?

dmitry-timofeev: Here and elsewhere: Why don't we add imports for this (_in this particular case,_ I'd even prefer `JavaType#from_str/parse("[[B")` to such a long and deeply nested chain)?

exonum-java-binding-core/rust/src/proxy/transaction.rs
```diff
@@ -136,7 +148,7 @@ fn check_transaction_execution_result<T>(
         JniErrorKind::JavaException => {
             let exception = get_and_clear_java_exception(env);
             let message = unwrap_jni(get_exception_message(env, exception));
-            if !unwrap_jni(env.is_instance_of(exception, CLASS_TRANSACTION_EXCEPTION)) {
+            if !unwrap_jni(env.is_instance_of(exception, &jni_cache::get_tee_class())) {
```
dmitry-timofeev: `tee` doesn't work :shirt:

exonum-java-binding-core/rust/src/utils/jni_cache.rs
dmitry-timofeev: For the repeated class FQN, I'd go for it. E.g., `TRANSACTION_ADAPTER_FQN`.

exonum-java-binding-core/rust/src/utils/jni_cache.rs
dmitry-timofeev: No, it's good

exonum-java-binding-core/rust/src/utils/jni_cache.rs
```diff
@@ -0,0 +1,158 @@
...
+    let env = vm.get_env().expect("Cannot get reference to the JNIEnv");
+    cache_methods(&env);
+
+    JNI_VERSION_1_8
+}
+
+/// Caches all required classes and methods ids.
+pub fn cache_methods(env: &JNIEnv) {
+    unsafe {
+        OBJECT_CACHED_GET_CLASS =
```
dmitry-timofeev: Is this implementation thread-safe? Is a write in the thread performing the initialization *guaranteed* to be visible in a thread performing a read?

exonum-java-binding-core/rust/src/utils/jni_cache.rs
```diff
@@ -0,0 +1,158 @@
...
+/// Helper method. Produces `JMethodID` for a particular class dealing with lifetime.
+fn get_method_id(env: &JNIEnv, class: &str, name: &str, sig: &str) -> Option<JMethodID<'static>> {
+    env.get_method_id(class, name, sig)
+        // we need this line to erase lifetime in order to save underlying raw pointer in static
+        .map(|mid| mid.into_inner().into())
+        .ok()
+}
+
+/// Returns cached `JClass` for `java/lang/Error` as a `GlobalRef`
+pub fn get_error_class() -> GlobalRef {
```
dmitry-timofeev: Wouldn't it be useful to group the methods by the class, as @bullet-tooth did for some constants in Java? So that the client code has a choice of 
```
// A method id:
transaction_adapter::execute_id()
// or, if transaction_adapter is apparent from the context
execute_id()
// or, if someone is crazy
jni_cache::transaction_adapter::execute_id()
--------------------------------
// A global class reference:
transaction_adapter::get_class() // <-- A naturally looking thing for Java devs
```

exonum-java-binding-core/rust/src/utils/jni_cache.rs
```diff
@@ -0,0 +1,158 @@
...
+    unsafe { CLASS_CACHED_ERROR.clone().unwrap() }
+}
+
+/// Returns cached `JClass` for `TransactionExecutionException` as a `GlobalRef`
+pub fn get_tee_class() -> GlobalRef {
```
dmitry-timofeev: :shirt:

exonum-java-binding-core/rust/src/utils/jni_cache.rs
dmitry-timofeev: Will it work in Rust?

exonum-java-binding-core/rust/src/utils/jni_cache.rs
dmitry-timofeev: I'd consider an option of grouping constants related to (some) classes together in a namespace, please see below. Don't know if it can be achieved easily :man_shrugging:

exonum-java-binding-core/rust/src/utils/jni_cache.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+/// Returns cached `JMethodID` for `UserTransactionAdapter.execute()`
+pub fn get_uta_execute() -> JMethodID<'static> {
+    unsafe { UTA_CACHED_EXECUTE.unwrap() }
+}
+
+/// Returns cached `JMethodID` for `UserTransactionAdapter.info()`
+pub fn get_uta_info() -> JMethodID<'static> {
+    unsafe { UTA_CACHED_INFO.unwrap() }
```
dmitry-timofeev: Is it possible to sprinkle some macros to remove the repeated code? Something like:

```rust
transaction_adapter {
  class_ref(UTA_CLASS_REF)
  method_id(verify, UTA_VERIFY_ID)
  method_id(execute, UTA_EXECUTE_ID)
  method_id(info, UTA_INFO_ID)
}
// becomes some four methods in a namespace?
```

@alekseysidorov, WDYT?

exonum-java-binding-core/rust/src/utils/jni_cache.rs
```diff
@@ -0,0 +1,158 @@
+use jni::{
+    objects::{GlobalRef, JMethodID},
+    sys::{jint, JNI_VERSION_1_8},
+    JNIEnv, JavaVM,
+};
+use std::os::raw::c_void;
+
+static mut OBJECT_CACHED_GET_CLASS: Option<JMethodID> = None;
```
dmitry-timofeev: If it is a cache, we can (probably) drop the CACHED (but I'm not sure). See also the comment about the namespaces.

exonum-java-binding-core/rust/src/proxy/transaction.rs
dmitry-timofeev: :tea:

exonum-java-binding-core/rust/src/proxy/service.rs
vitvakatu: Importing enum variants are usually considered bad practice. I agree that we can use `from_str` here.

dmitry-timofeev: … but, hurts the readability of the thing. 

Also, some of the examples in https://github.com/jni-rs/jni-rs/issues/137 are not correct :smile:

exonum-java-binding-core/rust/src/utils/jni_cache.rs
vitvakatu: 🤔 Actually, it's not thread-safe.

exonum-java-binding-core/rust/src/utils/jni_cache.rs
vitvakatu: Every proposed variant is possible.

exonum-java-binding-core/rust/src/utils/jni_cache.rs
vitvakatu: Possible, but from my point of view macros bring a few major issues:
1. They are not supported by IDE well, and therefore it's impossible to use code navigation, refactoring tools and (in most cases) auto-completion
2. Macros are much harder to read and understand while briefly looking over code.
3. In case of any special needs macros will no more be an option.

exonum-java-binding-core/rust/src/utils/jni_cache.rs
dmitry-timofeev: I meant if it will be easy enough to group these things in namespaces (whether they are namespaces or structs)? :upside_down_face:

exonum-java-binding-core/rust/integration_tests/tests/utils_errors.rs
skletsun: Yes, for the moment we're using it in three places of integrations tests when testing our executors.

exonum-java-binding-core/rust/src/utils/jni_cache.rs
skletsun: I also thought about this, but the initial requirement of this task was that change should not bring additional complexity to the existing code. I'd better implement such improvement as a separate task possibly with adding caching for some other entities. And yes, macros have pros and cons, to me, it looks like it worth trying if we need to cache more things.

exonum-java-binding-core/rust/src/utils/jni_cache.rs
dmitry-timofeev: For methods that do not have `public` visibility — use your judgement to decide if they need a documentation based on their frequency of use, complexity, their name, distance from their usage, etc. I would agree, however, that if you do add a comment to the method, make it a doc comment so that it can be easily looked up in an IDE (and gets appropriate highlighting). If anything isn't this way, it's a bug.

exonum-java-binding-core/rust/src/utils/jni_cache.rs
skletsun: Implemented it using Rust modules.

exonum-java-binding-core/rust/src/utils/jni_cache.rs
```diff
@@ -31,128 +31,157 @@ pub extern "system" fn JNI_OnLoad(vm: JavaVM, _: *mut c_void) -> jint {
...
-/// Returns cached `JMethodID` for `UserServiceAdapter.convertTransaction()`
-pub fn get_usa_convert_tx() -> JMethodID<'static> {
-    unsafe { USA_CACHED_CONVERT_TRANSACTION.unwrap() }
+/// Provides access to various cached classes.
+pub mod classes_refs {
+    use super::*;
+
+    /// Returns cached `JClass` for `java/lang/Error` as a `GlobalRef`.
+    pub fn java_lang_error() -> GlobalRef {
+        unsafe { JAVA_LANG_ERROR.clone().unwrap() }
```
dmitry-timofeev: 1. Why is it `unsafe`?

2. Do we clone an Option first, and then unwrap? Shan't we do the opposite?

vitvakatu: 1. Accessing mutable static variables is unsafe
2. The order doesn't matter because cloning `None` doesn't involve any computation.

dmitry-timofeev: The order might not matter in terms of performance, but in terms of readability it raises an unnecessary question why you clone Option.

dmitry-timofeev: That makes sense, thanks.

exonum-java-binding-core/rust/src/utils/jni_cache.rs
```diff
@@ -23,7 +28,7 @@ static mut TRANSACTION_EXECUTION_EXCEPTION: Option<GlobalRef> = None;
 #[no_mangle]
 pub extern "system" fn JNI_OnLoad(vm: JavaVM, _: *mut c_void) -> jint {
     let env = vm.get_env().expect("Cannot get reference to the JNIEnv");
-    cache_methods(&env);
+    CACHE_INIT.call_once(|| cache_methods(&env));
```
dmitry-timofeev: It is not a correct usage of once: https://doc.rust-lang.org/std/sync/struct.Once.html#examples-1

It won’t give you the memory guarantees that are needed in *reader* threads.

exonum-java-binding-core/rust/src/utils/jni_cache.rs
```diff
@@ -28,79 +43,91 @@ static mut TRANSACTION_EXECUTION_EXCEPTION: Option<GlobalRef> = None;
...
-                && SERVICE_ADAPTER_CONVERT_TRANSACTION.is_some()
-                && JAVA_LANG_ERROR.is_some()
-                && TRANSACTION_EXECUTION_EXCEPTION.is_some(),
-            "Error caching Java entities"
-        );
-
-        info!("Done caching references to Java classes and methods.");
+        if !CACHE_INITIALIZED {
+            cache_methods(env);
+            CACHE_INITIALIZED = true;
```
dmitry-timofeev: Why is a spin lock used above? Can't we just perform an atomic write here (Release), and an atomic read before each access (Acquire) as we don't expect concurrent *initializations*?

exonum-java-binding-core/rust/integration_tests/tests/jni_cache.rs
```diff
@@ -0,0 +1,72 @@
...
+    thread::spawn,
+};
+
+lazy_static! {
+    pub static ref VM: Arc<JavaVM> = create_vm_for_tests_with_fake_classes();
+    pub static ref EXECUTOR: MainExecutor = MainExecutor::new(VM.clone());
+}
+
+#[test]
+fn concurrent_cache_read() {
```
dmitry-timofeev: I'd consider explicitly stating that this test is not supposed to reliably catch synchronization errors because that will require _a lot_ more iterations, but that, in turn, requires starting in the initial state (which in case of statics, will unnecessarily complicate the thing).

exonum-java-binding-core/rust/integration_tests/benches/jni_cache.rs
```diff
@@ -0,0 +1,41 @@
...
+
+use integration_tests::mock::transaction::create_mock_transaction_proxy;
+use integration_tests::vm::create_vm_for_benchmarks_with_fakes;
+use java_bindings::exonum::blockchain::Transaction;
+use java_bindings::jni::JavaVM;
+use java_bindings::MainExecutor;
+
+use std::sync::Arc;
+use test::{black_box, Bencher};
+
```
dmitry-timofeev: I wouldn't name this benchamrk file `jni_cache` as this code does not measure its effect (there is no baseline you can compare to).

exonum-java-binding-core/rust/src/utils/jni_cache.rs
dmitry-timofeev: Even if we do expect, a lock would be required in initializer only; and it could be a simple lock, no need to DIY.

exonum-java-binding-core/rust/integration_tests/benches/jni_cache.rs
skletsun: I added a couple of benchmarks that measures non-cached implementations. And it says simple operations became about nine times faster.

exonum-java-binding-core/rust/src/utils/jni_cache.rs
skletsun: Simplified it with parking_lot::Once that actually does exactly what we want (locks for initialization and atomic reads for accesing the state of initialization)

exonum-java-binding-core/rust/integration_tests/benches/jni_cache.rs
```diff
@@ -20,22 +20,98 @@ extern crate java_bindings;
...
+}
+
+#[bench]
+pub fn is_instance_of_cached(b: &mut Bencher) {
+    EXECUTOR
+        .with_attached(|env| {
+            let exception = create_exception(env);
+            let tx_exec_exception_ref = jni_cache::classes_refs::transaction_execution_exception();
+
+            b.iter(|| black_box(env.is_instance_of(exception, &tx_exec_exception_ref)));
```
dmitry-timofeev: I guess it must access the cache on each iteration.

exonum-java-binding-core/rust/integration_tests/benches/jni_cache.rs
```diff
@@ -20,22 +20,98 @@ extern crate java_bindings;
...
+        })
+        .unwrap();
+}
+
+#[bench]
+pub fn is_instance_of_not_cached(b: &mut Bencher) {
+    EXECUTOR
+        .with_attached(|env| {
+            let exception = create_exception(env);
+            b.iter(|| black_box(env.is_instance_of(exception, ARITHMETIC_EXCEPTION_CLASS)));
```
dmitry-timofeev: Does `iter` perform measurements immediately (i.e., when you call `iter`, not when this code returns)?

Also, as this code does not check for results, it may return `Err`s and one will never notice, will they?

exonum-java-binding-core/rust/integration_tests/benches/jni_cache.rs
skletsun: Good point, thanks

exonum-java-binding-core/rust/integration_tests/benches/jni_cache.rs
skletsun: The `.iter()` method measures time spent to iterations. And that summary data gets processed (average, min, max) after return from this method.
And the method doesn't check for the result. But in our case, there should always be `true` I believe.
The general recommendation for writing benchmarks using the standard Rust abilities is to put all setup code outside of the `.iter()` method and only put the part one wants to measure inside.

dmitry-timofeev: > And the method doesn't check for the result.

Shan't it `unwrap` to ensure that we do get _some_ results and not an error, which is somewhat easy with JNI?

dmitry-timofeev:
Are benchmarks in stable Rust already? Are they compiled with tests? Do they run with tests (not the multiple measurements, but a single run to ensure they work)?

skletsun:
Benchmark still are in nightly Rust only, they don't compile and run with tests, unfortunately. For the moment, in order to check if they work, I manually modify the line in the `./run_native_integration_tests.sh` to `cargo +nightly bench`. There are a couple of external tools to make benchmarks on table Rust, the most popular is [Criterion.rs](https://bheisler.github.io/post/benchmarking-with-criterion-rs/) which we probably could try.

dmitry-timofeev:
OK, thanks! Can explore that later, especially if it provides any extra benefits, as the standard Rust tool is incredibly limited compared to the industry standard in Java, [JMH](https://openjdk.java.net/projects/code-tools/jmh/).

PR #580, opened 2018-11-28
Reproduce tx reception failure [ECR-2650]

dmitry-timofeev:
## Overview

Add a test of handling random transactions that reproduces the issue:

Add content type header to the request, equal to the value
sent by the frontend (application/x-www-form-urlencoded). 
This makes some tests fail.

---
See a more detailed description in the ticket: https://jira.bf.local/browse/ECR-2650


dmitry-timofeev:
@dip56 you might use that to develop a regression test, once a fix is integrated.

PR #583, opened 2018-11-29
Fix BinaryTransactionMessage equals and hashcode [ECR-2648]

bullet-tooth:
## Overview
Fixed position changing for internal byte buffer that impacts equals and hashcode methods at BinaryTransactionMessage.

---
See: https://jira.bf.local/browse/ECR-2648

### Definition of Done

- [x] Public API has Javadoc


bullet-tooth:
@MakarovS  good catch!

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/BinaryTransactionMessage.java
```diff
@@ -49,13 +49,15 @@
 
     this.rawTransaction = ByteBuffer.allocate(messageSize).order(LITTLE_ENDIAN);
     this.rawTransaction.put(slice);
+    resetPosition();
   }
 
   @Override
   public PublicKey getAuthor() {
```
bullet-tooth: Do methods that modifies BB state need to be thread safe?

dmitry-timofeev: That's a good question. As this class feels like a value, one might expect it to be. We can achieve that by duplicating the buffer before doing any operations that require modifying the buffer marks?

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/BinaryTransactionMessage.java
```diff
@@ -127,4 +120,16 @@ public String toString() {
     return Arrays.toString(rawTransaction.array());
   }
 
+  /**
+   * Returns bytes by slicing raw transaction to avoid changing it's position.
+   */
+  private byte[] getBytes(int startOffset, int size) {
+    byte[] bytes = new byte[size];
+    ByteBuffer slice = rawTransaction.slice();
```
dmitry-timofeev: Nit: `duplicate` will be a better fit because `slice` might generally have a smaller capacity than the original buffer, therefore, `position` won't work correctly. It won't happen in this case as position is suppossed to be always equal to 0, but one still have to think about it.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/BinaryTransactionMessage.java
```diff
@@ -88,10 +84,7 @@ public HashCode hash() {
   @Override
```
dmitry-timofeev: >      // We can't use BB directly for hashing because rawTransaction#position might be changed
     // and it causes having different hashes for the same message.
// and it causes having different hashes for the same message.

This comment doesn't seem correct any longer.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/message/BinaryTransactionMessage.java
```diff
@@ -77,7 +77,6 @@ public short getTransactionId() {
...
   public HashCode hash() {
     // We can't use BB directly for hashing because rawTransaction#position might be changed
```
dmitry-timofeev: This class does not change it.

PR #604, opened 2018-12-04
Implemented exceptions handling [ECR-2651]

bullet-tooth:
## Overview

---
See: https://jira.bf.local/browse/ECR-2651

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding-core/src/test/java/com/exonum/binding/storage/indices/ProofMapIndexProxyIntegrationTest.java
+++ b/exonum-java-binding-core/src/test/java/com/exonum/binding/storage/indices/ProofMapIndexProxyIntegrationTest.java
@@ -55,6 +55,7 @@ import com.exonum.binding.proxy.Cleaner;
 import com.exonum.binding.proxy.CloseFailuresException;
 import com.exonum.binding.storage.database.View;
 import com.exonum.binding.test.Bytes;
+import com.exonum.binding.test.CiOnly;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Streams;
@@ -373,6 +374,7 @@ class ProofMapIndexProxyIntegrationTest
     });
   }
 
+  @CiOnly
   @Test
   /*
     Takes quite a lot of time (validating 257 proofs), but it's an integration test, isn't it? :-)
--- a/exonum-java-testing/src/main/java/com/exonum/binding/test/CiOnly.java
+++ b/exonum-java-testing/src/main/java/com/exonum/binding/test/CiOnly.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2018 The Exonum Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


exonum-java-binding-core/src/main/java/com/exonum/binding/service/adapters/UserServiceAdapter.java
```diff
@@ -83,16 +84,25 @@ public String getName() {
    * @throws IllegalArgumentException if message is not a valid transaction message of this service
```
DENMROOT: No changes required for afterCommit() method ?

bullet-tooth:
@dmitry-timofeev not sure about extracting logging to separate class. There are not so much logging statements but it requires quite many changes and probably it not worth that.

exonum-java-binding-core/src/main/java/com/exonum/binding/service/adapters/UserTransactionAdapter.java
```diff
@@ -86,6 +86,6 @@ public String info() {
   }
 
   private void logUnexpectedException(Throwable e) {
-    logger.error("Unexpected exception:", e);
+    logger.error("Unexpected exception at transaction {}:", transaction, e);
```
dmitry-timofeev: in transaction?

exonum-java-binding-core/src/main/java/com/exonum/binding/service/adapters/UserServiceAdapter.java
```diff
@@ -129,21 +143,34 @@ public UserTransactionAdapter convertTransaction(byte[] transactionMessage) {
...
-    service.createPublicApiHandlers(node, router);
-    server.mountSubRouter(serviceApiPath(), router);
+    try {
+      checkState(node == null, "There is a node already: are you calling this method twice?");
+      node = new NodeProxy(nodeNativeHandle, viewFactory);
+      Router router = server.createRouter();
+      service.createPublicApiHandlers(node, router);
+      server.mountSubRouter(serviceApiPath(), router);
+    } catch (IllegalStateException e) {
+      logger.warn(e);
```
dmitry-timofeev: I'd not distinguish ISE and Exception and consider them the same error.

exonum-java-binding-core/src/main/java/com/exonum/binding/service/adapters/UserServiceAdapter.java
```diff
@@ -83,16 +84,25 @@ public String getName() {
...
+      assert message.getServiceId() == getId() :
+          "Message id is distinct from the service id";
+
+      Transaction transaction = service.convertToTransaction(message);
+      checkNotNull(transaction, "Invalid service implementation: "
+              + "Service#convertToTransaction must never return null.\n"
+              + "Throw an exception if your service does not recognize this message id (%s)",
+          message.getMessageType());
+      return new UserTransactionAdapter(transaction, viewFactory);
+    } catch (NullPointerException | IllegalArgumentException e) {
```
dmitry-timofeev: The NPE above is not a warning, but an error. On the other hand, if a service uses protobuf which allows **any** field to be optional, NPEs are likely in the user code (= and therefore may be regarded as "acceptable" in this case — if your transaction message does not contain a required parameter, than it is not valid). WDYT?

bullet-tooth: I think NPE is "acceptable" in this case. And because of that, I placed `warn`

PR #605, opened 2018-12-04
Fixed bug with core collection key serialization [ECR-2710]

bullet-tooth:
## Overview

---
It was a bug with key serialization. We used little-endian byte order, but core uses big-endian one.

See: https://jira.bf.local/browse/ECR-2710

### Definition of Done

- [ ] There is no TODOs left in the code
- [ ] Public API has Javadoc


exonum-java-binding-qa-service/src/main/java/com/exonum/binding/qaservice/QaServiceImpl.java
```diff
@@ -96,7 +96,7 @@ protected Schema createDataSchema(View view) {
     List<HashCode> stateHashes = super.getStateHashes(snapshot);
     // Log the state hashes, so that the values passed to the native part of the framework
     // are known.
-    logger.error("state hashes: {}", stateHashes);
+    logger.debug("state hashes: {}", stateHashes);
```
vitvakatu: Btw, it was one of the requirements of QA service (you can easily find it in wiki), not sure the reason though

dmitry-timofeev: Thanks, that's exactly the reason. If there isn't a comment clarifying that, please add.

bullet-tooth: Reverted this change. What is the reason for the `error` level requirement? Seeing errors in the log confuses me

bullet-tooth:
Probably yes, it can be tested using system tests only. Do you have any clue how can we cover such things by unit or integration tests?

dmitry-timofeev:
Probably, by writing to this (*internal core*) collection, but I'd abstain from that because that capability is likely to be temporary.

PR #615, opened 2018-12-13
Use StandardSerializers#protobuf [ECR-2712]

MakarovS:
## Overview
Use StandardSerializers#protobuf in protobuf serializers.
This PR is WIP and will be updated when ECR-2542 would be merged.

---
See: https://jira.bf.local/browse/ECR-2712

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding-cryptocurrency-demo/src/main/java/com/exonum/binding/cryptocurrency/HistoryEntitySerializer.java
```diff
@@ -16,18 +16,21 @@
...
 import com.exonum.binding.common.hash.HashCode;
 import com.exonum.binding.common.serialization.Serializer;
 import com.exonum.binding.cryptocurrency.transactions.TxMessageProtos;
 import com.google.protobuf.ByteString;
-import com.google.protobuf.InvalidProtocolBufferException;
 
 public enum HistoryEntitySerializer implements Serializer<HistoryEntity> {
   INSTANCE;
 
+  private static final Serializer<TxMessageProtos.HistoryEntity> historyEntityProtobufSerializer =
```
dmitry-timofeev: As Serializers do not have side effects, they are considered to be 'constants', hence, must use screaming snake case. Also, as scope is small, I'd consider a shorter name (PROTO_SERIALIZER), though this is good too.

MakarovS:
@dmitry-timofeev @bullet-tooth We could also use this method in our Tx classes - `CreateWalletTx`, `TransferTx` and transactions in QA service to get rid of `InvalidProtocolBufferException` exception handling. Do we want to?
@dip56 Do we have any integration tests that check protobuf serialization of our QA transactions?

dmitry-timofeev:
> Do we have any integration tests that check protobuf serialization of our QA transactions?

Currently the QA service creates transactions, not the test driver. Therefore, changes in their serialization must not affect the system tests. But may I ask why you are asking: the transaction data in services already uses protobuf, doesn't it?

> We could also use this method in our Tx classes - CreateWalletTx, TransferTx and transactions in QA service to get rid of InvalidProtocolBufferException exception handling. Do we want to?

I would not) These methods are likely to go away with separated messages.

PR #617, opened 2018-12-13
Use first four bytes of SHA-256 hash as the Block#hashCode [ECR-2739]

dmitry-timofeev:
## Overview

As SHA-256 already has a good distribution, there is no need
to include all fields in hashCode computation.

---
See: https://jira.bf.local/browse/ECR-2739

### Definition of Done

- [x] Public API has Javadoc


dmitry-timofeev:
Rebase onto `master`.

PR #623, opened 2018-12-17
Lombok prototype [ECR-2754]

MakarovS:
## Overview
Prototype to compare Lombok with AutoValue.

---
See: https://jira.bf.local/browse/ECR-2754

### Definition of Done

- [x] Public API has Javadoc


bullet-tooth:
Also lombok has another helpful features such us https://projectlombok.org/features/SneakyThrows
which can be used, for example, at `com.exonum.binding.storage.indices.BaseIndexGroupTestable#close`

bullet-tooth:
Checked the javadoc question 1 ⬆️ 
You can write the javadoc on the field. And then this test is copied to the getters, setters and builder's fields as is without  any modifications

dmitry-timofeev:
Thanks, that makes sense! Although it violates the code style, it will work in most cases, but if one needs to specify preconditions, exceptions, etc., then the second question arises.

MakarovS:
@dmitry-timofeev @bullet-tooth yes, you can have separate docs for getters and setters:
`
NEW in lombok v1.12.0: javadoc on the field will now be copied to generated getters and setters. Normally, all text is copied, and @return is moved to the getter, whilst @param lines are moved to the setter. Moved means: Deleted from the field's javadoc. It is also possible to define unique text for each
getter/setter. To do that, you create a 'section' named GETTER and/or SETTER. A section is a line in your javadoc containing 2 or more dashes, then the text 'GETTER' or 'SETTER', followed by 2 or more dashes, and nothing else on the line. If you use sections, @return and @param stripping for that section is no longer done (move the @return or @param line into the section).
`

Like this:
```
  /**
   * Name of the person.
   * -- SETTER --
   * Changes the name of this person.
   * 
   * @param name The new value.
   */
```

And literally just 1.5 months ago they fixed builder javadocs - now they are copied the same way as they are copied to setter, here's the PR https://github.com/rzwitserloot/lombok/pull/2008

If there are no more questions, we can close this PR.

dmitry-timofeev:
That's good, thanks!

PR #624, opened 2018-12-17
Add more blockchain accessors [ECR-2738]

dmitry-timofeev:
## Overview

- `containsBlock(Block) -> boolean`,
- `getBlock(long height) -> Block`

Also, rename `getBlock(HashCode)` to `findBlock` so that it is clearer that there might not be a block with the given hash.

---
See: https://jira.bf.local/browse/ECR-2738

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
+++ b/exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
@@ -98,7 +98,7 @@ public final class Blockchain {
    *
    * @param height block height starting from 0
    * @throws IllegalArgumentException if the height is invalid: negative or exceeding
-   *     {@linkplain #getHeight() the blockchain height}.
+   *     the {@linkplain #getHeight() blockchain height}
    */
   public ProofListIndexProxy<HashCode> getBlockTransactions(long height) {
     return schema.getBlockTransactions(height);
@@ -186,9 +186,9 @@ public final class Blockchain {
    * Returns the block at the given height.
    *
    * @param height the height of the block; must be non-negative and less than or equal to
-   *     the current blockchain height
+   *     the current {@linkplain #getHeight() blockchain height}
    * @return a block at the height
-   * @throws IllegalArgumentException
+   * @throws IllegalArgumentException if the height is not valid
    */
   public Block getBlock(long height) {
     checkHeight(height);
--- a/exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
+++ b/exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
@@ -200,10 +200,10 @@ public final class Blockchain {
   }
 
   private void checkHeight(long height) {
-    long maxHeight = getHeight();
-    if (height < 0 || height > maxHeight) {
-      throw new IndexOutOfBoundsException("Blockchain height (" + height + ") is out of range [0, "
-          + maxHeight + "]");
+    long blockchainHeight = getHeight();
+    if (height < 0 || height > blockchainHeight) {
+      throw new IndexOutOfBoundsException("Block height (" + height + ") is out of range [0, "
+          + blockchainHeight + "]");
     }
   }
 
--- a/exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
+++ b/exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
@@ -69,9 +69,9 @@ public final class Blockchain {
   }
 
   /**
-   * Returns the height of the latest committed block in the blockchain,
-   * or <em>blockchain height</em>. The height is a distance between the last block
-   * and the "genesis", or initial, block. Therefore, the height is equal to the number
+   * Returns the <em>blockchain height</em> which is the height of the latest committed block
+   * in the blockchain. The block height is a distance between the last block
+   * and the "genesis", or initial, block. Therefore, the blockchain height is equal to the number
    * of blocks plus one.
    *
    * <p>For example, the "genesis" block has height {@code h = 0}. The latest committed block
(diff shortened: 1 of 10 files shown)


exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/Blocks.java
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+
+package com.exonum.binding.blockchain;
+
+import com.exonum.binding.blockchain.serialization.BlockSerializer;
+import com.exonum.binding.common.hash.HashCode;
+import com.exonum.binding.common.hash.Hashing;
+
+public class Blocks {
```
MakarovS: `final`?

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/BlockchainIntegrationTest.java
```diff
@@ -0,0 +1,293 @@
...
+      MapIndex<HashCode, Block> blocks = MapIndexProxy.newInstance(TEST_BLOCKS, fork,
+          StandardSerializers.hash(), BlockSerializer.INSTANCE);
+      int blockHeight = 1;
+      block = withProperHash(aBlock(blockHeight)
+          .numTransactions(expectedBlockTransactions.size())
+          .build());
+      blocks.put(block.getBlockHash(), block);
+      when(coreSchema.getBlocks()).thenReturn(blocks);
+
+      // Setup tx hashes
```
dmitry-timofeev: This setup code is needed only for `getBlockTransactions` tests, hence `lenient`. It might be moved in another `@Nested` test class so that it is clear _which_ tests require what initialization code, WDYT about such approach?

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -55,11 +56,26 @@ public static Blockchain newInstance(View view) {
...
+   */
+  public boolean containsBlock(Block block) {
+    return findBlock(block.getBlockHash())
+        .map(block::equals)
+        .orElse(false);
+  }
+
+  /**
+   * Returns the height of the latest committed block in the blockchain,
+   * or <em>blockchain height</em>. The height is a distance between the last block
```
bullet-tooth:
```suggestion
   * in the other words <em>blockchain height</em>. The height is a distance between the last block
```

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/BlockchainIntegrationTest.java
```diff
@@ -0,0 +1,293 @@
...
+
+  /**
+   * Creates a builder of a block, fully initialized with defaults, inferred from the height.
+   * An invocation with the same height will produce exactly the same builder. The block hash
+   * is <strong>not</strong> equal to the hash of the block with such parameters.
+   *
+   * @param blockHeight a block height
+   * @return a new block builder
+   */
+  private static Builder aBlock(long blockHeight) {
```
bullet-tooth: what is the reason for having such method name? And configure checkstyle for it. How about `newBlock`?

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/Blocks.java
```diff
@@ -0,0 +1,47 @@
...
+        .height(block.getHeight())
+        .numTransactions(block.getNumTransactions())
+        .blockHash(actualBlockHash)
+        .previousBlockHash(block.getPreviousBlockHash())
+        .txRootHash(block.getTxRootHash())
+        .stateHash(block.getStateHash())
+        .build();
+  }
+
+  private Blocks() {}
```
bullet-tooth: sometime in such places we put `throw new AssertionError("Non-instantiable");`
But I not fully understand the benefit of such assertions. Reflection protection? 🤔

dmitry-timofeev: I'd say, a clear documentation of intent (= an idiomatic workaround to define a utility class); and self-protection from instantiation inside that class. However, I would not consider adding this assertion as a required thing in _smallish_ utility classes.

exonum-java-binding-core/src/test/java/com/exonum/binding/blockchain/BlockchainIntegrationTest.java
dmitry-timofeev: It's a kind of a commonly used pattern for naming methods in test classes that return a preconfigured builder with some defaults (hence 'a' = 'some', 'one about properties of which I do not care in this case') that are overridden in tests that need to set some properties.

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
dmitry-timofeev: Or the other way around: returns the blockchain height, which is the height of the latest committed block in the blockchain?

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -169,13 +182,38 @@ public long getHeight() {
...
+    checkHeight(height);
+
+    ListIndex<HashCode> blockHashes = getBlockHashes();
+    HashCode blockHash = blockHashes.get(height);
+    MapIndex<HashCode, Block> blocks = getBlocks();
+    return blocks.get(blockHash);
+  }
+
+  private void checkHeight(long height) {
+    long maxHeight = getHeight();
```
MakarovS: Use `blockHeight` instead of `height` and `blockchainHeight` instead of `maxHeight`?

PR #626, opened 2018-12-19
Use an empty string as 'no description' [ECR-2748]

dmitry-timofeev:
## Overview

It is believed to be slightly easier to use in cases where
an empty description is just fine: conversion to JSON, using in log
messages.

---
See: https://jira.bf.local/browse/ECR-2748

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/TransactionResult.java
+++ b/exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/TransactionResult.java
@@ -21,7 +21,6 @@ import static com.google.common.base.Preconditions.checkArgument;
 import com.exonum.binding.transaction.TransactionExecutionException;
 import com.google.auto.value.AutoValue;
 import com.google.common.base.Strings;
-import java.util.Optional;
 import java.util.OptionalInt;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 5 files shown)


exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/TransactionResult.java
```diff
@@ -114,10 +106,10 @@ private static TransactionResult valueOf(
   public abstract OptionalInt getErrorCode();
 
   /**
-   * Returns an optional description of a transaction if its execution resulted in an error.
-   * @return a description of an error, or {@code Optional.empty()} if there is no description
+   * Returns the description of a transaction if its execution resulted in an error;
+   * or an empty string if there is no description. Never {@code null}.
```
MakarovS: I think it's redundant here to explicitly state that it's never `null`. If method is not annotated as `@Nullable`, user should be confident that he won't get `null` from it.

exonum-java-binding-core/src/main/java/com/exonum/binding/blockchain/TransactionResult.java
```diff
@@ -21,15 +21,14 @@
...
-import java.util.Optional;
 import java.util.OptionalInt;
 import javax.annotation.Nullable;
 
 /**
  * Returns a result of the transaction execution. This result may be either a success, or an error,
  * if execution has failed.
  * Errors might be either service-defined or unexpected. Service-defined errors consist of an error
- * code and an optional description. Unexpected errors consist of an optional description only.
+ * code and an optional description. Unexpected errors include only a description.
```
MakarovS: `only an optional description`?

dmitry-timofeev: I decided to skip it here because it is mentioned above that a description can be optional.

PR #656, opened 2019-01-14
Fix the matchers:

dmitry-timofeev:
## Overview

`contains` turns out to take either a vararg of elements or matchers;
or a list of _matchers_ (but not the list of _elements_, as used to be
passed). Therefore, the code matched the first element against
a _list_, not its first element.

---

### Definition of Done

- [x] Public API has Javadoc
- [x] The continuous integration build passes (:running_man:)


dmitry-timofeev:
Hamcrest APIs turn out to be not that type-safe :upside_down_face:

dmitry-timofeev:
> The continuous integration build passes (running_man)

Java _integration_ tests pass locally, however, native ITs still do _not_.

PR #657, opened 2019-01-15
Fixed broken native tests [ECR-2794].

skletsun:
## Overview
Fixed issue with initialization of JNI cache in integration tests. Refactoring to make the code cleaner.

---
See: https://jira.bf.local/browse/ECR-2794

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ branches:
   only:
     - master
     - ejb-app-configuration
+    # TODO: remove
+    - new-message-format-1
     # Also build tagged revisions, e.g. v1.2.3-beta
     - /^v\d+\.\d+(\.\d+)?(-\S*)?$/
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,9 +47,9 @@ matrix:
       env: CHECK_RUST=false
     - name: "OSX JDK 8 CHECK_RUST=false"
       os: osx
-      # Specify the image containing JDK 8
+      # Specify the image containing JDK 8 (temporary switched to Java 11)
       # See: https://docs.travis-ci.com/user/reference/osx#os-x-version
-      osx_image: xcode9.3
+      osx_image: xcode10.1
       env: CHECK_RUST=false
     - name: "Linux JDK 11 CHECK_RUST=false"
       os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,7 +85,7 @@ before_install:
   - cargo install --list
   # Install OSX requirements
   # TODO: Temporary fix, the problem is described here: https://jira.bf.local/browse/ECR-2795
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install "$OSX_PACKAGES" || brew install "$OSX_PACKAGES"; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install $OSX_PACKAGES || brew install $OSX_PACKAGES; fi
   # force building instead of using from apt.
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then export ROCKSDB_BUILD=1; export SNAPPY_BUILD=1; fi
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,6 @@ branches:
   only:
     - master
     - ejb-app-configuration
-    # TODO: remove after merging this branch into master
-    - new-message-format-1
     # Also build tagged revisions, e.g. v1.2.3-beta
     - /^v\d+\.\d+(\.\d+)?(-\S*)?$/
 
@@ -47,9 +45,9 @@ matrix:
       env: CHECK_RUST=false
     - name: "OSX JDK 8 CHECK_RUST=false"
       os: osx
-      # Specify the image containing JDK 8 (temporary switched to Java 11)
+      # Specify the image containing JDK 8
       # See: https://docs.travis-ci.com/user/reference/osx#os-x-version
-      osx_image: xcode10.1
+      osx_image: xcode9.3
       env: CHECK_RUST=false
     - name: "Linux JDK 11 CHECK_RUST=false"
       os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ branches:
   only:
     - master
     - ejb-app-configuration
-    # TODO: remove
+    # TODO: remove after merging this branch into master
     - new-message-format-1
     # Also build tagged revisions, e.g. v1.2.3-beta
     - /^v\d+\.\d+(\.\d+)?(-\S*)?$/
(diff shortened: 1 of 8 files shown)


exonum-java-binding-core/rust/integration_tests/src/executor.rs
```diff
@@ -50,6 +54,18 @@ pub fn check_detached(vm: &JavaVM) {
     assert!(!is_attached(vm));
 }
 
+/// Creates MainExecutor and initializes JNI cache
+pub fn create_executor_with_cache_initialized(vm: Arc<JavaVM>) -> MainExecutor {
```
dmitry-timofeev: I was trying to come up with a better name, but came to a realisation that I don't understand why this is done when the executor is created and not a VM (and, therefore, requires a lengthy name)? The jni cache, as a vm, is supposed to be initialized whenever EJB classes are needed (with fakes), and only once, but you can create many executors and they don't imply a need for a cache.

Wouldn't it be better to initialize the cache once a VM is created _with_ fakes (= with efb-fakes classpath) in 
`fn create_vm(debug: bool, with_fakes: bool) -> JavaVM`?

vitvakatu: Looks reasonable

skletsun: Implemented. The one drawback is that it turned out `JavaVM#new()` detaches current thread just after VM creation and we have to re-attach it, but I believe in the context of integration tests this should not be a problem.

exonum-java-binding-core/rust/integration_tests/tests/jni_cache.rs
```diff
@@ -14,34 +14,23 @@
...
 };
 
-lazy_static! {
-    pub static ref VM: Arc<JavaVM> = create_vm_for_tests_with_fake_classes();
-    pub static ref EXECUTOR: MainExecutor = MainExecutor::new(VM.clone());
-}
 
 #[test]
 // NOTE: This test is not supposed to reliably catch synchronization errors.
 fn concurrent_cache_read() {
```
dmitry-timofeev: Will the test below always work? (cache_not_initialized)

Are these tests run in _separate_ processes?

exonum-java-binding-core/rust/integration_tests/src/vm.rs
```diff
@@ -72,7 +73,16 @@ fn create_vm(debug: bool, with_fakes: bool) -> JavaVM {
         .build()
         .unwrap_or_else(|e| panic!("{:#?}", e));
 
-    JavaVM::new(jvm_args).unwrap_or_else(|e| panic!("{:#?}", e))
+    let vm = JavaVM::new(jvm_args).unwrap_or_else(|e| panic!("{:#?}", e));
+
+    // Initialize JNI cache for testing with fakes
+    if with_fakes {
+        // Current implementation of JavaVM#new() detaches current thread so we have to re-attach it
```
dmitry-timofeev: Nit: this comment seems unnecessary given how the JNI API works in Rust (= Env guard)

.travis.yml
```diff
@@ -82,7 +84,7 @@ before_install:
   - cargo install --list
   # Install OSX requirements
   # TODO: Temporary fix, the problem is described here: https://jira.bf.local/browse/ECR-2795
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install libsodium rocksdb pkg-config || brew install libsodium rocksdb pkg-config; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install libsodium rocksdb pkg-config protobuf || brew install libsodium rocksdb pkg-config protobuf; fi
```
dmitry-timofeev: I am far from a bash expert, but will an environment variable work here (something like `brew install $OSX_PACKAGES`)?

exonum-java-binding-core/rust/integration_tests/tests/jni_cache.rs
skletsun: Well, it worked fine until now, but yes - probably it's a good idea to put it to a separate file. Tests are executed in the single process but in parallel in `RUST_TEST_THREADS` (or --test-threads) threads.

exonum-java-binding-core/rust/integration_tests/src/vm.rs
skletsun: I put this comment because this part doesn't follow the general way of JNI API. I expected that thread that creates JVM should become registered (moreover - the main thread according to the specification). But I was a little surprised by the explicit detach just after JVM creation in the `jni_rs`:
```
            jni_error_code_to_result(sys::JNI_CreateJavaVM(
                &mut ptr as *mut _,
                &mut env as *mut *mut sys::JNIEnv as *mut *mut c_void,
                args.inner_ptr(),
            ))?;

            let vm = Self::from_raw(ptr)?;
            java_vm_unchecked!(vm.0, DetachCurrentThread);
```

.travis.yml
skletsun: Good point, thanks.

exonum-java-binding-core/rust/integration_tests/src/vm.rs
dmitry-timofeev: > that thread that creates JVM should become registered 

Dunno, when you register a thread explicitly, you get a guard. Here you get a VM and no guard.

skletsun: Ok, sounds reasonable...

dmitry-timofeev:
@skletsun , Does it work locally? If so, could you please revert all the experimental changes for Travis, so that we can integrate the test fixes and procede with fixing platform-specific problems in a separate PR?

skletsun:
Yes, it works well on Linux and also on OSX but with java installed not by homebrew. Going to remove experimental changes.

exonum-java-binding-core/rust/integration_tests/benches/jni_cache.rs
```diff
@@ -41,15 +42,7 @@ const TX_EXEC_EXCEPTION_CLASS: &str =
...
-    pub static ref EXECUTOR: MainExecutor = {
-        let ex = MainExecutor::new(VM.clone());
-        ex.with_attached(|env| {
-            jni_cache::init_cache(env);
-            Ok(())
-        })
-        .unwrap();
-        ex
-    };
+    pub static ref EXECUTOR: MainExecutor = create_executor_with_cache_initialized(VM.clone());
```
dmitry-timofeev: How comes it works if there is no such method? 😕

skletsun: Oops, missed that. But as I said before, benches aren't compiled/executed within our current testing process.

dmitry-timofeev:
🎉 , thanks!

PR #658, opened 2019-01-15
Use an explicit per-View modification counter [ECR-2804]

dmitry-timofeev:
## Overview

Use an explicit modification counter in each View to get rid of
static state in ViewModificationCounter singleton and simplify
testing. As previously, a ModificationCounter is associated with each
View (either a Fork or Snapshot), but this relationship is now more
transparent.

---
See: https://jira.bf.local/browse/ECR-2804

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding-core/src/main/java/com/exonum/binding/storage/database/ModificationCounter.java
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Returns the current value of the counter. No assumptions must be made on how it changes
+   * when a notification is received.
+   */
+  int getCurrentValue();
+
+  /**
+   * Notifies this counter that the source object is modified, updating its current value.
+   *
+   * @throws IllegalStateException if this counter corresponds to a read-only (immutable) object,
```
dmitry-timofeev: Not sure if a counter shall concern itself with immutable things (on top of that, the AbstractIndexProxy does this check itself)

exonum-java-binding-core/src/test/java/com/exonum/binding/storage/database/ForkTest.java
```diff
@@ -31,22 +28,16 @@
 
 @PrepareForTest({
     Views.class,
-    ViewModificationCounter.class,
 })
 @Disabled
-// TODO Won't run on Java 10 till Powermock is updated [ECR-1614].
+// TODO Won't run on Junit 5 till Powermock is updated [ECR-1614].
```
dmitry-timofeev: to be more precise, till Powermock gains support of JUnit 5 (we use the most recent version)

exonum-java-binding-core/src/test/java/com/exonum/binding/storage/indices/AbstractIndexProxyTest.java
```diff
@@ -74,23 +57,31 @@ void constructorFailsIfNullView() {
...
+
+    Pattern pattern = Pattern.compile("Cannot modify the view: .*[Ss]napshot.*"
+        + "\\nUse a Fork to modify any collection\\.", Pattern.MULTILINE);
     assertThat(thrown.getMessage(), matchesPattern(pattern));
+
+    assertFalse(counter.isModifiedSince(initialValue));
   }
 
   @Test
   void notifyModifiedAcceptsFork() {
```
dmitry-timofeev: This test verifies that the index does use a counter from the Fork. Not sure if we shall add an integration test of Collection + Iterator + ModificationCounter, verifying that modification of another collection created with the same Fork invalidates the iterator of the first collection.

exonum-java-binding-core/src/main/java/com/exonum/binding/storage/database/ModificationCounter.java
```diff
@@ -0,0 +1,51 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.exonum.binding.storage.database;
+
+/**
+ * A counter of events-modifications of some objects (e.g., a collection, or a database view).
```
MakarovS: `A counter of modification events`?

exonum-java-binding-core/src/main/java/com/exonum/binding/storage/database/ModificationCounter.java
MakarovS: I think it's better to be safe and throw an exception in case of immutable counter (as it is now)

exonum-java-binding-core/src/main/java/com/exonum/binding/storage/indices/ConfigurableRustIter.java
```diff
@@ -32,8 +32,10 @@
 final class ConfigurableRustIter<E> extends AbstractNativeProxy implements RustIter<E> {
 
   private final LongFunction<E> nextFunction;
+  // todo: This view is kept only for diagnostic purposes (to put in error message) —
```
MakarovS: Views don't have `toString` overridden, so would it be helpful to log things like `com.exonum.binding.storage.database.Fork@131ef10`? Or maybe it would make sense to return `View#getViewNativeHandle` for diagnostic purposes? If no to both questions, then I'd remove this field.

exonum-java-binding-core/src/main/java/com/exonum/binding/storage/database/ModificationCounter.java
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * Returns true if the counter was modified since the given value (if {@link #notifyModified()}
+   * has been invoked); false — otherwise.
+   *
+   * @param lastValue the last value of the counter
+   */
+  boolean isModifiedSince(int lastValue);
+
+  // todo: We _might_ consider adding a special type for this timestamp, but at the moment it feels like an overkill
```
MakarovS: Maybe create a task instead of todo? Do you think it's realistic that a counter might overflow `int` range? If yes, we could add throwing an exception in `notifyModified` for such a case

exonum-java-binding-core/src/main/java/com/exonum/binding/storage/database/IncrementalModificationCounter.java
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.exonum.binding.storage.database;
+
+final class IncrementalModificationCounter implements ModificationCounter {
+
+  private static final int INITIAL_VALUE = 0;
```
MakarovS: Seems excessive to me, as we only use it once in the next line. Or maybe make it public? That way it'd be clear for users what the initial value is (because right now in tests to get `initialValue` we call `getCurrentValue` which is error-prone. We could use a public field in tests as well). But I do not insist, it's good as it is as well :)

exonum-java-binding-core/src/main/java/com/exonum/binding/storage/database/Fork.java
```diff
@@ -18,16 +18,16 @@
...
 import com.exonum.binding.proxy.ProxyDestructor;
 
 /**
  * A fork is a database view, allowing both read and write operations.
  *
- * <p>A fork allows to perform a transaction: a number of independent writes to a database,
- * which then may be <em>atomically</em> applied to the database state.
+ * <p>A fork allows to perform
+ * a {@linkplain com.exonum.binding.transaction.Transaction transaction}: a number of independent
+ * writes to a database, which then may be <em>atomically</em> applied to the database state.
```
bullet-tooth:
```suggestion
 * writes to the database, which then may be <em>atomically</em> applied (i.e. committed) to the database and change the database state.
```

exonum-java-binding-core/src/main/java/com/exonum/binding/storage/database/ModificationCounter.java
dmitry-timofeev: The reasoning was to hide the knowledge of how that timestamp is represented to be able to change later, but in an internal interface it seems to be unnecessary.

I've documented, however, that counters must detect up to 4B modifications, that shall be enough :-)

exonum-java-binding-core/src/main/java/com/exonum/binding/storage/database/IncrementalModificationCounter.java
dmitry-timofeev: I do not think that the clients _shall_ care about the initial value, what they shall is _current_ value (= the value at the moment when an (spl)iterator is bound to source). _initial_ value is an implementation detail of this counter.

The constnant can probably be inlined, and the field renamed to `counter`

PR #665, opened 2019-01-23
[DO NOT MERGE] Test building against updated RocksDB

skletsun:
## Overview
Just a testing PR to track the issue with tests that is reproducible only on Travis.

---


skletsun:
Closed because the problem seems fixed by #697

PR #667, opened 2019-01-24
Add time module [ECR-2811]

### LINKED ISSUE #447: Exonum time oracle service support

MakarovS:
## Overview
WIP: Add time service.

---
See: 
- https://jira.bf.local/browse/ECR-2811
- Closes #447 

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/CHANGELOG.md
+++ b/exonum-java-binding/CHANGELOG.md
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `--jvm-debug` command line argument that allows JDWP debugging of node. (#629)
 - Support of separated messages format Exonum v0.10. (#574)
 - `ListIndexProxy#stream` to enable stream processing of list elements. (#661)
+- Support of Time oracle. Instruction on how to enable built-in services can be found
+  [here](https://exonum.com/doc/version/0.4/get-started/java-binding/#built-in-services). (#667)
 
 ### Changed
 - `com.exonum.binding.storage.indices.MapEntry` moved to package
@@ -34,9 +36,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   `Generate-Config`. Also, the value of `--jvm-args-append` is not saved to any of the configuration
   files. (#629)
 - `Node#getPublicKey` to return `PublicKey` instead of `byte[]`. (#651)
-- `com.exonum.binding.transaction.Transaction#execute` now accepts 
+- `com.exonum.binding.transaction.Transaction#execute` now accepts
   `com.exonum.binding.transaction.TransactionContext`
-  instead of `com.exonum.binding.storage.database.View`  
+  instead of `com.exonum.binding.storage.database.View`
 
 ### Removed
 - `com.exonum.binding.common.proofs.map.MapEntry` — moved to package
@@ -44,7 +46,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `ViewModificationCounter` replaced with per-`View` modification counters to simplify
   their relationship and testing. (#658)
 - Exonum v0.9 message format related classes.
-  
+
 ### Fixed
 - A bug in the cryptocurrency demo frontend that sometimes resulted in rejected transactions and/or
 wrong response code (#621).
--- a/exonum-java-binding/core/rust/Cargo.lock
+++ b/exonum-java-binding/core/rust/Cargo.lock
@@ -29,7 +29,7 @@ dependencies = [
  "tokio-reactor 0.1.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "tokio-signal 0.2.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "tokio-tcp 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
- "tokio-timer 0.2.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tokio-timer 0.2.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "trust-dns-proto 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "trust-dns-resolver 0.10.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "uuid 0.7.2 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -46,7 +46,7 @@ dependencies = [
  "log 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "mio 0.6.16 (registry+https://github.com/rust-lang/crates.io-index)",
  "net2 0.2.33 (registry+https://github.com/rust-lang/crates.io-index)",
- "num_cpus 1.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "num_cpus 1.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "slab 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "tokio 0.1.15 (registry+https://github.com/rust-lang/crates.io-index)",
  "tokio-codec 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -54,7 +54,7 @@ dependencies = [
  "tokio-io 0.1.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "tokio-reactor 0.1.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "tokio-tcp 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
- "tokio-timer 0.2.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tokio-timer 0.2.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "tower-service 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "trust-dns-resolver 0.10.3 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
@@ -86,12 +86,12 @@ dependencies = [
  "mime_guess 2.0.0-alpha.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "mio 0.6.16 (registry+https://github.com/rust-lang/crates.io-index)",
  "net2 0.2.33 (registry+https://github.com/rust-lang/crates.io-index)",
- "num_cpus 1.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "num_cpus 1.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "parking_lot 0.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "percent-encoding 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand 0.6.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 1.0.85 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.87 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.38 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_urlencoded 0.5.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "sha1 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -103,7 +103,7 @@ dependencies = [
  "tokio-io 0.1.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "tokio-reactor 0.1.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "tokio-tcp 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
- "tokio-timer 0.2.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tokio-timer 0.2.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "url 1.7.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "v_htmlescape 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "version_check 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -129,7 +129,7 @@ name = "aho-corasick"
 version = "0.6.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "memchr 2.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr 2.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -229,7 +229,7 @@ dependencies = [
  "rand 0.3.23 (registry+https://github.com/rust-lang/crates.io-index)",
  "rust-crypto 0.2.36 (registry+https://github.com/rust-lang/crates.io-index)",
  "secp256k1 0.11.6 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 1.0.85 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.87 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -267,7 +267,7 @@ dependencies = [
  "pretty_assertions 0.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "secp256k1 0.11.6 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 1.0.85 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.87 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_str 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -341,7 +341,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "num-integer 0.1.39 (registry+https://github.com/rust-lang/crates.io-index)",
  "num-traits 0.2.6 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 1.0.85 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.87 (registry+https://github.com/rust-lang/crates.io-index)",
  "time 0.1.42 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -377,13 +377,13 @@ dependencies = [
 
 [[package]]
 name = "combine"
-version = "3.6.6"
+version = "3.6.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "ascii 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "byteorder 1.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "either 1.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "memchr 2.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr 2.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "unreachable 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -445,7 +445,7 @@ dependencies = [
  "crossbeam-epoch 0.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "crossbeam-utils 0.6.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 1.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "num_cpus 1.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "num_cpus 1.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "parking_lot 0.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -591,10 +591,11 @@ version = "0.5.0-SNAPSHOT"
 dependencies = [
  "exonum-btc-anchoring 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "exonum-configuration 0.10.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "exonum-time 0.10.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "java_bindings 0.5.0-SNAPSHOT",
- "serde 1.0.85 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde_derive 1.0.85 (registry+https://github.com/rust-lang/crates.io-index)",
- "tempfile 3.0.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.87 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_derive 1.0.87 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tempfile 3.0.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "toml 0.4.10 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -657,7 +658,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "encoding_rs"
-version = "0.8.15"
+version = "0.8.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "cfg-if 0.1.6 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -677,10 +678,10 @@ dependencies = [
 
 [[package]]
 name = "erased-serde"
-version = "0.3.8"
+version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "serde 1.0.85 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.87 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -715,7 +716,7 @@ dependencies = [
  "clap 2.32.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "ctrlc 3.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "env_logger 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "erased-serde 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "erased-serde 0.3.9 (registry+https://github.com/rust-lang/crates.io-index)",
  "exonum-build 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "exonum-crypto 0.10.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "exonum-derive 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -730,8 +731,8 @@ dependencies = [
  "rand 0.6.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand_xorshift 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "rust_decimal 0.10.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 1.0.85 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde_derive 1.0.85 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.87 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_derive 1.0.87 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.38 (registry+https://github.com/rust-lang/crates.io-index)",
  "snow 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "tokio 0.1.15 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -770,8 +771,8 @@ dependencies = [
  "protobuf 2.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "secp256k1 0.11.6 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 1.0.85 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde_derive 1.0.85 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.87 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_derive 1.0.87 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.38 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_str 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "structopt 0.2.14 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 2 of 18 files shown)


exonum-java-binding-time-service/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>com.exonum.binding</groupId>
+            <artifactId>exonum-java-binding-core</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>${assertj.version}</version>
```
bullet-tooth: use inheritance from the parent.

exonum-java-binding-time-service/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+        <dependencies>
+            <dependency>
+                <groupId>com.exonum.binding</groupId>
+                <artifactId>exonum-java-binding-bom</artifactId>
+                <version>${exonum.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+
+            <dependency>
```
bullet-tooth: it's not required. Such as we inherit whole `dependencyManagement` from the patent.

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/TimeSchema.java
```diff
@@ -0,0 +1,44 @@
...
+   * Returns stored time.
+   */
+  EntryIndexProxy<ZonedDateTime> getTime();
+
+  /**
+   * Returns the table that stores time for every validator.
+   */
+  ProofMapIndexProxy<PublicKey, ZonedDateTime> getValidatorsTimes();
+
+  List<HashCode> getStateHashes();
```
bullet-tooth: Put `@Override` or remove

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/TimeSchemaProxy.java
```diff
@@ -0,0 +1,89 @@
...
+import com.exonum.binding.storage.indices.EntryIndexProxy;
+import com.exonum.binding.storage.indices.ProofMapIndexProxy;
+
+import java.time.ZonedDateTime;
+import java.util.Collections;
+import java.util.List;
+
+public class TimeSchemaProxy implements TimeSchema {
+
+  private static final Serializer<PublicKey> publicKeySerializer = StandardSerializers.publicKey();
```
bullet-tooth: upper cases

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/TimeSchemaProxy.java
```diff
@@ -0,0 +1,89 @@
...
+  private static final Serializer<ZonedDateTime> zonedDateTimeSerializer = StandardSerializers.zonedDateTime();
+
+  private final NativeHandle nativeHandle;
+  private final View dbView;
+
+  private TimeSchemaProxy(NativeHandle nativeHandle, View dbView) {
+    this.nativeHandle = nativeHandle;
```
bullet-tooth: unused

CHANGELOG.md
```diff
@@ -22,6 +22,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   that is invoked after each block commit event. (#550)
 - Support of Json serialization in a common way. (#611)
 - Added the `--jvm-debug` command line argument that allows JDWP debugging of node. (#629)
+- Support of Time oracle.
+
+  To enable this service put `ejb_app_services.toml` file into EJB App's directory
+  with the following content:
+  ```toml
+  services = ["configuration", "time"]
+  ``` (#667)
```
dmitry-timofeev: This blood-red seems to be for a reason :-)

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/TimeSchema.java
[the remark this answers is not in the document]
MakarovS: As we discussed, it won't be needed for now.

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/TimeSchema.java
```diff
@@ -0,0 +1,44 @@
...
+import java.time.ZonedDateTime;
+import java.util.List;
+
+/**
+ * Exonum time service database schema.
+ */
+public interface TimeSchema extends Schema {
+
+  /**
+   * Returns stored time.
```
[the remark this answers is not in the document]
MakarovS: Will add more 👌

exonum-java-binding-core/rust/src/storage/time_schema.rs
```diff
@@ -0,0 +1,52 @@
...
+use utils::{self, Handle};
+
+enum SchemaType {
+    SnapshotSchema(TimeSchema<&'static Snapshot>),
+    ForkSchema(TimeSchema<&'static mut Fork>),
+}
+
+/// Returns pointer to created TimeSchemaProxy object
+#[no_mangle]
+pub extern "system" fn Java_com_exonum_binding_time_TimeSchemaProxy_nativeCreate(
```
dmitry-timofeev: Why is this needed at all if there is no object providing _any_ useful operations?

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/TimeSchema.java
```diff
@@ -0,0 +1,44 @@
...
+import com.exonum.binding.service.Schema;
+import com.exonum.binding.storage.indices.EntryIndexProxy;
+import com.exonum.binding.storage.indices.ProofMapIndexProxy;
+import java.time.ZonedDateTime;
+import java.util.List;
+
+/**
+ * Exonum time service database schema.
+ */
+public interface TimeSchema extends Schema {
```
dmitry-timofeev: Why does it extend the Schema?

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/TimeSchemaProxy.java
```diff
@@ -0,0 +1,86 @@
...
+  private final View dbView;
+
+  private TimeSchemaProxy(View dbView) {
+    this.dbView = dbView;
+  }
+
+  /**
+   * Constructs a schema proxy for a given dbView.
+   */
+  static TimeSchemaProxy newInstance(View dbView) {
```
dmitry-timofeev: Why does it need to be a proxy of a native object, if that native object does not provide any useful operations?

exonum-java-binding-time-service/src/test/java/com/exonum/binding/time/TimeSchemaProxyIntegrationTest.java
```diff
@@ -0,0 +1,57 @@
...
+    assertSchema((schema) -> assertFalse(schema.getTime().isPresent()));
+  }
+
+  @Test
+  void getValidatorsTimesTest() {
+    assertSchema((schema) -> assertFalse(schema.getValidatorsTimes().entries().hasNext()));
+  }
+
+  private static void assertSchema(Consumer<TimeSchemaProxy> assertion) {
+    try (MemoryDb db = MemoryDb.newInstance(); Cleaner cleaner = new Cleaner()) {
```
dmitry-timofeev: Please do not put several statements in a single line: https://google.github.io/styleguide/javaguide.html#s4.3-one-statement-per-line

exonum-java-binding-time-service/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+        <version>0.4-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>exonum-java-binding-time-service</artifactId>
+    <packaging>jar</packaging>
+
+    <name>Exonum Time Service</name>
+    <url>https://exonum.com/doc/version/latest/advanced/time/</url>
+
+    <properties>
```
dmitry-timofeev: This POM does not seem to be formatted according to our conventions (use 2 spaces, and apply tidy-pom).

exonum-java-binding-time-service/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+        <ejb-core.nativeLibPath>${project.parent.basedir}/exonum-java-binding-core/rust/target/debug
+        </ejb-core.nativeLibPath>
+        <junit.jupiter.version>5.3.2</junit.jupiter.version>
+    </properties>
+
+    <dependencyManagement>
```
dmitry-timofeev: Why do we need 'depManagement' and EJB-BOM? Aren't all dependencies already listed in 'depManagement' in the *parent*?

exonum-java-binding-time-service/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <argLine>
+                        ${jacoco.args}
+                        -Djava.library.path=${ejb-core.nativeLibPath}
+                        ${java.vm.assertionFlag}
+                    </argLine>
+                </configuration>
+            </plugin>
+        </plugins>
```
dmitry-timofeev: Just to confirm: this project won't be released if we do it now, will it?

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/TimeSchemaProxy.java
vitvakatu: Well, this code is using approach with constructing time service collections by their names. I like this approach because it's much simpler (and in fact we don't need any Rust code at all), but from future perspective I think we need to encapsulate collections inside the Rust proxy.

dmitry-timofeev: We will need that at one point, but I don't agree with putting some code that _pretends_ to do something, but doesn't until some time in the future (that might not come). Therefore I'd either keep things as simple as they need to be now, or finish the thing to create collections through a TimeSchema _in Rust_. The latter will also require adding appropriate factories to Java collection proxies, because now you can only create them by name, not from a random handle.

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/TimeSchema.java
```diff
@@ -17,28 +17,28 @@
...
 
 /**
  * Exonum time service database schema.
  */
-public interface TimeSchema extends Schema {
+public interface TimeSchema {
 
   /**
-   * Returns stored time.
+   * Consolidated time output by the service, which can be used by other business logic on the
```
dmitry-timofeev: The summary must start with a verb, e.g., "Returns ..."

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/TimeSchema.java
```diff
@@ -17,28 +17,28 @@
...
+   * blockchain.
+   * At the time when a new blockchain is launched, the consolidated time is unknown until the
+   * transactions from at least 2f + 1 validator nodes are processed.
+   *
+   * @throws NoSuchElementException if there is no consolidated time
```
dmitry-timofeev: Is it true? This does not seem right to return ~ Optional *and* throw an exception if there is no value present.

I'd keep this method as is, not throwing exceptions, and consider (maybe, not in this PR) adding a method that _requires_ that the time is known and throws otherwise, like `getRequiredTime`.

exonum-java-binding-time-service/pom.xml
MakarovS: Thanks, added `<skip>true</skip>` for this module.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/serialization/ZonedDateTimeSerializer.java
```diff
@@ -17,22 +17,43 @@
...
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.time.Instant;
+import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 
 enum ZonedDateTimeSerializer implements Serializer<ZonedDateTime> {
   INSTANCE;
 
+  private static final int BYTE_BUFFER_CAPACITY = Long.BYTES + Integer.BYTES;
```
dmitry-timofeev: It is used as a byte buffer capacity, but it seems to rather be "serialized date-time size".

exonum-java-binding-common/src/main/java/com/exonum/binding/common/serialization/ZonedDateTimeSerializer.java
```diff
@@ -17,22 +17,43 @@
...
+    ByteBuffer buffer = ByteBuffer.wrap(serializedValue)
+        .order(ByteOrder.LITTLE_ENDIAN);
+    return retrieveZdtFromBuffer(buffer);
   }
 
+  private ZonedDateTime retrieveZdtFromBuffer(ByteBuffer buffer) {
+    long seconds = buffer.getLong();
+    int nanos = buffer.getInt();
+    Instant instant = Instant.ofEpochSecond(seconds, nanos);
+    return ZonedDateTime.ofInstant(instant, ZoneOffset.UTC);
```
dmitry-timofeev: Is this serializer publicly available (through a method, as it is package-private)? If so, the fact that it _ignores_ time zone during serialization does not seem right _in a general-purpose_ date time serializer. 

I think we shall either
1. Change it to the serializer of Instant (and convert instant from core to utc somewhere else).
2. Make it internal, change the name to UtcZonedDTS, add a check that passed DTS is indeed in UTC.

The first might still require a serializer, because we provide direct access to indices (cannot convert anywhere but in serializer), hence the latter seems to be the path to take.

exonum-java-binding-common/src/test/java/com/exonum/binding/common/serialization/ZonedDateTimeSerializerTest.java
```diff
@@ -0,0 +1,51 @@
...
+import com.exonum.binding.test.Bytes;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+class ZonedDateTimeSerializerTest {
+  private Serializer<ZonedDateTime> serializer = ZonedDateTimeSerializer.INSTANCE;
```
dmitry-timofeev: A test of non-UTC?

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/TimeSchemaImpl.java
```diff
@@ -24,22 +24,28 @@
 import com.exonum.binding.storage.indices.ProofMapIndexProxy;
 import java.time.ZonedDateTime;
 
-public class TimeSchemaProxy implements TimeSchema {
+public class TimeSchemaImpl implements TimeSchema {
```
dmitry-timofeev: IMO, the previous (concrete) name was better than generic 'impl' which gives zero clue about the class.

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/TimeSchemaImpl.java
```diff
@@ -53,6 +59,15 @@ static TimeSchemaProxy newInstance(View dbView) {
         ZONED_DATE_TIME_SERIALIZER);
   }
 
+  private void checkIfEnabled() {
+    if (!isTimeServiceEnabled()) {
```
dmitry-timofeev: checkState(isTimeServiceEnabled(), "…");

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/TimeSchemaImpl.java
```diff
@@ -53,6 +59,15 @@ static TimeSchemaProxy newInstance(View dbView) {
...
+      throw new IllegalStateException("Time service is not enabled. To enable it, put 'time' into " +
+          "'ejb_app_services.toml' file.");
+    }
+  }
+
+  native boolean isTimeServiceEnabled();
```
dmitry-timofeev: Why isn't it private?

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/TimeSchemaImpl.java
```diff
@@ -24,22 +24,28 @@
...
+public class TimeSchemaImpl implements TimeSchema {
 
   private static final Serializer<PublicKey> PUBLIC_KEY_SERIALIZER = StandardSerializers.publicKey();
   private static final Serializer<ZonedDateTime> ZONED_DATE_TIME_SERIALIZER = StandardSerializers.zonedDateTime();
 
   private final View dbView;
 
-  private TimeSchemaProxy(View dbView) {
+  private TimeSchemaImpl(View dbView) {
+    checkIfEnabled();
```
dmitry-timofeev: Will it work in ITs?

dmitry-timofeev: 1. In ITs of this class?
2. In ITs that happen to use TimeSchema?

MakarovS: Will fix after native part implementation

dmitry-timofeev: But what is the specified behaviour? It can't be arbitrary at this moment.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/serialization/StandardSerializers.java
```diff
@@ -153,7 +153,7 @@
    * Returns a serializer of ZonedDateTime instances.
    */
   public static Serializer<ZonedDateTime> zonedDateTime() {
```
dmitry-timofeev: Why is this public even though this class is suppossed to be internal (especially with no documentation of its limitations)?

exonum-java-binding-common/src/main/java/com/exonum/binding/common/serialization/UtcZonedDateTimeSerializer.java
```diff
@@ -18,23 +18,25 @@
...
-enum ZonedDateTimeSerializer implements Serializer<ZonedDateTime> {
+enum UtcZonedDateTimeSerializer implements Serializer<ZonedDateTime> {
   INSTANCE;
 
-  private static final int BYTE_BUFFER_CAPACITY = Long.BYTES + Integer.BYTES;
+  private static final int SERIALIZED_ZDT_SIZE = Long.BYTES + Integer.BYTES;
 
   @Override
   public byte[] toBytes(ZonedDateTime value) {
+    checkArgument(value.getZone() == ZoneOffset.UTC, "Serialized ZonedDateTime value should be in UTC");
```
dmitry-timofeev: 1. Please add the actual value
2. Remove "Serialized" as the value is just an object.

exonum-java-binding-common/src/main/java/com/exonum/binding/common/serialization/UtcZonedDateTimeSerializer.java
```diff
@@ -18,23 +18,25 @@
...
 import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 
-enum ZonedDateTimeSerializer implements Serializer<ZonedDateTime> {
+enum UtcZonedDateTimeSerializer implements Serializer<ZonedDateTime> {
   INSTANCE;
 
-  private static final int BYTE_BUFFER_CAPACITY = Long.BYTES + Integer.BYTES;
+  private static final int SERIALIZED_ZDT_SIZE = Long.BYTES + Integer.BYTES;
```
dmitry-timofeev: 1. The abbreviation looks confusing.
2. It is a lie — no zone is included.

...DATE_TIME_SIZE?

exonum-java-binding-time-service/src/test/java/com/exonum/binding/time/TimeSchemaProxyIntegrationTest.java
```diff
@@ -29,8 +29,9 @@
 import java.util.function.Consumer;
 import org.junit.jupiter.api.Test;
 
+//TODO: fix after native part implementation
```
dmitry-timofeev: Please `@Disable/Ignore` it, leaving the TODO, so that this PR builds and can be integrated without native part.

exonum-java-binding-core/rust/src/utils/time_service.rs
```diff
@@ -0,0 +1,95 @@
...
+pub fn init_with_blockchain(blockchain: Blockchain) {
+    unsafe {
+        INIT.call_once(|| BLOCKCHAIN = Some(blockchain));
+    }
+}
+
+fn is_time_service_enabled() -> bool {
+    INIT.state().done() &&
+    unsafe { BLOCKCHAIN.clone() }.map_or(false, |blockchain|{
+        blockchain.service_map().get(&TIME_SERVICE_ID).is_some()
```
vitvakatu: Unfortunately this won't work properly in some scenarios... As service IDs are not fixed in Exonum, any user can create an own service with SERVICE_ID=4, hence this check will return true. 🤔

skletsun: Yes, this is rather an open topic on how to check it. Since the user is able to provide the same name as for time service, probably it worth checking the exact type (if possible) as well.

dmitry-timofeev: But they are fixed in the services we include in the application, aren't they (configuration, btc-anchoring & time)? If so, it might be reasonable to forbid users to set such ids for their services.

exonum-java-binding-core/rust/src/utils/services.rs
```diff
@@ -0,0 +1,118 @@
...
+    let mut toml = String::new();
+    file.read_to_string(&mut toml)?;
+    let ServicesToEnable { services } =
+        toml::from_str(&toml).expect("Invalid list of services to enable");
+    Ok(services)
+}
+
+/// Determines whether particular service name is defined in the services configuration file.
+/// TODO: For the moment we're just checking for the service's name presence in the configuration
+/// file. As soon as we have dynamic services implemented this checking should happen in runtime.
```
vitvakatu: nit: 
```suggestion
/// file. As soon as we have dynamic services implemented this check should happen at runtime.
```

exonum-java-binding-core/rust/src/utils/services.rs
```diff
@@ -0,0 +1,118 @@
...
+pub fn is_service_enabled(service_name: &str) -> bool {
```
vitvakatu: 🤔 What about using enum instead of string-API?

exonum-java-binding-core/rust/src/utils/services.rs
```diff
@@ -0,0 +1,118 @@
...
+    is_service_enabled_in_file(service_name, PATH_TO_SERVICES_TO_ENABLE)
```
vitvakatu: wow. such java. so generic. much enterprise :D

exonum-java-binding-core/rust/src/utils/services.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+fn is_service_enabled_in_file<P: AsRef<Path>>(service_name: &str, path: P) -> bool {
+    load_enabled_services(path)
+        .map(|services| {
+            for service in services {
```
vitvakatu: nit:
```rust
services.iter().any(|service| service.trim() == service_name)
```

Matter of taste though

exonum-java-binding-core/rust/src/utils/services.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    #[test]
+    fn no_config() {
+        let res = load_enabled_services("nonexistent");
+        assert!(res.is_err());
+    }
+
+    #[test]
+    fn all_services() {
```
vitvakatu: 🤔 Tests look pretty similar to the ones in ejb_app. Shan't we remove one of them?

exonum-java-binding-core/rust/src/utils/services.rs
skletsun: I was also thinking about this but decided to make things simple because we anyway have to re-write this when we have dynamic services.

exonum-java-binding-core/rust/src/utils/services.rs
skletsun: Well, otherwise unit-testing it with temporary files (like /some/where/deeeeply/in/tmp/app_services_random_blah_blah.toml_still_random) becomes a little more complicated :) And on the other hand, I'm almost sure that compiler will inline that function in production code, so... :)

exonum-java-binding-core/rust/src/utils/services.rs
skletsun: Well, perhaps it would be reasonable to clean up possible duplications in the node_builder concentrating more on testing logic that belongs to node builder (like adding defaults values).

exonum-java-binding-core/rust/src/utils/services.rs
```diff
@@ -0,0 +1,115 @@
...
+        .unwrap_or(false)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::io::Write;
+    use tempfile::{Builder, TempPath};
+
+    fn create_config(filename: &str, cfg: &str) -> TempPath {
```
dmitry-timofeev: Please move to the bottom (after callers)

exonum-java-binding-core/rust/src/utils/time_service.rs
```diff
@@ -12,123 +12,16 @@
...
-        assert!(!is_time_service_enabled());
-    }
-
-    #[test]
-    fn initialized_time_service_presented() {
-        let blockchain = create_blockchain(vec![Box::new(EmptyService), Box::new(StubTimeService)]);
-        init_with_blockchain(blockchain);
-        assert!(is_time_service_enabled());
-    }
+    is_service_enabled(TIME_SERVICE).into()
```
dmitry-timofeev: Is this code seriously loads a file on each invocation (each time a TimeSchema is accessed)?

exonum-java-binding-core/rust/src/utils/services.rs
```diff
@@ -0,0 +1,115 @@
...
+    file.read_to_string(&mut toml)?;
+    let ServicesToEnable { services } =
+        toml::from_str(&toml).expect("Invalid list of services to enable");
+    Ok(services)
+}
+
+/// Determines whether particular service name is defined in the services configuration file.
+/// TODO: For the moment we're just checking for the service's name presence in the configuration
+/// file. As soon as we have dynamic services implemented this checking should happen in runtime.
+pub fn is_service_enabled(service_name: &str) -> bool {
```
dmitry-timofeev: I can't agree with hiding of loading a file under such a name.

exonum-java-binding-core/rust/src/utils/services.rs
```diff
@@ -0,0 +1,115 @@
...
+    is_service_enabled_in_file(service_name, PATH_TO_SERVICES_TO_ENABLE)
+}
+
+fn is_service_enabled_in_file<P: AsRef<Path>>(service_name: &str, path: P) -> bool {
+    load_enabled_services(path)
+        .map(|services| {
+            services
+                .iter()
```
dmitry-timofeev: Does this code iterate over hashset (O(N)), when it can _contains_ (O(1))?

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/serialization/UtcZonedDateTimeSerializer.java
```diff
@@ -15,29 +15,32 @@
  *
  */
 
-package com.exonum.binding.common.serialization;
+package com.exonum.binding.time.serialization;
```
dmitry-timofeev: Can we put it in `com.exonum.binding.time` and make package-private?

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/serialization/UtcZonedDateTimeSerializer.java
```diff
@@ -15,29 +15,32 @@
...
 
-  private static final int SERIALIZED_ZDT_SIZE = Long.BYTES + Integer.BYTES;
+  private static final int SERIALIZED_DATE_TIME_SIZE = Long.BYTES + Integer.BYTES;
 
   @Override
   public byte[] toBytes(ZonedDateTime value) {
-    checkArgument(value.getZone() == ZoneOffset.UTC,
-            "Serialized ZonedDateTime value should be in UTC");
+    checkArgument(value.getZone() == ZoneOffset.UTC, "ZonedDateTime value should be in UTC, but was %s",
+        value.getZone().toString());
```
dmitry-timofeev: Will `value.getZone()` work (IIRC, checkX invoke `toString` for you and lazily)?

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/serialization/UtcZonedDateTimeSerializer.java
dmitry-timofeev: Or will it be useful in tests that want to (though this is not recommended, but nothing better is available at the moment) put something in TimeSchema?

exonum-java-binding-core/rust/src/utils/services.rs
skletsun: So I compared tests in `services` and `node_builder`and I'm not sure if we need anything to delete - while tests in the `services` module cover _the loading of data_, tests in the `node_builder` cover any additional logic around it.

exonum-java-binding-core/rust/src/utils/time_service.rs
skletsun: Yes, this is a heritage of previous checking from runtime )). Thanks for pointing this out - corrected.

exonum-java-binding-core/rust/src/utils/services.rs
skletsun: Corrected.

exonum-java-binding-time-service/src/test/java/com/exonum/binding/time/UtcZonedDateTimeSerializerTest.java
```diff
@@ -15,7 +15,11 @@
  *
  */
 
-package com.exonum.binding.time.serialization;
+package com.exonum.binding.time;
+
+import static org.hamcrest.CoreMatchers.equalTo;
```
dmitry-timofeev: You can install https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml to have IDEA take care of that.

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/serialization/UtcZonedDateTimeSerializer.java
dmitry-timofeev: @MakarovS Have you researched that?

exonum-java-binding-core/rust/src/utils/time_service.rs
```diff
@@ -0,0 +1,34 @@
...
+    static ref IS_TIME_SERVICE_ENABLED: jboolean =
+        is_service_enabled_in_config_file(TIME_SERVICE, PATH_TO_SERVICES_TO_ENABLE).into();
+}
+
+#[no_mangle]
+pub extern "system" fn Java_com_exonum_binding_time_TimeSchemaProxy_isTimeServiceEnabled(
+    _: JNIEnv,
+    _: JClass,
+) -> jboolean {
+    // TODO: For the moment we're just checking for the service's name presence in the configuration
```
dmitry-timofeev: Please file an issue in "Dynamic Services" to resolve that in the future.

dmitry-timofeev:
The Travis build does not seem to work.

exonum-java-binding-core/rust/src/utils/time_service.rs
skletsun: Done

exonum-java-binding-time-service/src/main/java/com/exonum/binding/time/serialization/UtcZonedDateTimeSerializer.java
MakarovS: I couldn't think of any cases when someone wants to put something in TimeSchema that we want to support right now. I'd leave it private and change it in the future in case such a need ever appears (and it's very possible that we'd move to protobuf earlier and would change it anyway).

exonum-java-binding-time-service/src/test/java/com/exonum/binding/time/UtcZonedDateTimeSerializerTest.java
MakarovS: Yes, thanks, I needed to reimport that.

exonum-java-binding/CHANGELOG.md
```diff
@@ -24,6 +24,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `--jvm-debug` command line argument that allows JDWP debugging of node. (#629)
 - Support of separated messages format Exonum v0.10. (#574)
 - `ListIndexProxy#stream` to enable stream processing of list elements. (#661)
+- Support of Time oracle.
+
+  To enable this service put `ejb_app_services.toml` file into EJB App's directory
```
bullet-tooth: are you sure that this part should be here (in changelog)?

dmitry-timofeev: probably no, we could reference documentation instead

exonum-java-binding/CHANGELOG.md
```diff
@@ -24,14 +24,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 - `ListIndexProxy#stream` to enable stream processing of list elements. (#661)
-- Support of Time oracle.
-
-  To enable this service put `ejb_app_services.toml` file into EJB App's directory
-  with the following content:
-  ```toml
-  services = ["configuration", "time"]
-  ```
-  (#667)
+- Support of Time oracle. Instruction on how to enable built-in services can be found
```
dmitry-timofeev: Please pull it up, to unreleased :-)

dmitry-timofeev: O, it's already there. Ok, will resolve during release->master merge

PR #697, opened 2019-02-11
Enable native integration tests on Linux [ECR-2789]

vitvakatu:
& use rocksdb and snappy apt packages on Travis CI

## Overview

---
See: https://jira.bf.local/browse/ECR-2789

### Definition of Done

- [x] Public API has Javadoc


dmitry-timofeev:
Hopefully, as there are no apparent changes that could resolve https://github.com/exonum/exonum-java-binding/commit/b38cbd2898dbe10a10910171d5c4a6869985fcc3

dmitry-timofeev:
Might be related to https://github.com/facebook/rocksdb/issues/4877, though there are a couple of differences/things to consider:
- We do not use RocksDB JNI interface
- The RocksDB on 16.04 is way older than in the issue (4.1-1 vs 5.2.0)
- Failing tests aren't the ITs (= don't even launch JVM)

dmitry-timofeev:
Travis does not seem to be working :disappointed: 

> Update to the newest available rocksdb version (5.17.2)

Don't we also have to remove the package installation from `.travis.yml` as it has different major version (4.2)?

vitvakatu:
I checked this locally, the update of this PR is kinda side effect 🙃

dmitry-timofeev:
> I checked this locally, the update of this PR is kinda side effect upside_down_face

Are you saying it no longer crashes on Linux where it used to? :open_mouth:

vitvakatu:
Oh, no, it still crashes :D

.travis.yml
```diff
@@ -16,9 +16,12 @@ addons:
     sources:
       - sourceline: 'ppa:chris-lea/libsodium'
       - sourceline: 'ppa:maarten-fonville/protobuf'
+      - sourceline: 'ppa:exonum/rocksdb'
     packages:
       - build-essential
       - libsodium-dev
+      - librocksdb5.17
```
dmitry-timofeev: Shall we specify a particular verison?

vitvakatu: This is the package name, so yes

dmitry-timofeev: Why not librocksdb-dev with no version?

https://launchpad.net/~exonum/+archive/ubuntu/rocksdb/+sourcepub/9938499/+listing-archive-extra

atrykush: The packages are named according the Debian/Ubuntu naming convention

vitvakatu: This is the package from our ppa, not ubuntu repos

dmitry-timofeev: @vitvakatu  If there is librocksdb-dev present in our PPA, why don't we use that?

atrykush: The Ubuntu package building tool (debuild) uses this convention when building packages

vitvakatu: 🤔 Because -dev contains headers, and we need only dynamic library to link with?

dmitry-timofeev: I see, that makes sense, thanks. But is it OK to fix the version (won't we forget to update it)?

vitvakatu: Not sure, but I think it's okay for now. Anyway, we'll need to test new version before actually using it. And not locking the version may accidentally break our code after update

dmitry-timofeev: :ok_man: , does the core have an issue to update the docs with Rocksdb installation instructions?

dmitry-timofeev: Would you please add a changelog entry that Rocksdb from our PPA must be installed on linux?

PR #700, opened 2019-02-12
Do not check proofs when sending them to Java [ECR-1802]

vitvakatu:
## Overview
No more checking of map proof before sending it to Java.

Also renamed `create_java_unchecked_flat_map_proof` to `create_java_unchecked_map_proof`

---
See: https://jira.bf.local/browse/ECR-1802

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/rust/src/storage/proof_map_index.rs
+++ b/exonum-java-binding/core/rust/src/storage/proof_map_index.rs
@@ -210,18 +210,8 @@ fn convert_to_java_proof<'a>(
 
     let native_entries: Vec<_> = proof.all_entries_unchecked().collect();
 
-    let mut missing_keys = Vec::with_capacity(native_entries.len());
-    let mut existing_entries = Vec::with_capacity(native_entries.len());
-
-    for entry in &native_entries {
-        match entry {
-            (key, Some(value)) => existing_entries.push((*key, *value)),
-            (key, None) => missing_keys.push(*key),
-        }
-    }
-
-    let missing_keys: JObject = create_java_missing_keys(&env, &missing_keys)?;
-    let map_entries: JObject = create_java_map_entries(&env, &existing_entries)?;
+    let map_entries: JObject = create_java_map_entries(&env, &native_entries)?;
+    let missing_keys: JObject = create_java_missing_keys(&env, &native_entries)?;
 
     create_java_unchecked_map_proof(&env, proof_nodes, map_entries, missing_keys)
 }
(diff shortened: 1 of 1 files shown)


exonum-java-binding/core/rust/src/storage/proof_map_index.rs
```diff
@@ -207,13 +207,23 @@ fn convert_to_java_proof<'a>(
...
+
+    for entry in &native_entries {
+        match entry {
+            (key, Some(value)) => existing_entries.push((*key, *value)),
+            (key, None) => missing_keys.push(*key),
+        }
+    }
+
+    let missing_keys: JObject = create_java_missing_keys(&env, &missing_keys)?;
+    let map_entries: JObject = create_java_map_entries(&env, &existing_entries)?;
```
dmitry-timofeev: Nit: reorder in the order they are subsequently used

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
```diff
@@ -207,13 +207,23 @@ fn convert_to_java_proof<'a>(
...
     let proof_nodes: JObject = create_java_proof_nodes(&env, &proof)?;
-    let missing_keys: JObject = create_java_missing_keys(&env, &proof)?;
 
-    // TODO: avoid checking proofs (ECR-1802) and reorder the surrounding operations
-    let checked_proof = proof.check().unwrap();
-    let map_entries: JObject = create_java_map_entries(&env, &checked_proof)?;
+    let native_entries: Vec<_> = proof.all_entries_unchecked().collect();
 
-    create_java_unchecked_flat_map_proof(&env, proof_nodes, map_entries, missing_keys)
+    let mut missing_keys = Vec::with_capacity(native_entries.len());
```
dmitry-timofeev: Please refactor so that this code is easier to grasp. Currently it operates on various level of abstractions (`create_java_proof_nodes` does everything in one go, then there is a loop that does something).

I'd suggest making `create_java_X` below accept a proof (or native_entries in case of performance considerations) so that this code looks
```
create_java_proof_nodes
create_java_missing_keys
create_java_map_entries
create_java_unchecked_map_proof
```

vitvakatu: 🤔 Passing `native_entries` to `create_java_missing_keys` and `create_java_map_entries` will lead to two iterations over `native_entries` instead of one. Not a huge concern, but I want to make this code reasonably effective. Do you think it's premature optimization?

dmitry-timofeev: It is good ­— the difference is dwarfed by JNI calls cost :)

PR #701, opened 2019-02-12
Bump project version for the next iteration

dmitry-timofeev:
## Overview

---

### Definition of Done

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 
   <groupId>com.exonum.binding</groupId>
   <artifactId>exonum-java-parent</artifactId>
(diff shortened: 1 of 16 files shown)


pom.xml
```diff
@@ -4,7 +4,7 @@
 
   <groupId>com.exonum.binding</groupId>
   <artifactId>exonum-java-parent</artifactId>
-  <version>0.4-SNAPSHOT</version>
+  <version>0.1</version>
```
dmitry-timofeev: Since it is a non-published pom (and mustn't be in sync with either ejb or lc), I think any will do.

PR #710, opened 2019-02-13
Reenabled Rust checks [ECR-2750]

skletsun:
## Re-enabled Rust checks on Travis

---
See: https://jira.bf.local/browse/ECR-2750

### Definition of Done

### DIFF
--- a/exonum-java-binding/core/rust/src/runtime/cmd.rs
+++ b/exonum-java-binding/core/rust/src/runtime/cmd.rs
@@ -273,10 +273,12 @@ where
     // Retrieve config by key
     let config = service_config
         .get(key)
-        .expect(&format!(
-            "Unable to read config with key [{}] from node configuration",
-            key
-        ))
+        .unwrap_or_else(|| {
+            panic!(
+                "Unable to read config with key [{}] from node configuration",
+                key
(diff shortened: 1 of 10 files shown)


exonum-java-binding/core/rust/src/runtime/cmd.rs
```diff
@@ -273,10 +273,12 @@ where
     // Retrieve config by key
     let config = service_config
         .get(key)
-        .expect(&format!(
-            "Unable to read config with key [{}] from node configuration",
-            key
-        ))
+        .unwrap_or_else(|| {
+            panic!(
```
dmitry-timofeev: Is it to format things lazily?

skletsun: Yes, this prevents [unconditional call of functions inside .expect()](https://rust-lang.github.io/rust-clippy/master/index.html#expect_fun_call). First, I ended up with `#[allow(clippy::expect_fun_call)]`, but then I decided that it's better to fix a couple of lines.

dmitry-timofeev: good, thanks!

PR #727, opened 2019-02-21
Switched to concurrent HashMap implementation in the resource-manager [ECR-2934]

skletsun:
## Overview
Switched from RwLock<HashMap> to chashmap::CHashMap for the handles storage in the resources manager.

---

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/rust/src/utils/resource_manager/imp.rs
+++ b/exonum-java-binding/core/rust/src/utils/resource_manager/imp.rs
@@ -12,14 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use chashmap::CHashMap;
 use std::any::TypeId;
-use std::collections::HashMap;
-use std::sync::RwLock;
 
 use utils::Handle;
 
+const MAP_CAPACITY: usize = 512;
+
 lazy_static! {
-    static ref HANDLES_MAP: RwLock<HashMap<Handle, HandleInfo>> = RwLock::new(HashMap::new());
+    static ref HANDLES_MAP: CHashMap<Handle, HandleInfo> = CHashMap::with_capacity(MAP_CAPACITY);
 }
 
 /// Represents `Handle` ownership model.
@@ -57,12 +58,9 @@ fn add_handle_impl<T: 'static>(handle: Handle, ownership: HandleOwnershipType) {
     assert_ne!(handle, 0);
     assert!(
         HANDLES_MAP
-            .write()
-            .expect("Unable to obtain write-lock")
             .insert(handle, HandleInfo::new(TypeId::of::<T>(), ownership))
             .is_none(),
         "Trying to add the same handle for the second time: {:X}",
-        handle
     )
 }
 
@@ -74,10 +72,7 @@ fn add_handle_impl<T: 'static>(handle: Handle, ownership: HandleOwnershipType) {
 fn remove_handle_impl<T: 'static>(handle: Handle, ownership: HandleOwnershipType) {
     check_handle_impl::<T>(handle, Some(ownership));
     // Return value is ignored because `check_handle_impl` already checks that handle is present.
-    HANDLES_MAP
-        .write()
-        .expect("Unable to obtain write-lock")
-        .remove(&handle);
+    HANDLES_MAP.remove(&handle);
 }
 
 /// Checks given handle for validity.
@@ -86,11 +81,7 @@ fn remove_handle_impl<T: 'static>(handle: Handle, ownership: HandleOwnershipType
 ///
 /// Panics if handle is unknown or its type or ownership model is wrong.
 fn check_handle_impl<T: 'static>(handle: Handle, ownership: Option<HandleOwnershipType>) {
-    match HANDLES_MAP
-        .read()
-        .expect("Unable to obtain read-lock")
-        .get(&handle)
-    {
+    match HANDLES_MAP.get(&handle) {
         Some(info) => {
             let actual_object_type = TypeId::of::<T>();
             assert_eq!(
@@ -159,10 +150,7 @@ pub fn check_handle<T: 'static>(handle: Handle) {
 
 /// Returns the number of known handles.
 pub fn known_handles() -> usize {
-    HANDLES_MAP
-        .read()
-        .expect("Unable to obtain read-lock")
-        .len()
+    HANDLES_MAP.len()
 }
 
 #[cfg(test)]
--- a/exonum-java-binding/core/rust/Cargo.lock
+++ b/exonum-java-binding/core/rust/Cargo.lock
@@ -335,6 +335,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "chashmap"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "owning_ref 0.2.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "parking_lot 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "chrono"
 version = "0.4.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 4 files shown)


exonum-java-binding/core/rust/src/utils/resource_manager/imp.rs
```diff
@@ -74,10 +72,7 @@ fn add_handle_impl<T: 'static>(handle: Handle, ownership: HandleOwnershipType) {
 fn remove_handle_impl<T: 'static>(handle: Handle, ownership: HandleOwnershipType) {
     check_handle_impl::<T>(handle, Some(ownership));
     // Return value is ignored because `check_handle_impl` already checks that handle is present.
-    HANDLES_MAP
-        .write()
-        .expect("Unable to obtain write-lock")
-        .remove(&handle);
+    HANDLES_MAP.remove(&handle);
```
vitvakatu: Interesting, CHashMap provides methods with internal mutability

exonum-java-binding/core/rust/src/utils/resource_manager/imp.rs
```diff
@@ -12,14 +12,15 @@
...
 // limitations under the License.
 
+use chashmap::CHashMap;
 use std::any::TypeId;
-use std::collections::HashMap;
-use std::sync::RwLock;
 
 use utils::Handle;
 
+const MAP_CAPACITY: usize = 512;
```
skletsun: I've chosen this value with the idea that reasonably big value will keep the load factor small enough spending not much memory to the initial pre-allocation.

dmitry-timofeev: Wait, how the map can determine the load factor from the capacity (for an empty map)? These are supposedly different things. Why don't we either pass a load factor (if you have a really good intuition of which is right) or don't pass anything?

exonum-java-binding/core/rust/src/utils/resource_manager/imp.rs
```diff
@@ -57,12 +58,9 @@ fn add_handle_impl<T: 'static>(handle: Handle, ownership: HandleOwnershipType) {
     assert_ne!(handle, 0);
     assert!(
         HANDLES_MAP
-            .write()
-            .expect("Unable to obtain write-lock")
             .insert(handle, HandleInfo::new(TypeId::of::<T>(), ownership))
             .is_none(),
         "Trying to add the same handle for the second time: {:X}",
```
dmitry-timofeev: Assuming `insert` returns the previous value, does there happen an *overwrite* of the previous value (i.e., the change of the state of the map in case of an invalid argument)?

exonum-java-binding/core/rust/src/utils/resource_manager/imp.rs
```diff
@@ -74,10 +72,7 @@ fn add_handle_impl<T: 'static>(handle: Handle, ownership: HandleOwnershipType) {
 fn remove_handle_impl<T: 'static>(handle: Handle, ownership: HandleOwnershipType) {
```
dmitry-timofeev: This method is not correctly synchronized, as it won't detect concurrent double-free:
```
  T1       T2
  c->OK
           c->OK
  r->removed
           r->nothing to remove (OK)
```

dmitry-timofeev:
Why don't we also push all these operations in a struct that can be instantiated in both lazy_static and unit tests? Otherwise it is quite hard to test operations on a static shared state.

exonum-java-binding/core/rust/src/utils/resource_manager/imp.rs
skletsun: Sorry for confusing you, I just meant that `CHashMap::with_capacity(N)` pre-allocates some amount of buckets depending on N, thus I supposed that defining some reasonable value here could prevent it from expensive re-allocations in future. In other words, it's rather a question if it is possible to estimate this or it is fully unpredictable and there is no sense in such optimizations.

exonum-java-binding/core/rust/src/utils/resource_manager/imp.rs
skletsun: Re-written using CHashMap.alter()

exonum-java-binding/core/rust/src/utils/resource_manager/imp.rs
```diff
@@ -160,111 +189,129 @@ mod tests {
...
-    fn manage_handles() {
-        // Add Java-owned handle.
-        enum T1 {}
-        add_handle::<T1>(MANAGE_HANDLES_FIRST_HANDLE);
-        check_handle::<T1>(MANAGE_HANDLES_FIRST_HANDLE);
-        check_handle_impl::<T1>(
-            MANAGE_HANDLES_FIRST_HANDLE,
-            Some(HandleOwnershipType::JavaOwned),
+    fn check_handle_type_and_ownership_ok() {
+        let handle_info = prepare_handle_info();
```
dmitry-timofeev: Here and elsewhere: I do not think `prepare_handle_info` works in such cases because subsequent assertions *do* rely on what it puts there. Either
* This method must be parameterized.
* Converted to a builder, relevant to the test data must be set up in each test that depends on that data
* If no parameterization needed — given a name that reflects what it creates.

Currently it is not clear what all these tests using `prepare_handle_info` test.

exonum-java-binding/core/rust/src/utils/resource_manager/imp.rs
```diff
@@ -160,111 +189,129 @@ mod tests {
...
+        HandlesStorage::check_handle_type_and_ownership::<HandleFooType>(
```
dmitry-timofeev: Do we need tests using this internal static method? Wouldn't it be simpler to 
1. Create a fresh HandleStorage.
2. Setup the test data using public API.
3. Perform the action we'd like to test.
4. Verify.
?

exonum-java-binding/core/rust/src/utils/resource_manager/imp.rs
```diff
@@ -160,111 +189,129 @@ mod tests {
...
-        // Add native-owned handle.
-        enum T3 {}
-        register_handle::<T3>(MANAGE_HANDLES_NON_OWNED_HANDLE);
-        check_handle::<T3>(MANAGE_HANDLES_NON_OWNED_HANDLE);
-        check_handle_impl::<T3>(
-            MANAGE_HANDLES_NON_OWNED_HANDLE,
-            Some(HandleOwnershipType::NativeOwned),
-        );
+    #[test]
+    fn manage_handles() {
```
dmitry-timofeev: Please split into more focused tests for operations of rm. There are already some good focused tests below — I'd consider identifying which cases are currently missing and extending the suite of such focused tests.

exonum-java-binding/core/rust/src/utils/resource_manager/imp.rs
```diff
@@ -160,111 +189,129 @@ mod tests {
...
+        storage.check_handle::<T>(INVALID_HANDLE, None);
     }
 
-    #[test]
-    #[should_panic(expected = "Wrong type id for")]
-    fn check_wrong_type_handle() {
-        add_handle::<T>(WRONG_TYPE_HANDLE);
-        enum OtherT {}
-        check_handle::<OtherT>(WRONG_TYPE_HANDLE);
+    fn prepare_storage() -> HandlesStorage {
```
dmitry-timofeev: create_storage?

exonum-java-binding/core/rust/src/utils/resource_manager/imp.rs
dmitry-timofeev: Please see also https://testing.googleblog.com/2018/02/testing-on-toilet-cleanly-create-test.html

exonum-java-binding/core/rust/src/utils/resource_manager/imp.rs
dmitry-timofeev: FYI, I bumped into another library, DashMap, it has proper [get_or_insert](https://docs.rs/dashmap/1.2.1/dashmap/struct.DashMap.html#method.get_or_insert) in its API.

Hovewer, even though it is 1.0+, there are no concurrent tests and even complete functional tests 👎

PR #731, opened 2019-02-22
Minor EJB fixes

MakarovS:
## Overview
Various minor changes throughout the project.

---
### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,6 +48,12 @@
       <name>Ilya Bogdanov</name>
       <url>https://github.com/vitvakatu</url>
     </developer>
+
+    <developer>
+      <id>skletsun</id>
+      <name>Sergey Kletsun</name>
+      <url>https://github.com/skletsun</url>
+    </developer>
   </developers>
 
   <modules>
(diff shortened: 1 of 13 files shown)


pom.xml
```diff
@@ -48,6 +48,12 @@
...
     </developer>
+
+    <developer>
+      <id>skletsun</id>
+      <name>Sergey Kletsun</name>
+      <url>https://github.com/skletsun</url>
+    </developer>
   </developers>
 
   <modules>
```
dmitry-timofeev: BTW, Why do we have developers/scm/issueManagement in the parent POM if it is not published anywhere (unlike ejb-parent & light-client)? Shan't we remove that as redundant?

vitvakatu: Additionally @DENMROOT is not in this list 🙃

dmitry-timofeev: From https://maven.apache.org/pom.html#Developers (from SO):

> A good rule of thumb is, if the person should not be contacted about the project, they need not be listed here.

As we don't include emails here, this whole section does not seem to make much sense to me.

PR #733, opened 2019-02-22
Proto files and serializers refactoring

bullet-tooth:
## Overview

---
Core proto files moved to common module
Serializers moved to common module
Created `BlockchainSerializers` class
`CoreProtos` renamed to `BlockchainProtos`

### Definition of Done

- [ ] Public API has Javadoc


exonum-java-binding/common/src/main/java/com/exonum/binding/common/blockchain/Block.java
```diff
@@ -71,14 +68,11 @@
 
   /**
    * Root hash of the Merkle tree of transactions in this block.
-   * These transactions can be accessed with {@link Blockchain#getBlockTransactions(Block)}.
```
dmitry-timofeev: I'd not just remove these — that's a regression — but think how the docs can be improved.

dmitry-timofeev:
I think there are some questions to be discussed before we go ahead and move things into common:

1. Blockchain proto-files and auto-generated classes:
  - a) Move all to common (but why do we expose core internal structures, on top of Block, in that API? Are they internal — or might they be required in some proofs? TransactionResult might actually be needed as it comes from a ProofMap)
  - b) Move only BlockProto. (But it's harder to manage, and TxResult might be needed ^ ).
  - c) Require anyone who needs to (de)serialize a block to generate classes using proto files — not the friendliest solution.
  
2. Serializers:
  - a) Move all to common (as in PR). There are two things that must be considered: (1) Block serializer is not just a serializer, it also works as a converter, calculating the hash of the block. That means that the roundtrip does not work **unless** you have a correct hash set. It is OK for an internal *read-only* serializer, but does not seem so for a public one. (2) Are other serializers useful to clients?
  - b) Move TxLocation & TxResult only, as they roundtrip properly. Add `Block#fromSerialized|binary(serializedBlock)` for those who need to create a block from a binary representation.
  - c) Don't move anything, but add `Block#fromSerialized|binary(serializedBlock)`.
  
  
3. Block:
  - a) Keep `getBlockHash()`
  - b) Discard `getBlockHash` — is it much needed?

bullet-tooth:
Created ECR-2957 and ECR-2958 as substitutions.

PR #736, opened 2019-02-25
[ECR-2959] LC: Update health check json response format

bullet-tooth:
## Overview

---
Updated health check response format
```
{
     "consensus_status": "Active",
     "connected_peers": 0
}
```
This change currently is unreleased i.e. is available at Exonum master only.

See: https://jira.bf.local/browse/ECR-2959

### Definition of Done

- [x] Public API has Javadoc


bullet-tooth:
Added do not merge label such as it's currently unreleased change.

dmitry-timofeev:
We have decided to not bother with releasing LC for 0.10, so I think we can merge this, is that correct?

bullet-tooth:
yes. After merge health check will work with 0.11 only.

PR #750, opened 2019-03-01
Temporary switching back to Rust 1.32.0

skletsun:
## Overview
Switching back to Rust 1.32.0 until bugs introduced by latest compiler version fixed.

---

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,8 @@ You need to install the following dependencies:
   * Linux or macOS. Windows support is coming soon. <!-- TODO: Link Java roadmap when it is published -->
   * [JDK 1.8+](http://jdk.java.net/10/).
   * [Maven 3.5+](https://maven.apache.org/download.cgi).
-  * [Stable Rust](https://www.rust-lang.org/).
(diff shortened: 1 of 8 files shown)


CONTRIBUTING.md
```diff
@@ -19,7 +19,8 @@ You need to install the following dependencies:
   * Linux or macOS. Windows support is coming soon. <!-- TODO: Link Java roadmap when it is published -->
   * [JDK 1.8+](http://jdk.java.net/10/).
   * [Maven 3.5+](https://maven.apache.org/download.cgi).
-  * [Stable Rust](https://www.rust-lang.org/).
+  * [Rust 1.32.0](https://www.rust-lang.org/).
+  To install a specific Rust version, use `rustup install 1.32.0` command.
   * The [system dependencies](https://exonum.com/doc/version/0.10/get-started/install/) of Exonum.
```
skletsun: Would it be better to use an abstract `VERSION` here?

dmitry-timofeev: I think it is OK.

We might even keep both instructions (one commented out) because we need to switch between the two rather frequently.

PR #766, opened 2019-03-06
Use dependency management for gson

bullet-tooth:
## Overview

---
Added dependency management to parent pom for gson.

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/pom.xml
+++ b/exonum-java-binding/pom.xml
@@ -166,6 +166,12 @@
       </dependency>
 
       <dependency>
+        <groupId>com.google.code.gson</groupId>
+        <artifactId>gson</artifactId>
+        <version>${gson.version}</version>
+      </dependency>
+
+      <dependency>
         <groupId>com.ryanharter.auto.value</groupId>
         <artifactId>auto-value-gson-annotations</artifactId>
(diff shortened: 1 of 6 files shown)


bullet-tooth:
@dmitry-timofeev @MakarovS 
Do we really need gson dependency at `service-archetype/src/main/resources/archetype-resources/pom.xml` ?

exonum-java-binding/pom.xml
```diff
@@ -165,6 +165,12 @@
         <version>${protobuf.version}</version>
       </dependency>
 
+      <dependency>
+        <groupId>com.google.code.gson</groupId>
+        <artifactId>gson</artifactId>
+        <version>${gson.version}</version>
```
dmitry-timofeev: What's funny is that the property is already there.

dmitry-timofeev:
> @dmitry-timofeev @MakarovS
Do we really need gson dependency at service-archetype/src/main/resources/archetype-resources/pom.xml ?

That's a question similar to "shall we add a protobuf configuration" :-D As we have added a JsonSerializer to ~~core~~ common, we might consider this dependency (now transitive) as not required :thinking:

PR #769, opened 2019-03-07
Archetype minor optimizations

bullet-tooth:
## Overview

---
* Optimize dependencies for archetype pom
* Fix `MyService` link in javadocs

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/service-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -85,8 +78,8 @@
 
     <dependency>
       <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
```
dmitry-timofeev: Why do you remove this one? Although mockito-junit-jupiter has it as a transitive, it is better to have both set as explicit dependencies.

bullet-tooth: Because it provides [mockito extensions](https://static.javadoc.io/org.mockito/mockito-junit-jupiter/2.23.4/org/mockito/junit/jupiter/MockitoExtension.html) and includes `mockito-core` as transitive dependency. 
but, if you want we can include it explicitly

bullet-tooth:
@MakarovS yes, checked. Idea marks read `$.MyService` and works well with `MyService`

PR #773, opened 2019-03-11
Make the ServiceRuntime load artifacts using ServiceLoader [ECR-3005]

dmitry-timofeev:
## Overview

- Create the ServiceRuntime with Guice 
- Make the ServiceRuntime load artifacts using ServiceLoader
- Configure projects-services to package the classes in a fat JAR with
meta data describing them. The service dependencies that are
currently the framework dependencies are marked as 'provided', however,
such decision is to be discussed (Are we certain we want to make
these dependencies guaranteed to be available with the framework?).
ServiceModules are also renamed for clarity.
---
See: https://jira.bf.local/browse/ECR-3005

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/core/src/main/java/com/exonum/binding/runtime/ServiceRuntimeBootstrap.java
```diff
@@ -20,12 +20,16 @@
...
 import com.google.inject.Injector;
 import com.google.inject.Module;
+import com.google.inject.Stage;
 
 /**
  * A bootstrap loader of the service runtime.
  */
 final class ServiceRuntimeBootstrap {
 
+  // TODO: Shall it be configurable (Stage.DEV in tests, Stage.PRODUCTION in prod)?
```
MakarovS: Where other than in `ServiceRuntimeTest` would we want to use `Stage.DEVELOPMENT`?

exonum-java-binding/core/rust/ejb-app/start_cryptocurrency_node.sh
```diff
@@ -38,6 +38,7 @@ echo "PROJ_ROOT=${EJB_ROOT}"
 header "PREPARE PATHS"
 
 CORE_TXT="core/target/ejb-core-classpath.txt"
+# FIXME: Remove the CRYPTOCURRENCY_TXT as it is no longer generated/needed, see PR for ECR-3005
```
MakarovS: When should it be removed?

exonum-java-binding/core/src/main/java/com/exonum/binding/runtime/ServiceRuntimeBootstrap.java
dmitry-timofeev: In native ITs, I think.

dmitry-timofeev: There aren't many of them either.

exonum-java-binding/core/rust/ejb-app/start_cryptocurrency_node.sh
dmitry-timofeev: When configuration is stabilized in the target feature branch. @skletsun knows :-)

exonum-java-binding/core/src/main/java/com/exonum/binding/runtime/ServiceRuntime.java
```diff
@@ -33,28 +39,33 @@
...
 
   private final Injector frameworkInjector;
+  private final ServiceLoader serviceLoader;
 
   /**
    * Creates a new runtime with the given framework injector. Starts the server on instantiation.
    *
    * @param frameworkInjector the injector that has been configured with the Exonum framework
    *     bindings. It serves as a parent for service injectors
+   * @param serviceLoader the loader of service artifacts
```
bullet-tooth:
```suggestion
   * @param serviceLoader a loader of service artifacts
```

exonum-java-binding/core/src/test/java/com/exonum/binding/runtime/ServiceRuntimeTest.java
```diff
@@ -0,0 +1,159 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+/**
+ * An integration test of ServiceRuntime + Guice interop.
+ */
+@ExtendWith(MockitoExtension.class)
+class ServiceRuntimeTest {
+
+  static final int PORT = 25000;
```
bullet-tooth: it can be a situation when the port is taken (hopefully really rare). 
How about to make `com.exonum.binding.transport.Server#start` returns actual port, or extend the interface with `getActualPort` method? Currently, if we pass `0` as a port to the `Server` it will run on a random port and there is no way to get it.

exonum-java-binding/core/src/test/java/com/exonum/binding/runtime/ServiceRuntimeTest.java
```diff
@@ -0,0 +1,159 @@
...
+    @BeforeEach
+    void setUp() {
+      rootInjector = Guice.createInjector(Stage.DEVELOPMENT, new TestFrameworkModule());
+      serviceRuntime = rootInjector.getInstance(ServiceRuntime.class);
+    }
+
+    @Test
+    void startsServerOnInstantiation() {
+      Server server = rootInjector.getInstance(Server.class);
+      verify(server).start(PORT);
```
bullet-tooth: when does it stop?

bullet-tooth:
I like the way defining dependencies with `provided` scope. But we need to clarify if there any cases when no dependencies available with the framework?  I don't know such... 🤔

dmitry-timofeev:
> But we need to clarify if there any cases when no dependencies available with the framework?

I think with packaging (cc @vitvakatu ) they will be included with the application and available on the system classpath. Even with _no_ packaging, i.e., multiple services as is, the users will have to provide the system classpath that _must_ include exonum-java-binding-core and its transitive dependencies (it is possible, in this case, to miss some of them and not get a ClassNotFoundException _immediately_, e.g., if you don't load some class till the service operation, but it is an unlikely scenario).

I'd reformulate my question this way: "Are we comfortable saying that we will not remove these dependencies in a compatible version?" Actually, a future compatible, as with 0.x any is incompatible :smirk_cat: Probably, after dynamic services we will be able to tell definitely.

exonum-java-binding/core/src/test/java/com/exonum/binding/runtime/ServiceRuntimeTest.java
dmitry-timofeev: Currently, never. If we decide to support service **runtime** unloading (i.e., stopping the whole runtime) for some reason, we might need to do that.

exonum-java-binding/core/src/test/java/com/exonum/binding/runtime/ServiceRuntimeTest.java
dmitry-timofeev: I see, that is a sensible thing to do for integration tests that use `VertxServer`. In production, however, the server must either start on the configured port or fail. I think we can discuss it in a separate issue: https://jira.bf.local/browse/ECR-3016 

These tests use a mock of `Server`, so in this case it is totally reliable.

PR #785, opened 2019-03-15
Add factories of various service artifacts for native ITs [ECR-3031]

dmitry-timofeev:
## Overview

Add factories of various service artifacts for native ITs.

---
See: https://jira.bf.local/browse/ECR-3031

### Definition of Done

- [x] There are no TODOs left in the code
- [x] Change is covered by automated [tests](https://github.com/exonum/exonum-java-binding/blob/master/CONTRIBUTING.md#tests)
- [x] The [coding guidelines](https://github.com/exonum/exonum-java-binding/blob/master/CONTRIBUTING.md#the-code-style) are followed
- [x] Public API has Javadoc
- [x] Method preconditions are checked and documented in the Javadoc of the method
- [x] Changelog is updated if needed (in case of notable or breaking changes)
- [ ] The continuous integration build passes

PR #776, opened 2019-03-12
EJB App Packaging for Mac & Linux

vitvakatu:
## Overview

---
See: https://jira.bf.local/browse/ECR-XYZ

### Definition of Done

- [ ] There are no TODOs left in the code
- [ ] Change is covered by automated [tests](https://github.com/exonum/exonum-java-binding/blob/master/CONTRIBUTING.md#tests)
- [ ] The [coding guidelines](https://github.com/exonum/exonum-java-binding/blob/master/CONTRIBUTING.md#the-code-style) are followed
- [ ] Public API has Javadoc
- [ ] Method preconditions are checked and documented in the Javadoc of the method
- [ ] Changelog is updated if needed (in case of notable or breaking changes)
- [ ] The continuous integration build passes

PR #787, opened 2019-03-18
Support for several services in native [ECR-2999]

skletsun:
## Overview
Implemented support for several services on the native side. Services are loaded from artifacts via service loader.

---
See: https://jira.bf.local/browse/ECR-2999

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/core/rust/ejb-app/start_cryptocurrency_node.sh
```diff
@@ -35,15 +35,20 @@ echo "CURRENT_DIR=${EJB_APP_DIR}"
 EJB_ROOT=$(realpath "../../..")
 echo "PROJ_ROOT=${EJB_ROOT}"
 
+# Find the artifact
+ARTIFACT_PATH="$(find ${EJB_ROOT} -type f -name exonum-java-binding-cryptocurrency-demo-*-artifact.jar)"
+echo "ARTIFACT_PATH=${ARTIFACT_PATH}"
+
+# Prepare the services configuration file
+SERVICES_CONFIG_FILE="ejb_app_services.toml"
+SERVICE_NAME="cryptocurrency-demo-service"
```
skletsun: Those names (as well as for `qa_service`) are required to be passed to Exonum via the `ServiceFactory` interface (and are used to get a list of services that are executed by node). I set it to the `JavaService.getName()` for consistency.

exonum-java-binding/core/rust/ejb-app/TUTORIAL.md
```diff
@@ -42,14 +42,29 @@ export LD_LIBRARY_PATH="${LIBJVM_PATH}:${RUST_LIB_PATH}:${EJB_LIBPATH}"
...
+
+The sample of `ejb_app_services.toml` file that enables all possible built-in Exonum services 
+and two user services:
+```toml
+enabled_services = ["configuration", "btc-anchoring", "time"]
+
+[user_services]
+service_name1 = '/path/to/service1_artifact.jar'
+service_name2 = '/path/to/service2_artifact.jar'
+```
```
skletsun: Are names like `enabled_services` and `user_services` OK? `enabled_services` -> `builtin_services`?

exonum-java-binding/core/rust/integration_tests/tests/service_runtime.rs
```diff
@@ -19,20 +19,18 @@ extern crate integration_tests;
...
 use exonum_testkit::TestKitBuilder;
-use integration_tests::vm::{get_fakes_classpath, get_libpath};
-use java_bindings::{Config, JavaServiceRuntime, JvmConfig, RuntimeConfig, ServiceConfig};
+use integration_tests::vm::{get_fake_service_artifact_path, get_fakes_classpath, get_libpath};
+use java_bindings::{Config, JavaServiceRuntime, JvmConfig, RuntimeConfig};
 
 #[test]
-#[ignore]
-// TODO: Reenable after ECR-2999/ECR-3011
+// TODO: reenable this test after ECR-2789
```
dmitry-timofeev: Why is this commented and not removed?

exonum-java-binding/core/rust/src/runtime/java_service_runtime.rs
```diff
@@ -49,27 +42,14 @@ pub struct JavaServiceRuntime {
 }
 
 impl JavaServiceRuntime {
-    /// Creates new runtime from provided config or returns the one created earlier.
-    ///
-    /// There can be only one `JavaServiceRuntime` instance at a time.
```
dmitry-timofeev: Why is this statement removed? Multiple JVMs are not supported in the same process.

exonum-java-binding/core/rust/src/runtime/java_service_runtime.rs
```diff
@@ -49,27 +42,14 @@ pub struct JavaServiceRuntime {
...
-            });
-            // Return global runtime.
-            JAVA_SERVICE_RUNTIME
-                .clone()
-                .expect("Trying to return runtime, but it's uninitialized")
+    /// Creates new runtime from provided config.
+    pub fn create_java_runtime(config: Config) -> Self {
+        let java_vm = Self::create_java_vm(&config.jvm_config, &config.runtime_config);
+        let executor = MainExecutor::new(Arc::new(java_vm));
+        let service_runtime = Self::create_service_runtime(config.runtime_config, executor.clone());
```
dmitry-timofeev: The difference between
`create_java_runtime`
and
`create_service_runtime` 
is not clear. I'd think of better names (swap them?).

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
```diff
@@ -0,0 +1,103 @@
...
+impl ServiceFactory for JavaServiceFactoryAdapter {
+    fn service_name(&self) -> &str {
+        &self.name
+    }
+
+    fn command(&mut self, command_name: &str) -> Option<Box<CommandExtension>> {
+        use exonum::helpers::fabric;
+        // Execute EJB configuration steps along with standard Exonum Core steps.
+        let mut command_ext: Option<Box<CommandExtension>> = None;
+        if command_name == fabric::Run.name() {
```
dmitry-timofeev: I don't understand what this conditional means — please clarify.

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
```diff
@@ -0,0 +1,103 @@
...
+                JAVA_SERVICE_RUNTIME = Some(runtime);
+            }
+        });
+
+        unsafe { JAVA_SERVICE_RUNTIME.clone() }
+            .expect("Trying to return runtime, but it's uninitialized")
+    }
+
+    // Extracts EJB and JVM configuration from Context
+    fn extract_config(context: &Context) -> Config {
```
dmitry-timofeev: Why don't we push this operation to the ServiceFactory implementation and pass a Config to the method above so that these operations on the Context are closer to where they make sense (ServiceFactory)?

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
```diff
@@ -0,0 +1,103 @@
...
+use std::sync::{Once, ONCE_INIT};
+
+static mut JAVA_SERVICE_RUNTIME: Option<JavaServiceRuntime> = None;
+static JAVA_SERVICE_RUNTIME_INIT: Once = ONCE_INIT;
+static JAVA_SERVICE_CREATE_COMMAND: Once = ONCE_INIT;
+
+/// Adapts current single-service interface of Exonum to multiple EJB services until dynamic
+/// services are implemented. Tracks the `JavaServiceRuntime` instantiation and command extension
+/// and represents factory for every user service relying on current `ServiceFactory` interface.
+pub struct JavaServiceFactoryAdapter {
```
dmitry-timofeev: Are there any tests for all these operations?

exonum-java-binding/core/rust/src/utils/services.rs
```diff
@@ -14,36 +14,46 @@
...
 pub const CONFIGURATION_SERVICE: &str = "configuration";
 pub const BTC_ANCHORING_SERVICE: &str = "btc-anchoring";
 pub const TIME_SERVICE: &str = "time";
-pub const EJB_SERVICE: &str = "ejb-service";
 
 #[derive(Serialize, Deserialize)]
-struct ServicesToEnable {
-    services: HashSet<String>,
+pub struct EjbAppServices {
+    pub enabled_services: Option<HashSet<String>>,
```
dmitry-timofeev: system_services — aren't user_services enabled too?

exonum-java-binding/core/rust/src/utils/services.rs
```diff
@@ -14,36 +14,46 @@
...
+
+// Produces the EjbAppServices structure from string representation in TOML format.
+fn parse_services(data: String) -> Result<EjbAppServices> {
+    let services: EjbAppServices =
+        toml::from_str(&data).expect("Invalid format of the file with EJB services definition");
     Ok(services)
 }
 
 /// Determines whether particular service name is defined in the specific TOML configuration file.
 pub fn is_service_enabled_in_config_file<P: AsRef<Path>>(service_name: &str, path: P) -> bool {
```
dmitry-timofeev: The documentation of this method must be updated to reflect which services it supports (seemingly not user services).

exonum-java-binding/core/rust/src/runtime/java_service_runtime.rs
```diff
@@ -111,16 +91,16 @@ impl JavaServiceRuntime {
     /// Loads an artifact from the specified location involving verification of the artifact.
```
dmitry-timofeev: >  pub fn create_service(&self, artifact_id: &str) -> ServiceProxy {

Why doesn't it return a Result? This operation can easily fail in Java.

exonum-java-binding/core/rust/integration_tests/src/vm.rs
```diff
@@ -27,6 +27,8 @@ pub const KIB: usize = 1024;
 /// Mebibyte
 pub const MIB: usize = KIB * KIB;
 
+const CONV_FAILED: &str = "Failed to convert FS path into utf-8";
```
dmitry-timofeev: I think a file-wide constant deserves a clearer name (CONVERSION_FAILED_MESSAGE?)

exonum-java-binding/core/rust/ejb-app/start_qa_node.sh
```diff
@@ -34,15 +34,20 @@ echo "CURRENT_DIR=${EJB_APP_DIR}"
...
 
+# Find the artifact
+ARTIFACT_PATH="$(find ${EJB_ROOT} -type f -name exonum-java-binding-qa-service-*-artifact.jar)"
+echo "ARTIFACT_PATH=${ARTIFACT_PATH}"
+
+# Prepare the services configuration file
+SERVICES_CONFIG_FILE="ejb_app_services.toml"
+SERVICE_NAME="ejb-qa-service"
+echo "[user_services]" > ${SERVICES_CONFIG_FILE}
+echo "${SERVICE_NAME} = '${ARTIFACT_PATH}'" >> ${SERVICES_CONFIG_FILE}
```
vitvakatu: I think adding special parameter to the Run configuration would be better than a file in working directory, won't it?

exonum-java-binding/core/rust/src/runtime/cmd.rs
```diff
@@ -42,36 +40,6 @@ const JVM_ARGS_APPEND: &str = "JVM_ARGS_APPEND";
 // TOML keys for EJB configuration
 pub const EJB_CONFIG_SECTION_NAME: &str = "ejb";
 
-/// Encapsulates processing of extensions of the `finalize` command. At this step we gather some of
```
vitvakatu: Docs at the start of the file also need changes

exonum-java-binding/core/rust/src/runtime/java_service_runtime.rs
vitvakatu: I think this method actually needs to be called `new` (current JavaServiceRuntime::create_java_runtime)

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
```diff
@@ -0,0 +1,103 @@
+// Copyright 2019 The Exonum Team
```
vitvakatu: The name of this file is actually conflicting with the name from #776. (it contains the same kind of refactoring)

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
```diff
@@ -0,0 +1,103 @@
...
+/// services are implemented. Tracks the `JavaServiceRuntime` instantiation and command extension
+/// and represents factory for every user service relying on current `ServiceFactory` interface.
+pub struct JavaServiceFactoryAdapter {
+    name: String,
+    artifact_path: String,
+}
+
+impl JavaServiceFactoryAdapter {
+    /// Creates new instance with given service name and path to the artifact.
+    pub fn new<S: AsRef<str>>(name: S, artifact_path: S) -> Self {
```
vitvakatu: Keep in mind, that such signature requires both parameters to have the exact same type.

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
```diff
@@ -0,0 +1,103 @@
...
+        JavaServiceFactoryAdapter {
+            name: name.as_ref().to_owned(),
+            artifact_path: artifact_path.as_ref().to_owned(),
```
vitvakatu: And actually, what is bad in accepting just Strings?

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
```diff
@@ -0,0 +1,103 @@
...
+    fn service_name(&self) -> &str {
+        &self.name
+    }
+
+    fn command(&mut self, command_name: &str) -> Option<Box<CommandExtension>> {
+        use exonum::helpers::fabric;
+        // Execute EJB configuration steps along with standard Exonum Core steps.
+        let mut command_ext: Option<Box<CommandExtension>> = None;
+        if command_name == fabric::Run.name() {
+            // We return command extension only once for all instances of ServiceFactory otherwise
```
vitvakatu: Interesting restriction

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
```diff
@@ -0,0 +1,103 @@
...
+        command_ext
+    }
+
+    fn make_service(&mut self, context: &Context) -> Box<Service> {
+        let runtime = Self::get_or_create_java_service_runtime(context);
+
+        // load service from artifact and create corresponding proxy
+        let artifact_id = runtime
+            .load_artifact(&self.artifact_path)
+            .expect("Unable to load artifact");
```
vitvakatu: Let's add info about what artifact failed

exonum-java-binding/core/rust/src/utils/services.rs
```diff
@@ -53,45 +63,74 @@ mod tests {
     use tempfile::{Builder, TempPath};
 
     #[test]
-    fn no_config() {
-        let res = load_enabled_services("nonexistent");
-        assert!(res.is_err());
+    fn missed_enabled_services_section() {
+        let cfg = "[user_services]\nservice_name1 = '/path/to/artifact1'\nservice_name2 = '/path/to/artifact2'".to_owned();
```
vitvakatu: Maybe use constants with raw string literal syntax? Would be much more readable.

exonum-java-binding/core/rust/ejb-app/TUTORIAL.md
```diff
@@ -42,14 +42,29 @@ export LD_LIBRARY_PATH="${LIBJVM_PATH}:${RUST_LIB_PATH}:${EJB_LIBPATH}"
...
-the [Maven Shade Plugin](https://maven.apache.org/plugins/maven-shade-plugin/index.html)
-and pass a path to the service artefact during application configuration as `--ejb-classpath`
-parameter. Alternatively, you may assemble a classpath that includes the service and all of 
-its dependencies and pass it instead.
+Classpath is used to locate Java classes of the Exonum Java Binding and its dependencies.
+ Note, that classpath has nothing to do with classes of user services. 
+
+#### Services definition
+Services are loaded as `jar` artifacts and should be described in the 
+[ejb_app_services.toml](https://exonum.com/doc/version/0.10/get-started/java-binding/)
```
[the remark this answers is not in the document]
bullet-tooth: I would rewrite the sentence. Something like:
"It 's required for services to be configured for being available in the network. Services are configured in the [`ejb_app_services.toml`][<link>] file."

And move "Services are loaded as `jar`..." section to the user services description

exonum-java-binding/core/rust/ejb-app/TUTORIAL.md
```diff
@@ -42,14 +42,29 @@ export LD_LIBRARY_PATH="${LIBJVM_PATH}:${RUST_LIB_PATH}:${EJB_LIBPATH}"
...
+is one-word description of the service and `artifact` is full path to the service's artifact. 
+At least one service must be defined.
+
+The sample of `ejb_app_services.toml` file that enables all possible built-in Exonum services 
+and two user services:
+```toml
+enabled_services = ["configuration", "btc-anchoring", "time"]
+
+[user_services]
+service_name1 = '/path/to/service1_artifact.jar'
```
bullet-tooth: Why does it have `'`, but enabled_services has `"`?

exonum-java-binding/core/rust/ejb-app/TUTORIAL.md
```diff
@@ -97,11 +109,12 @@ $ cargo run -- run -d testnet/db -c testnet/node.toml --public-api-address 127.0
     --ejb-port 6000
 ```
 
-There are few specific optional parameters here:
+There are few optional parameters here:
 - `--jvm-args-prepend` and `--jvm-args-append`: Additional parameters for JVM that prepend and
  append the rest of arguments. Must not have a leading dash. For example, `Xmx2G`.
 - `--jvm-debug`: Allows JVM being remotely debugged over the `JDWP` protocol. Takes a socket address as a parameter in form
  of `HOSTNAME:PORT`. For example, `localhost:8000`.
+- `--ejb-log-config-path`: Path to log4j configuration file.
```
bullet-tooth: What should be used as a default behavior? Disabled logging or console output?

exonum-java-binding/core/rust/ejb-app/start_qa_node.sh
dmitry-timofeev: Which parameter? If the list of services, then, IIRC, it is needed _before_ arguments are processed by the framework.

vitvakatu: Ah, indeed, always forget about that fact

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
dmitry-timofeev: Will it be the first message the user sees? I also wonder if the error message from the Java exception will be included in the output.

We could test that with an "unloadable" artifact from #785 

Also https://jira.bf.local/browse/ECR-3043

exonum-java-binding/core/rust/ejb-app/start_qa_node.sh
dmitry-timofeev: This file is a temporary configuration feature that will get replaced by (?) some transactions.

exonum-java-binding/core/rust/ejb-app/TUTORIAL.md
skletsun: Well, the `toml` crate serializes it that way by default, so I applied it for the new pieces of config file just automatically. And yes, it's better to choose one quotes format for consistency. Thanks, going to change it.

exonum-java-binding/core/rust/ejb-app/TUTORIAL.md
skletsun: Console output AFAIR

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
skletsun: Well, the main idea was to extract Config from Context only once and when it's only required - at the service runtime creation step. In other words - why any other service that is being created after the creation of the service runtime should depend on possible errors of extracting Config from Context (even if they aren't logically possible within current implementation) if Context is not needed anymore? And WRT to the assumption that this code is temporary until dynamic services in core I considered this acceptable.

dmitry-timofeev: Yes, it is acceptable and may remain as is. 

Is this context mutable for that hypothetical failure to be a concern?

I see that this change will involve extracting the config even when it is not needed — but this is neither a frequent nor expensive operation to be a concern. The temporary part here is likely ServiceFactory and Context, not the runtime. So it also makes sense to move this operation operating on the context to the factory, and pass to JSR method a thing that it understands — its configs.

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
```diff
@@ -41,30 +40,18 @@ fn service_factories() -> HashMap<String, Box<ServiceFactory>> {
...
-    });
-
-    // Add EJB_SERVICE if it's missing
-    services.insert(EJB_SERVICE.to_owned());
+pub fn create() -> fabric::NodeBuilder {
+    let EjbAppServices {
+        enabled_services,
+        user_services,
+    } = load_services_definition(PATH_TO_SERVICES_DEFINITION)
+        .expect("Unable to load services definition");
```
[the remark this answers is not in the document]
skletsun: Hmm... It can fail not only because of missing file. I'd probably left it as is since `.expect()` includes original cause and in this case we'll get smth like `Unable to load services definition` (general error) and then detailed description of error: missing file, non-UTF symbols, wrong format whatever. WDYT?

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
skletsun: Yes, that sounds reasonable.

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
skletsun: Yes, `.expect()` includes passed error message in case of error.

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
```diff
@@ -76,81 +63,81 @@ pub fn create() -> fabric::NodeBuilder {
...
+            name,
+            artifact_path,
+        )));
+    }
+
     builder
 }
 
+// Extracts defined services or inserts the configuration service otherwise.
+fn validate_services(services: Option<HashSet<String>>) -> HashSet<String> {
```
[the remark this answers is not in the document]
skletsun: Well, yes... On one hand, it validates the content of the HashSet - it inserts a value if it's empty. But on the other hand, it has nothing to do with validation of services as it comes from the name. Probably `extract_services` fits better.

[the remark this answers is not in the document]
skletsun: @vitvakatu not sure if I get what you mean, but the intention of introducing this method was to simplify testing a little WRT to the idea this method will most probably be inlined into resulting code.

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
```diff
@@ -76,81 +63,81 @@ pub fn create() -> fabric::NodeBuilder {
...
+        assert!(services.contains(TIME_SERVICE));
     }
 
     #[test]
     fn all_services() {
         let cfg = create_config(
             "all.toml",
-            "services = [\"configuration\", \"btc-anchoring\", \"time\"]",
+            "enabled_services = [\"configuration\", \"btc-anchoring\", \"time\"]\n\
+            [user_services]\nservice_name1 = '/path/to/artifact1'\nservice_name2 = '/path/to/artifact2'",
```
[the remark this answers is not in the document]
skletsun: Replaced with raw Rust literals (thanks @vitvakatu for the idea)

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
```diff
@@ -76,81 +63,81 @@ pub fn create() -> fabric::NodeBuilder {
...
+        assert_eq!(enabled_services.len(), 3);
+        assert!(enabled_services.contains(CONFIGURATION_SERVICE));
+        assert!(enabled_services.contains(BTC_ANCHORING_SERVICE));
+        assert!(enabled_services.contains(TIME_SERVICE));
 
         let service_factories = service_factories();
-        for service in &services_to_enable {
-            assert!(service_factories.get(service).is_some())
+        for service in enabled_services {
+            assert!(service_factories.get(&service).is_some())
```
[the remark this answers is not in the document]
skletsun: Not sure about the original meaning of this part because I just adapted the old test in node_builder but I agree that it looks redundant within this test.

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
skletsun: Yeap, better to keep it simple...

exonum-java-binding/core/rust/src/utils/services.rs
skletsun: Thanks, somehow I totally forgot about them

exonum-java-binding/core/rust/src/runtime/java_service_runtime.rs
skletsun: Perhaps I got it a little wrong - I thought it was related to the old code of singleton creation, thus moved it to the JavaServiceFactoryAdapter.get_or_create_java_service_runtime(). Copied it back.

exonum-java-binding/core/rust/integration_tests/tests/service_runtime.rs
skletsun: Fixed

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
```diff
@@ -76,81 +63,81 @@ pub fn create() -> fabric::NodeBuilder {
...
-    }
-
-    #[test]
-    fn with_anchoring() {
-        let cfg = create_config("anchoring.toml", "services = [\"btc-anchoring\"]");
-        let services_to_enable = services_to_enable(cfg);
-        assert_eq!(services_to_enable.len(), 2);
-        assert!(services_to_enable.contains(EJB_SERVICE));
-        assert!(services_to_enable.contains(BTC_ANCHORING_SERVICE));
+    fn validate_services_ok() {
```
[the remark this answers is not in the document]
skletsun: Done.

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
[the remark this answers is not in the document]
skletsun: The majority of `node_builder` code is rewritten in order to make it more testable.

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
```diff
@@ -76,81 +63,81 @@ pub fn create() -> fabric::NodeBuilder {
...
+        services.insert(BTC_ANCHORING_SERVICE.to_owned());
+        services.insert(TIME_SERVICE.to_owned());
+        let services = validate_services(Some(services));
+        assert_eq!(services.len(), 2);
+        assert!(services.contains(BTC_ANCHORING_SERVICE));
+        assert!(services.contains(TIME_SERVICE));
     }
 
     #[test]
     fn all_services() {
```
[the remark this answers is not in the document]
skletsun: I re-wrote tests in a more convenient way.

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
```diff
@@ -41,30 +40,18 @@ fn service_factories() -> HashMap<String, Box<ServiceFactory>> {
...
-    // Return default list if config file not found.
-    let mut services = load_enabled_services(path).unwrap_or_else(|_| {
-        let mut services = HashSet::new();
-        services.insert(CONFIGURATION_SERVICE.to_owned());
-        services
-    });
-
-    // Add EJB_SERVICE if it's missing
-    services.insert(EJB_SERVICE.to_owned());
+pub fn create() -> fabric::NodeBuilder {
```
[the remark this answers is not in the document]
skletsun: Re-written in order to cover as much as possible code.

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
skletsun: Corrected.

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
skletsun: Tried to cover as much as possible, but if it would be more or less trivial to "mock" the `NodeConfig` somehow (or simulate command line argument processing) it would be possible to cover more things.

exonum-java-binding/core/rust/src/utils/services.rs
skletsun: Done.

exonum-java-binding/core/rust/src/runtime/java_service_runtime.rs
skletsun: It's not trivial to deal with Result here because of different types of errors that could happen. Thus for the moment it just panics for error cases, which looks pretty natural for the current level of support of multiple services, but probably not the best solution for dynamic services.

exonum-java-binding/core/rust/integration_tests/src/vm.rs
skletsun: Done.

exonum-java-binding/core/rust/src/runtime/cmd.rs
skletsun: Done.

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
skletsun: Is it a problem? Perhaps it is, especially for one who's merging last ))

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
skletsun: Yeap, re-written to strings

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
```diff
@@ -17,16 +17,80 @@
...
+    let service_factories = prepare_service_factories(PATH_TO_SERVICES_DEFINITION);
+    let mut builder = fabric::NodeBuilder::new();
+    for service_factory in service_factories {
+        builder = builder.with_service(service_factory);
+    }
+    builder
+}
+
+// Prepares vector of `ServiceFactory` from services configuration file located at given path.
+// Panics in case of problems with reading/validating of service configuration file.
```
dmitry-timofeev: Will there be a *proper* user-friendly error message in case this code panics (**cc** @dip56 )? 

Shall panics be used in such a case at all?

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
```diff
@@ -17,16 +17,80 @@
...
+        services
+    });
+
+    // Prepare list of service factories from system and user services
+    let mut resulting_factories = Vec::new();
+
+    // Process system services first
+    let mut all_system_service_factories = service_factories();
+    for service_name in system_services {
+        match all_system_service_factories.remove(&service_name) {
```
dmitry-timofeev: (Again) I find this code unreadable and unnatural, because it removes elements for no apparent reason when it seems to just need to check if a factory is present and get it.

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
```diff
@@ -41,116 +105,101 @@ fn service_factories() -> HashMap<String, Box<ServiceFactory>> {
...
-        cfg_file.into_temp_path()
-    }
-
     #[test]
-    fn no_config() {
-        let services_to_enable = services_to_enable("");
-        assert_eq!(services_to_enable.len(), 2);
-        assert!(services_to_enable.contains(EJB_SERVICE));
-        assert!(services_to_enable.contains(CONFIGURATION_SERVICE));
+    #[should_panic(expected = "Unable to load services definition")]
```
dmitry-timofeev: If that's it, such user-facing error message is not good enough.

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
```diff
@@ -41,116 +105,101 @@ fn service_factories() -> HashMap<String, Box<ServiceFactory>> {
...
 
     #[test]
-    fn empty_list() {
-        let cfg = create_config("empty_list.toml", "services = []");
-        let services_to_enable = services_to_enable(cfg);
-        assert_eq!(services_to_enable.len(), 1);
-        assert!(services_to_enable.contains(EJB_SERVICE));
+    #[should_panic(expected = "At least one user service should be defined")]
+    fn missing_user_services() {
+        let cfg = create_config(
```
dmitry-timofeev: As it is *not* a config, but a path to it, please name accordingly.

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
```diff
@@ -41,116 +105,101 @@ fn service_factories() -> HashMap<String, Box<ServiceFactory>> {
...
     }
 
     #[test]
-    fn duplicated() {
+    #[should_panic(expected = "Unknown system service name")]
+    fn unknown_system_service() {
         let cfg = create_config(
-            "duplicated.toml",
-            "services = [\"btc-anchoring\", \"btc-anchoring\"]",
+            "no_user_services.toml",
```
dmitry-timofeev: Here and below, the file name is misleading. Why do we pass it *at all* if the method returns a path (to use the file name assertions :thinking:)?

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
```diff
@@ -41,116 +105,101 @@ fn service_factories() -> HashMap<String, Box<ServiceFactory>> {
...
-            assert!(service_factories.get(service).is_some())
-        }
+
+        let factories = prepare_service_factories(cfg);
+        assert_eq!(factories.len(), 5);
+        let f = factories
+            .iter()
+            .map(|factory| factory.service_name().to_string())
+            .collect::<Vec<String>>();
+        dbg!(f);
```
dmitry-timofeev: Is it temporary code?

exonum-java-binding/core/rust/integration_tests/tests/service_factory_adapter.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    // Make sure it returns None for all commands except `Run`.
+    assert!(factory_adapter.command("generate-template").is_none());
+    assert!(factory_adapter.command("generate-config").is_none());
+    assert!(factory_adapter.command("generate-testnet").is_none());
+    assert!(factory_adapter.command("finalize").is_none());
+    assert!(factory_adapter.command("run-dev").is_none());
+    assert!(factory_adapter.command("generate-template").is_none());
+
+    // Make sure it returns Some for the `Run` command for the first time and `None` for any other call.
```
dmitry-timofeev: I suppose among several instances of such factories _first_ must *always* return Some,
other — None: https://docs.rs/exonum/0.11.0/exonum/helpers/fabric/trait.ServiceFactory.html#method.command
The trait spec does not say "I will only call this method once".

exonum-java-binding/core/rust/src/utils/services.rs
```diff
@@ -14,36 +14,46 @@
...
+use std::collections::HashMap;
 use toml;
 
-pub const PATH_TO_SERVICES_TO_ENABLE: &str = "ejb_app_services.toml";
+pub const PATH_TO_SERVICES_DEFINITION: &str = "ejb_app_services.toml";
 pub const CONFIGURATION_SERVICE: &str = "configuration";
-pub const BTC_ANCHORING_SERVICE: &str = "btc-anchoring";
-pub const TIME_SERVICE: &str = "time";
-pub const EJB_SERVICE: &str = "ejb-service";
+pub const BTC_ANCHORING_SERVICE: &str = "btc_anchoring";
```
dmitry-timofeev: Why this change, requiring all docs updates?
I think the former are just as good (also, '_' is not used in Java artifact names, our Rust similarly do not use it).

time is good too.

exonum-java-binding/core/rust/src/utils/services.rs
```diff
@@ -53,45 +63,96 @@ mod tests {
     use tempfile::{Builder, TempPath};
 
     #[test]
-    fn no_config() {
-        let res = load_enabled_services("nonexistent");
-        assert!(res.is_err());
+    fn missed_system_services_section() {
```
dmitry-timofeev: At the moment it is unclear from method names if they test `parse_services` or the whole load definition. Could the thing under test be clarified in each case?

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
skletsun: Actually, I agree with you and in the branch with extended integration tests I removed the name parameter. Going to remove it here as well.

dmitry-timofeev: Well, as I mentioned, it _may_ be used in assertions on error messages, but it isn't now, so better be removed.

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
skletsun: Since it wasn't me, who introduced this piece of code, I considered it already reviewed and correct. From my understanding, removing semantic is used here in order to not to put service factory to the builder more than once. If we rely on checking only then it is possible to put it twice and more. @vitvakatu could you please elaborate on this (why removing is used)?
But on the other hand, we have a test for duplicates in the `system_services` section which relies on the `TOML` deserialization logic and thus with this particular implementation duplicates are not possible -> we could switch to checking.

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
skletsun: The idea is because this piece of code is used from the top-level code of the console app panics are pretty OK. We panic because we can't populate builder with required data.

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
skletsun: Yes, thanks for pointing it out...

exonum-java-binding/core/rust/src/utils/services.rs
skletsun: Because of testing of `node_builder::prepare_service_factories()`. The only way to check ServiceFactory is presented in the resulting vector is to iterate over the vector and compare for `ServiceFactory.service_name()`. And the names of system services (in their `ServiceFactory` implementations) are defined as "btc_anchoring", "configuration" and "exonum_time". So I just synchronized our constants with them in order to not support two sets of constants.

dmitry-timofeev: I see, but isn't it easier to add test-only constants than to have a configuration file dependent on the values from these services (and migrate all the docs, existent configuration files, putting that into changelog and the migration guide)?

skletsun: I'm in doubt these changes in constants are going to bring so many changes into the real picture, but I agree that it's better to add test-only constants.

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
vitvakatu: FYI, it's not supposed to be the full error message, but rather a substring of the message

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
vitvakatu: Honestly, can't remember. I think `get` is also okay

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
skletsun: Done. For the moment (until dynamic services) I maintain two copies of testing helper `create_config` in `core` and `ejb-app` because there is no way to share the common testing code for testing configuration only unless I create separate crate like "test_utils". Which looks kinda overhead for the moment but would be useful if we wanted such functionality in the integration tests.

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
skletsun: Actually `should_panic(expected)` checks for matching the beginning of error message. In fact, there will be smth like `Unable to load services definition: [REASON]`, where REASON could be smth like `Missing required "user_services"`

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
skletsun: And looks like this is the reason for an Option of return value :) https://docs.rs/exonum/0.11.0/exonum/helpers/fabric/trait.ServiceFactory.html#method.command

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
skletsun: Well, it turned out to be the more clear way to write this piece of code WRT the Rust move semantics with `.remove()` than using `.get()` or whatever else. I added a comment that notes that we're _moving_ factory instance to the new collection.

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
```diff
@@ -17,66 +17,78 @@
...
+use std::{
+    collections::{HashMap, HashSet},
+    path::Path,
+};
+
+/// Creates `NodeBuilder` using services configuration from `ejb_app_services.toml` located in the working directory.
+pub fn create() -> fabric::NodeBuilder {
+    let service_factories = prepare_service_factories(PATH_TO_SERVICES_DEFINITION);
+    let mut builder = fabric::NodeBuilder::new();
+    for (_, service_factory) in service_factories {
```
dmitry-timofeev: Why this private method returns a HashMap (and is more complicated) if it is not needed?

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
dmitry-timofeev: > panics are pretty OK

@vitvakatu Do you agree with this statement?

How the output looks like?

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
```diff
@@ -0,0 +1,115 @@
...
+                JAVA_SERVICE_RUNTIME = Some(runtime);
+            }
+        });
+
+        unsafe { JAVA_SERVICE_RUNTIME.clone() }
+            .expect("Trying to return runtime, but it's uninitialized")
+    }
+
+    // Extend command only when required.
+    fn provide_run_command_extension(&self) -> Option<Box<dyn CommandExtension>> {
```
dmitry-timofeev: Could we place the methods after their usages so that they are not scatterred randomly?

skletsun: As it doesn't belong to the trait implementation it's better to inline it in the parent method.

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
skletsun: Re-written back to Vector.

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
```diff
@@ -191,4 +188,20 @@ mod tests {
         writeln!(cfg_file, "{}", cfg).unwrap();
         cfg_file.into_temp_path()
     }
+
+    // Checks whether particular service is presented in the given collection.
+    fn contains_service(service_name: &str, factories: &[Box<dyn ServiceFactory>]) -> bool {
+        let name = if service_name == CONFIGURATION_SERVICE
```
dmitry-timofeev: I find this too fancy for a test utility method than a bunch of constants. Can't we use just the constants directly in assertions without these conversions (assert(contains_service(CONFIGURATION_FACTORY_NAME, ...)?

skletsun: Well, the previous variant with HashMap was used with constants directly with assertions, but it was a little overcomplicated for production code. With just a vector we have to iterate over it anyway, so this is the reason for `contains_service` method. Would it be better to make it `assert(contains_service(CONSTANT, collection))` -> `assert_contains_service(CONSTANT, collection)`?

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
```diff
@@ -17,66 +17,75 @@
...
-        }
+    resulting_factories
+}
+
+fn system_service_factory_for_name(name: &str) -> Box<dyn ServiceFactory> {
+    match name {
+        CONFIGURATION_SERVICE => Box::new(ConfigurationServiceFactory) as Box<ServiceFactory>,
+        BTC_ANCHORING_SERVICE => Box::new(BtcAnchoringServiceFactory) as Box<ServiceFactory>,
+        TIME_SERVICE => Box::new(TimeServiceFactory) as Box<ServiceFactory>,
+        _ => unreachable!("Unknown system service name \"{}\" has been found", name),
```
vitvakatu: This should be a panic actually, not `unreachable`

exonum-java-binding/core/rust/integration_tests/tests/service_factory_adapter.rs
```diff
@@ -0,0 +1,53 @@
...
+
+#[test]
+fn test_callbacks() {
+    const CMD_RUN: &str = "run";
+
+    let mut factory_adapter =
+        JavaServiceFactoryAdapter::new("service_name".to_owned(), "artifact_path".to_owned());
+    assert_eq!(factory_adapter.service_name(), "service_name");
+
+    // Make sure it returns None for all commands except `Run`.
```
vitvakatu: Interesting, why we're testing this, not Core? I don't think it's even possible for us to return something else.

exonum-java-binding/core/rust/integration_tests/tests/service_factory_adapter.rs
```diff
@@ -0,0 +1,53 @@
...
+    assert!(factory_adapter.command("generate-template").is_none());
+    assert!(factory_adapter.command("generate-config").is_none());
+    assert!(factory_adapter.command("generate-testnet").is_none());
+    assert!(factory_adapter.command("finalize").is_none());
+    assert!(factory_adapter.command("run-dev").is_none());
+    assert!(factory_adapter.command("generate-template").is_none());
+
+    // Make sure it always returns Some for the `Run` command.
+    let run_command_ext = factory_adapter.command("run");
```
vitvakatu: CMD_RUN?

exonum-java-binding/core/rust/ejb-app/src/node_builder.rs
dmitry-timofeev: No, I'd consider using the constants defined in each service in assertions instead of translating them with this conditional.

PR #789, opened 2019-03-19
Update the archetype to work with multiple services [ECR-3006]

dmitry-timofeev:
## Overview

Update the archetype so that it includes the configuration
to build an Exonum service artifact — a JAR with some metadata.

---
See: https://jira.bf.local/browse/ECR-3006

### Definition of Done

- [X] Change is covered by automated [tests](https://github.com/exonum/exonum-java-binding/blob/master/CONTRIBUTING.md#tests)
- [X] The [coding guidelines](https://github.com/exonum/exonum-java-binding/blob/master/CONTRIBUTING.md#the-code-style) are followed
- [X] Changelog is updated if needed (in case of notable or breaking changes)

### DIFF
--- a/exonum-java-binding/service-archetype/src/main/resources/archetype-resources/pom.xml
+++ b/exonum-java-binding/service-archetype/src/main/resources/archetype-resources/pom.xml
@@ -45,10 +45,10 @@
          in the service artifact. Some of transitive dependencies of Exonum are safe
          and recommended to use in a service in 'provided' scope as well, because they will not
          be changed in an incompatible way in a compatible release of Exonum. Such dependencies
-         are listed in exonum-java-binding-bom (e.g., guava, guice, gson, vertx, pf4j).
-         You can either use them implicitly as transitive dependencies (no extra POM modifications
-         needed) or define explicitly below with 'provided' scope and NO version — it will be
-         inherited from exonum-java-binding-bom.
+         are listed in exonum-java-binding-bom (e.g., guava, guice, gson, vertx, pf4j,
+         protobuf-java). You can either use them implicitly as transitive dependencies
+         (no extra POM modifications needed) or define explicitly below with 'provided' scope
+         and NO version — it will be inherited from exonum-java-binding-bom.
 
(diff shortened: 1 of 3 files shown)


dmitry-timofeev:
Not sure if we shall improve the ITs (or system tests) to check that the produced project can be loaded by Exonum. @dip56 WDYT?

exonum-java-binding/service-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -41,16 +41,24 @@
...
+         You can either use them implicitly as transitive dependencies (no extra POM modifications
+         needed) or define explicitly below with 'provided' scope and NO version — it will be
+         inherited from exonum-java-binding-bom.
+
+         See also https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Transitive_Dependencies
+         -->
     <dependency>
       <groupId>com.exonum.binding</groupId>
       <artifactId>exonum-java-binding-core</artifactId>
+      <scope>provided</scope>
```
bullet-tooth: I wonder if there any plugins to check scope for particular dependencies

dmitry-timofeev: Sounds interesting. I think we can make it a 'feature' with such implementation:
1. Load a plugin
2. Take its classloader
3. Try to resolve some Exonum class using this cl. If the resolved class is the same as we have (already loaded) — everything is fine, there is no Exonum in the plugin. If it is different — the plugin contains a copy of that Exonum class, which is forbidden, and must be rejected.

dmitry-timofeev: If we are determined enough, we can pick a class from each such dependency that **must** be 'provided' and do this check in a loop. As it happens once per service artifact, we can easily afford loading a couple of classes to prevent obscure ClassCastException (don't remember the exact type) "com.example.Foo cannot be assigned to com.example.Foo :trollface:"

dmitry-timofeev: https://jira.bf.local/browse/ECR-3046

PR #795, opened 2019-03-26
Test if all is fine on 12 and EA (why not?)

dmitry-timofeev:
## Overview

---

### Definition of Done


dmitry-timofeev:
12 is fine, EA does not work because EqualsVerifier needs to be updated (13, as any new release nowadays, bumps a class format version).

PR #815, opened 2019-04-01
Update library loading error message [ECR-3035]

dmitry-timofeev:
## Overview

Update the error message in case of library loading failure
so that it contains up-to-date instructions
for a packaged application.

Also, throw an exception with a full error message to make it
available even if no logging is configured.

---
See: https://jira.bf.local/browse/ECR-3035

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
```diff
@@ -46,26 +46,46 @@ private static void loadOnce() {
...
+    } else {
+      String dynamicLibVar = dynamicLibrariesEnvVar();
+      String dynamicLibValue = System.getenv(dynamicLibVar);
+      return String.format("Unknown error: please submit an issue including this error message.\n"
+              + "%s=%s\n%s=%s", JAVA_LIBRARY_PATH_PROPERTY, javaLibPath,
+          dynamicLibVar, dynamicLibValue);
+    }
+  }
+
+  private static boolean runningUnitTests() {
```
MakarovS: What do you think about moving this to a separate class and making this public so we could use this method in time service as well?

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
```diff
@@ -46,26 +46,46 @@ private static void loadOnce() {
...
+      // 1. These error messages are tailored for an installed application. If you *develop*
+      // Exonum, the instructions might not be applicable.
+      // 2. java.library.path has a default platform-specific value if not specified (on 11),
+      // hence it might never be empty.
+      return String.format(
+          "Check that %s system property includes a path to '${EXONUM_HOME}/lib/native' directory\n"
+              + "containing %s library, where 'EXONUM_HOME' denotes the Exonum installation "
+              + "directory.\n"
+              + "The code launching tests must usually set it explicitly, see "
+              + "https://exonum.com/doc/version/0.10/get-started/java-binding/#testing",
```
bullet-tooth: The description is a little confusing. The link doesn't say anything about Exonum installation. So it is not clear to me what path should I put for `EXONUM_HOME`. 
Also, I installed Exonum using https://exonum.com/doc/version/0.10/get-started/install and I do not see `/lib/native` directories inside neither the Exonum parent or `target` dirs.

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
dmitry-timofeev: I considered that, but this method seems so specific that I don't think we will need it anywhere else.

Won't we also remove the usage of the corresponding method from the time oracle when testkit is added?

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
```diff
@@ -46,26 +46,46 @@ private static void loadOnce() {
...
+      String dynamicLibValue = System.getenv(dynamicLibVar);
+      return String.format("Unknown error: please submit an issue including this error message.\n"
+              + "%s=%s\n%s=%s", JAVA_LIBRARY_PATH_PROPERTY, javaLibPath,
+          dynamicLibVar, dynamicLibValue);
+    }
+  }
+
+  private static boolean runningUnitTests() {
+    try {
+      Class.forName("org.junit.jupiter.api.Test");
```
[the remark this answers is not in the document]
dmitry-timofeev: > It could be broken if someone forgot to specify scope for JUnit

Do you mean they use the default 'compile' instead of 'test'? This code will be kind of broken, but I think we can live with that and use a simpler approach.

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
dmitry-timofeev: There is no installation section at the moment: https://jira.bf.local/browse/ECR-2990 and https://jira.bf.local/browse/ECR-3048 

ECR-2990 will update the testing section with appropriate surefire/failsafe configuration and link installation instructions.

dmitry-timofeev: At the moment we can clarify "Exonum _Java_ installation directory", I think

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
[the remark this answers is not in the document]
dmitry-timofeev: Also, libraries are loaded by our framework code whose classloader does not access to service classpath.

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
MakarovS: 👍 
> Won't we also remove the usage of the corresponding method from the time oracle when testkit is added?

Yes, indeed

dmitry-timofeev: Let's then keep this method under the covers.

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
vitvakatu: What about "Exonum Java *app* installation directory"?

PR #819, opened 2019-04-02
Add TestKit [ECR-3049]

MakarovS:
## Overview
Add TestKit P1 functionality.

---
See: https://jira.bf.local/browse/ECR-3049

### Definition of Done

- [x] Public API has Javadoc
- [x] Changelog is updated if needed (in case of notable or breaking changes) - will be updated in a later PR


exonum-java-binding/testkit/src/main/java/com.exonum.binding.testkit/TestKit.java
```diff
@@ -0,0 +1,197 @@
...
+  @VisibleForTesting
+  final static short MAX_SERVICE_NUMBER = 256;
+  private final static Injector frameworkInjector = Guice.createInjector(new TestKitFrameworkModule());
+
+  private final NativeHandle nativeHandle;
+  private final Map<Short, Service> services = new HashMap<>();
+
+  private TestKit(List<Class<? extends ServiceModule>> serviceModules, EmulatedNodeType nodeType,
+                  short validatorCount, @Nullable TimeProvider timeProvider) {
+    List<UserServiceAdapter> serviceAdapters = new ArrayList<>();
```
bullet-tooth:
```suggestion
    List<UserServiceAdapter> serviceAdapters = new ArrayList<>(serviceModules.size());
```

exonum-java-binding/testkit/src/main/java/com.exonum.binding.testkit/TestKit.java
```diff
@@ -0,0 +1,197 @@
...
+//    nativeHandle = new NativeHandle(
+//        nativeCreateTestKit(userServiceAdapters, isAuditorNode,
+//            validatorCount, timeProvider));
+  }
+
+  /**
+   * Creates a TestKit network with a single validator node for a single service.
+   */
+  public static TestKit forService(Class<? extends ServiceModule> serviceModule) {
+    return new TestKit(Collections.singletonList(serviceModule), EmulatedNodeType.VALIDATOR,
```
bullet-tooth: static imports

exonum-java-binding/testkit/src/main/java/com.exonum.binding.testkit/TestKit.java
```diff
@@ -0,0 +1,197 @@
...
+        (short) 0, null);
+  }
+
+  @SuppressWarnings("unchecked")
+  public <T extends Service> T getService(short serviceId, Class<T> serviceClass) {
+    Service service = services.get(serviceId);
+    checkArgument(service.getClass().equals(serviceClass), "Service with id %s is not of expected class %s",
```
bullet-tooth:
```suggestion
    checkArgument(service.getClass().equals(serviceClass), "Service with id %s is not an instance of expected class %s",
```

exonum-java-binding/testkit/src/main/java/com.exonum.binding.testkit/TestKit.java
```diff
@@ -0,0 +1,197 @@
...
+  /**
+   * Creates a service from the service module.
+   *
+   * @param moduleClass a class of the user service module
+   */
+  private static UserServiceAdapter createUserModule(Class<? extends ServiceModule> moduleClass) {
+    try {
+      Constructor constructor = moduleClass.getDeclaredConstructor();
+      Object moduleObject = constructor.newInstance();
+      checkArgument(moduleObject instanceof Module, "%s is not a sub-class of %s",
```
bullet-tooth:
```suggestion
      checkArgument(moduleObject instanceof Module, "%s is not a sub-class of the %s",
```

Btw, looks like this check is always `true` because `ServiceModule` extends `Module`. No? I believe we can cast to `ServiceModule` at initialization and remove this check.

exonum-java-binding/testkit/src/main/java/com.exonum.binding.testkit/TestKit.java
```diff
@@ -0,0 +1,197 @@
...
+    /**
+     * Creates the TestKit instance.
+     */
+    public TestKit build() {
+      checkMaxServiceNumber(services);
+      return new TestKit(services, nodeType, validatorCount, timeProvider);
+    }
+
+    private void checkMaxServiceNumber(List<Class<? extends ServiceModule>> serviceModules) {
+      checkArgument(serviceModules.size() < MAX_SERVICE_NUMBER,
```
bullet-tooth: Is it allowed to have 255 services, not 256?

exonum-java-binding/testkit/src/main/java/com.exonum.binding.testkit/TimeProvider.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.exonum.binding.testkit;
+
+import java.time.ZonedDateTime;
+
+public interface TimeProvider {
```
bullet-tooth: javadoc

exonum-java-binding/testkit/src/test/java/com.exonum.binding.testkit/TestKitTest.java
```diff
@@ -0,0 +1,178 @@
...
+        .withService(TestServiceModule.class)
+        .build();
+    Service service = testKit.getService(TestService.SERVICE_ID, TestService.class);
+    assertEquals(service.getId(), TestService.SERVICE_ID);
+    assertEquals(service.getName(), TestService.SERVICE_NAME);
+  }
+
+  @Test
+  void createTestKitWithBuilderForMultipleSameServices() {
+    List<Class<? extends ServiceModule>> serviceModules = new ArrayList<>();
```
bullet-tooth: `List.of(...)` or `ImmutableList.of()`

exonum-java-binding/testkit/src/main/java/com.exonum.binding.testkit/TestKit.java
```diff
@@ -0,0 +1,197 @@
...
+          "There shouldn't be more than %s services in the TestKit", MAX_SERVICE_NUMBER);
```
dmitry-timofeev: ... but x added.

exonum-java-binding/testkit/src/main/java/com.exonum.binding.testkit/TestKit.java
```diff
@@ -0,0 +1,197 @@
...
+
+    /**
+     * Creates the TestKit instance.
+     */
+    public TestKit build() {
+      checkMaxServiceNumber(services);
+      return new TestKit(services, nodeType, validatorCount, timeProvider);
+    }
+
+    private void checkMaxServiceNumber(List<Class<? extends ServiceModule>> serviceModules) {
```
dmitry-timofeev: Is there anything else that must be validated? E.g., non-empty list of services?

exonum-java-binding/testkit/src/main/java/com.exonum.binding.testkit/TimeProvider.java
dmitry-timofeev: Or in time-PR?

exonum-java-binding/testkit/src/test/java/com.exonum.binding.testkit/TestKitTest.java
```diff
@@ -0,0 +1,178 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+class TestKitTest {
+
+  @Test
+  void createTestKitForSingleService() {
+    TestKit testKit = TestKit.forService(TestServiceModule.class);
+    Service service = testKit.getService(TestService.SERVICE_ID, TestService.class);
+    assertEquals(service.getId(), TestService.SERVICE_ID);
```
dmitry-timofeev: Shan't we test the other side-effects of creating a testkit (= service initialization, mounting of web handlers)?
I think we can do that with a service implementation that produces some effects in its #initialize and returns
some value.

exonum-java-binding/testkit/src/test/java/com.exonum.binding.testkit/TestKitTest.java
```diff
@@ -0,0 +1,178 @@
...
+    assertEquals(service.getName(), TestService.SERVICE_NAME);
+  }
+
+  @Test
+  void createTestKitWithBuilderForMultipleSameServices() {
+    List<Class<? extends ServiceModule>> serviceModules = new ArrayList<>();
+    serviceModules.add(TestServiceModule.class);
+    serviceModules.add(TestServiceModule.class);
```
dmitry-timofeev: Shall we allow that? With the present framework version it will not work at all.

exonum-java-binding/testkit/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+
+  <artifactId>exonum-testkit</artifactId>
+  <version>0.6.0-SNAPSHOT</version>
+  <packaging>jar</packaging>
+
+  <name>Exonum TestKit</name>
+  <url>https://exonum.com/doc/version/latest/get-started/test-service/</url>
+
+  <properties>
+    <ejb-core.nativeLibPath>${project.parent.basedir}/core/rust/target/debug
```
dmitry-timofeev: I think since #818 it must be slightly different, @vitvakatu can advise

exonum-java-binding/testkit/src/main/java/com.exonum.binding.testkit/TestKit.java
dmitry-timofeev: if the constant is named max, it must be inclusive

exonum-java-binding/testkit/src/main/java/com.exonum.binding.testkit/TestKit.java
dmitry-timofeev: This code seems to be copied from the place where we instantiate a class and subsequently a module _by name_, therefore, some checks can be redundant.

I'd also look into `multiple-services` branch. There might be something reusable here.

dmitry-timofeev: `com.exonum.binding.runtime.ReflectiveModuleSupplier`

exonum-java-binding/testkit/src/main/java/com.exonum.binding.testkit/FakeTimeProvider.java
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.exonum.binding.testkit;
+
+import java.time.ZonedDateTime;
+
+public class FakeTimeProvider implements TimeProvider {
```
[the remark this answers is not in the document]
MakarovS: Yes, sorry for half-introducing classes from other tasks - I wanted native methods signatures to be ready for native tasks, so created these classes as placeholders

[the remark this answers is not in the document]
dmitry-timofeev: Let's keep them as placeholders to avoid redundant work then.

exonum-java-binding/testkit/src/main/java/com.exonum.binding.testkit/TestKit.java
```diff
@@ -0,0 +1,197 @@
...
+   */
+  public static final class Builder {
+
+    private EmulatedNodeType nodeType;
+    private short validatorCount;
+    private List<Class<? extends ServiceModule>> services = new ArrayList<>();
+    private TimeProvider timeProvider;
+
+    private Builder(EmulatedNodeType nodeType) {
+      this.nodeType = nodeType;
```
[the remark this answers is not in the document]
MakarovS: In core if TestKit is created with an auditor node and `validator_count` is set to `None` then it's unwrapped to `1`, so there is no way to create a network without at least one validator node. But I agree that it's better to explicitly state the number here.
I updated the code so that if the main node is an auditor, then set `validatorCount` to one, as if the main node is a validator, then `validatorCount` stays at 0. We could rename `validatorCount` to `additionalValidatorCount` in case it's not clear that the main node is not counted here

exonum-java-binding/testkit/src/main/java/com.exonum.binding.testkit/TestKit.java
MakarovS: I added such a check, even though it's possible to create a blockchain without any user services and if users want to run such a network (only with time service, for example), maybe we should let them, even if it's useless? But probably it's better to fail in such case to indicate that a user forgot to add a service

exonum-java-binding/testkit/src/test/java/com.exonum.binding.testkit/TestKitTest.java
MakarovS: IIRC there was a discussion on wiki some time ago stating that we should, but that could change. What's the current outlook on having several instances of the same service?

exonum-java-binding/testkit/pom.xml
vitvakatu: It's okay, as we don't have tests with release mode at the moment

exonum-java-binding/testkit/src/main/java/com.exonum.binding.testkit/EmulatedNode.java
```diff
@@ -0,0 +1,72 @@
...
+   * Returns a service key pair of this node. This key pair is used to sign transactions
+   * {@linkplain Node#submitTransaction(RawTransaction)} produced} by the service itself.
+   */
+  public KeyPair getServiceKeyPair() {
+    return serviceKeyPair;
+  }
+
+  /**
+   * Returns a consensus key pair of this node. This key pair is used to sign consensus message of
+   * this node.
```
skletsun: Nit: Plural for "transactions" in the comment for `.getServiceKeyPair()` and singular for "consensus message" here.

exonum-java-binding/testkit/src/test/java/com.exonum.binding.testkit/TestKitTest.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    @Override
+    public Transaction convertToTransaction(RawTransaction rawTransaction) {
+      throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void createPublicApiHandlers(Node node, Router router) {
+      throw new UnsupportedOperationException();
```
vitvakatu: We always call createPublicApiHandlers, even if no handlers are needed. I wonder if constructing this service will lead to panic

exonum-java-binding/testkit/src/main/java/com.exonum.binding.testkit/TestKit.java
```diff
@@ -0,0 +1,197 @@
...
+    /**
+     * If called, will create a TestKit with time service with given TimeProvider.
+     */
+    public Builder withTimeService(TimeProvider timeProvider) {
+      this.timeProvider = timeProvider;
+      return this;
+    }
+
+    /**
+     * Creates the TestKit instance.
```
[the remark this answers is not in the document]
vitvakatu: Should we mention genesis block creation? That's all I believe

exonum-java-binding/testkit/src/test/java/com.exonum.binding.testkit/TestKitTest.java
dmitry-timofeev: Even with multiple services, it is not possible to create two instances of the same service because their database schemas will be the same; their ids will be the same (so the testkit will likely fail somewhere in native). Dynamic services will require providing _instance_ parameters — i.e., something on top of a ServiceModule (e.g., a service instance name, or other initialization parameters), therefore the present interface won't work with them.

exonum-java-binding/testkit/src/test/java/com.exonum.binding.testkit/TestKitTest.java
MakarovS: I was planning to add that later, with the native part, rather than writing and ignoring them now. Should I do the latter?

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -0,0 +1,245 @@
...
+import java.util.function.Supplier;
+import javax.annotation.Nullable;
+
+/**
+ * TestKit for testing blockchain services. It offers simple network configuration emulation
+ * (with no real network setup). Although it is possible to add several validator nodes to this
+ * network, only one node will create the service instances and will execute their operations
+ * (e.g., {@link Service#afterCommit(BlockCommittedEvent)} method logic).
+ *
+ * <p>When TestKit is created, Exonum blockchain instance with given services is initialized and
```
dmitry-timofeev: @vitvakatu, Would you clarify please what is the exact order? Do we initialize after
genesis block? When do we create public API handlers?

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -0,0 +1,245 @@
...
+
+  /**
+   * Creates a TestKit network with a single validator node for a single service.
+   */
+  public static TestKit forService(Class<? extends ServiceModule> serviceModule) {
+    return new TestKit(singletonList(serviceModule), EmulatedNodeType.VALIDATOR, (short) 0, null);
+  }
+
+  /**
+   * Returns an instance of a service with the given service id and service class.
```
dmitry-timofeev: \+ "Only user-defined services can be requested, i.e., it is not possible to get an instance of a built-in service such as the time oracle."

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -0,0 +1,245 @@
...
+   *
+   * @return the service instance or null if there is no service with such id
+   * @throws NullPointerException if the service with given id was not found
```
dmitry-timofeev: It is not appropriate to throw NPE when it is not the client passing `null`s. `IllegalArgumentException` would be more appropriate (`checkArgument(services.containsKey…`)

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -0,0 +1,245 @@
...
+    Service service = services.get(serviceId);
+    checkNotNull(service, "Service with given id=%s was not found", serviceId);
+    checkArgument(service.getClass().equals(serviceClass),
+        "Service (id=%s, class=%s) cannot be cast to %s",
+        serviceId, service.getClass().getCanonicalName(), serviceClass.getCanonicalName());
+    return serviceClass.cast(service);
+  }
+
+  /**
+   * Creates a user service adapter from the service module.
```
dmitry-timofeev: Instantiates a service given its module and wraps in a UserServiceAdapter
for the native code. 

(we don't create a USA from the module, but a service, and then wrap it).

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -0,0 +1,245 @@
...
+   */
+  public static Builder builder(EmulatedNodeType nodeType) {
+    checkNotNull(nodeType);
+    return new Builder(nodeType);
+  }
+
+  /**
+   * Returns a list of user service adapters created from given service modules.
+   */
+  private List<UserServiceAdapter> toUserServiceAdapters(List<Class<? extends ServiceModule>> serviceModules) {
```
dmitry-timofeev: These methods are used by the constructor only, therefore, I'd move them closer to it
 - ctor
   - toUserServiceAdapters
     - createUserServiceAdapter
   - populateServiceMap

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -0,0 +1,245 @@
...
+    List<UserServiceAdapter> serviceAdapters = new ArrayList<>(serviceModules.size());
+    for (Class<? extends ServiceModule> moduleClass : serviceModules) {
+      UserServiceAdapter serviceAdapter = createUserServiceAdapter(moduleClass);
+      serviceAdapters.add(serviceAdapter);
+    }
+    return serviceAdapters;
```
dmitry-timofeev:
```suggestion
    return serviceModules.stream()
        .map(this::createUserServiceAdapter)
        .collect(Collectors.toList());
```

dmitry-timofeev: with static imports, probably, as @bullet-tooth would recommend :-)

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -0,0 +1,245 @@
...
+   */
+  private void populateServiceMap(List<UserServiceAdapter> serviceAdapters) {
+    for (UserServiceAdapter serviceAdapter: serviceAdapters) {
+      short serviceId = serviceAdapter.getId();
+      checkForDuplicateService(serviceId);
+      services.put(serviceId, serviceAdapter.getService());
+    }
+  }
+
+  private void checkForDuplicateService(short serviceId) {
```
dmitry-timofeev:
```suggestion
  private void checkForDuplicateService(UserServiceAdapter newService) {
    short serviceId = newService.getId();
    checkArgument(!services.containsKey(serviceId),
        "Service with id %s was added to the TestKit twice: %s and %s",
        serviceId, services.get(serviceId), newService.getService());
  }
```

dmitry-timofeev: Otherwise it might not be clear which, ids are just numbers

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -0,0 +1,245 @@
...
+    /**
+     * Adds a service with which the TestKit would be instantiated. Several services can be added.
+     */
+    public Builder withService(Class<? extends ServiceModule> serviceModule) {
+      services.add(serviceModule);
+      return this;
+    }
+
+    /**
+     * Adds a variable number of services with which the TestKit would be instantiated.
```
dmitry-timofeev: Here and below: omit the type (below it is no longer correct as any iterable is allowed, not just list): "Adds services with which the TestKit would be instantiated."

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -0,0 +1,245 @@
...
+     */
+    public Builder withServices(Class<? extends ServiceModule> serviceModule,
```
dmitry-timofeev: It seems to qualify for `@SafeVarargs`, but double-check that

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -0,0 +1,245 @@
...
+      List<Class<? extends ServiceModule>> services = asList(serviceModule, serviceModules);
+      this.services.addAll(services);
+      return this;
+    }
+
+    /**
+     * Adds a list of services with which the TestKit would be instantiated.
+     */
+    public Builder withServices(Iterable<Class<? extends ServiceModule>> serviceModules) {
+      serviceModules.forEach(services::add);
```
dmitry-timofeev: I find this one a little clearer (addAll instead of generic for each):
```suggestion
      Iterables.addAll(services, serviceModules);
```

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
dmitry-timofeev: If they are private, have only a single usage and are adjacent to it, some documentation comments might no longer needed (unless they clarify something possibly unclear).

exonum-java-binding/testkit/src/test/java/com.exonum.binding.testkit/TestKitTest.java
dmitry-timofeev: I think either is OK, if we don't forget to add the proper tests/update the already added to test all aspects.

MakarovS: Ok, I'll create a separate task for that then.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
vitvakatu: The order is `initialize -> genesis block commit -> create handlers`. Service::initialize is called on genesis block creation, so actual it looks more like `initialize and genesis block commit -> create handlers`

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
dmitry-timofeev: … and `checkForDuplicateService` obviously

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
dmitry-timofeev: Thanks, it seems this description needs to be updated to reflect the actual order of operations.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -0,0 +1,243 @@
...
+
+    /**
+     * Adds services with which the TestKit would be instantiated.
+     */
+    @SafeVarargs
+    public final Builder withServices(Class<? extends ServiceModule> serviceModule,
+                                      Class<? extends ServiceModule>... serviceModules) {
+      List<Class<? extends ServiceModule>> services = asList(serviceModule, serviceModules);
+      this.services.addAll(services);
+      return this;
```
dmitry-timofeev:
```suggestion
      return withServices(asList(serviceModule, serviceModules));
```

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -0,0 +1,241 @@
...
+      }
+      this.nodeType = nodeType;
+    }
+
+    /**
+     * Sets number of additional validator nodes in the TestKit network. Note that
+     * regardless of the configured number of validators, only a single service will be
+     * instantiated.
+     */
+    public final Builder withValidators(short validatorCount) {
```
bullet-tooth: `final` is redundant for methods such as the class is final.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -0,0 +1,241 @@
...
+    public final Builder withService(Class<? extends ServiceModule> serviceModule) {
+      services.add(serviceModule);
+      return this;
+    }
+
+    /**
+     * Adds services with which the TestKit would be instantiated.
+     */
+    @SafeVarargs
+    public final Builder withServices(Class<? extends ServiceModule> serviceModule,
```
bullet-tooth: actually, `serviceModule` param can be removed and `java.util.Arrays#asList` is used as the varargs wrapper. Or did I misunderstand something?)

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
dmitry-timofeev: Yep, I think it is only required (even in final classes) for @SafeVarargs methods.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
bullet-tooth: > a requirement of at least one service module
There is `withService` in this case.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
MakarovS: Yes, it was needed for @SafeVarargs method, so I added others for consistency.

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -0,0 +1,208 @@
...
+    @Override
+    protected void configure() {
+      bind(Service.class).to(TestService2.class).in(Singleton.class);
+      bind(TransactionConverter.class).toInstance(THROWING_TX_CONVERTER);
+    }
+  }
+
+  static final class TestService implements Service {
+
+    static short SERVICE_ID = 42;
```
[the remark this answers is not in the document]
MakarovS: Yes, I thought about that, but that shouldn't be a problem

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
MakarovS: Are there any disadvantages to the present signature? If not, I'd leave it as it is

MakarovS:
It appears that in native we can't get a private consensus key, so after discussion with @vitvakatu we decided to remove consensus keys altogether (even though the public key is available), as it seems consensus keys would be not necessary for users. But we could give access to consensus public key or even change the core and make it possible to access both public and private consensus keys. Does anyone sees any usefulness in those? @dmitry-timofeev @bullet-tooth

dmitry-timofeev:
I don't see immediate usefulness in the consensus public key access. You can validate precommits and other consensus messages with it — but why would you?

PR #824, opened 2019-04-08
Fix exonum-java scripts and clarify their usage [ECR-3094]

vitvakatu:
## Overview

---
See: https://jira.bf.local/browse/ECR-3094

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/rust/exonum-java/start_cryptocurrency_node.sh
+++ b/exonum-java-binding/core/rust/exonum-java/start_cryptocurrency_node.sh
@@ -1,4 +1,10 @@
 #!/usr/bin/env bash
+# Runs exonum-java executable from target directory and starts one Exonum node with
+# cryptocurrency service.
+#
+# This script is intended to be used by Exonum developers to speed up working workflow.
+# Running the executable from target doesn't need packaging and therefore can be performed
+# immediately after running tests, without additional recompilation.
 
 set -eu -o pipefail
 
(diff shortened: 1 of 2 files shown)


exonum-java-binding/core/rust/exonum-java/start_cryptocurrency_node.sh
```diff
@@ -49,29 +46,29 @@ header "PREPARE PATHS"
 
 EJB_LOG_CONFIG_PATH="${EJB_APP_DIR}/log4j-fallback.xml"
 
-export LD_LIBRARY_PATH="$JVM_LIB_PATH"
+export LD_LIBRARY_PATH="$JAVA_LIB_DIR"
```
dmitry-timofeev: I wonder if it is OK to use `JAVA_LIB_DIR` variable from the tests_profile :thinking: I considered setting RUSTFLAGS to be the only effect of tests_profile, not these intermediate variables. Shall we make it required to set this variable as well?

vitvakatu: These scripts are effectively protected from unset variables by `set -eu -o pipefail`. I think we can safely rely on the current tests_profile implementation, as these scripts are intended to be used by us (and any problem will be spotted just in time).

exonum-java-binding/core/rust/exonum-java/start_cryptocurrency_node.sh
```diff
@@ -1,4 +1,10 @@
 #!/usr/bin/env bash
+# Runs exonum-java executable from target directory and starts one Exonum node with
+# cryptocurrency service.
+#
+# This script is intended to be used by Exonum developers to speed up working workflow.
```
skletsun: "working workflow" -> just "workflow"?

exonum-java-binding/core/rust/exonum-java/start_qa_node.sh
```diff
@@ -1,4 +1,10 @@
 #!/usr/bin/env bash
+# Runs exonum-java executable from target directory and starts N Exonum nodes with
+# QA service, where N corresponds to the first argument of the script.
+#
+# This script is intended to be used by Exonum developers to speed up working workflow.
```
skletsun: Same again.

PR #830, opened 2019-04-10
Check no copies of application libraries in artifact [ECR-3046]:

dmitry-timofeev:
## Overview
Check that no copies of application libraries are included in the
service artifact during artifact loading. It helps to prevent
errors in build definitions when a dependency does not have
'provided' scope set, as recently discovered ECR-3091.

The present version cannot check if time-oracle is included
because it is not available on the core classpath.

---
See: 
- https://jira.bf.local/browse/ECR-3046
- https://jira.bf.local/browse/ECR-3091

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/src/main/java/com/exonum/binding/runtime/ClassLoadingScopeChecker.java
+++ b/exonum-java-binding/core/src/main/java/com/exonum/binding/runtime/ClassLoadingScopeChecker.java
@@ -53,38 +53,16 @@ class ClassLoadingScopeChecker {
     this.dependencyReferenceClasses = dependencyReferenceClasses;
   }
 
-  // todo: withOptionalDependencies (recommended)?
-
-  // todo: Shall we use "plugin classloader" (PF4J specific term) or "module classloader"?
   /**
    * Checks if there are copies of application classes on the classpath of the given classloader.
    * @param pluginClassloader a plugin parent-last classloader
    * @throws IllegalArgumentException if a copy of an application class is available on the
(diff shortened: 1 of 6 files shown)


exonum-java-binding/core/src/main/java/com/exonum/binding/runtime/ClassLoadingScopeChecker.java
```diff
@@ -0,0 +1,124 @@
...
+   * Checks if there are copies of application classes on the classpath of the given classloader.
+   * @param pluginClassloader a plugin parent-last classloader
+   * @throws IllegalArgumentException if a copy of an application class is available on the
+   *     classpath of the given classloader
+   * @throws IllegalStateException if the given classloader fails to delegate
+   *     to the application classloader
+   */
+  void checkNoCopiesOfAppClasses(ClassLoader pluginClassloader) {
+    /*
+    todo: Would you prefer the iterative version?
```
bullet-tooth: I like streams :)

dmitry-timofeev: Ok, let's use them (there is less going on in the `loadsCopyOf` method body than in the loop body)

dmitry-timofeev:
The first :trophy: of this checker: missed changes from 7706e83cfbe66c2a21350f4f506e013971b3b562 during merge of `master` into `multiple-services`.

PR #833, opened 2019-04-11
Add TestKit P2 methods [ECR-3050]

MakarovS:
## Overview
Add TestKit P2 methods, as well as `nativeFreeTestKit` method.

Questions:
Should we give a link to documentation in every method that mentions `pool`?
Is there a way to avoid making `frameworkInjector` static?

---
See: https://jira.bf.local/browse/ECR-3050

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -138,6 +162,104 @@ public static TestKit forService(Class<? extends ServiceModule> serviceModule) {
...
+  }
+
+  /**
+   * Creates a block with all in-pool transactions.
+   *
+   * @return created block
+   * @see <a href="https://exonum.com/doc/version/0.10/advanced/consensus/specification/#pool-of-unconfirmed-transactions">Pool of Unconfirmed Transactions</a>
+   */
+  public Block createBlock() {
+    byte[] block = nativeCreateBlock(nativeHandle.get());
```
vitvakatu: Native method returns null on error

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -138,6 +162,104 @@ public static TestKit forService(Class<? extends ServiceModule> serviceModule) {
...
+  /**
+   * Performs a given function with a snapshot of the current database state.
+   *
+   * @param snapshotFunction a function to execute
+   * @param <ResultT> a type the function returns
+   * @return the result of applying the given function to the database state
+   */
+  public <ResultT> ResultT withSnapshot(Function<Snapshot, ResultT> snapshotFunction) {
+    try (Cleaner cleaner = new Cleaner("TestKit#withSnapshot")) {
+      long snapshotHandle = nativeCreateSnapshot(nativeHandle.get());
```
vitvakatu: Returns null on error

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -138,6 +162,104 @@ public static TestKit forService(Class<? extends ServiceModule> serviceModule) {
...
+    } catch (CloseFailuresException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  /**
+   * Returns the emulated TestKit node context.
+   */
+  public EmulatedNode getEmulatedNode() {
+    return nativeGetEmulatedNode(nativeHandle.get());
```
vitvakatu: returns null on error

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -138,6 +162,104 @@ public static TestKit forService(Class<? extends ServiceModule> serviceModule) {
...
+
+  /**
+   * Creates a block with the given transaction. In-pool transactions will be ignored.
+   *
+   * @return created block
+   * @see <a href="https://exonum.com/doc/version/0.10/advanced/consensus/specification/#pool-of-unconfirmed-transactions">Pool of Unconfirmed Transactions</a>
+   */
+  public Block createBlockWithTransaction(TransactionMessage transaction) {
+    byte[][] transactionMessageArray = { transaction.toBytes() };
+    byte[] block = nativeCreateBlockWithTransactions(nativeHandle.get(), transactionMessageArray);
```
vitvakatu: .... and this one

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -138,6 +162,104 @@ public static TestKit forService(Class<? extends ServiceModule> serviceModule) {
...
+  }
+
+  /**
+   * Creates a block with the given transactions. In-pool transactions will be ignored.
+   *
+   * @return created block
+   * @see <a href="https://exonum.com/doc/version/0.10/advanced/consensus/specification/#pool-of-unconfirmed-transactions">Pool of Unconfirmed Transactions</a>
+   */
+  public Block createBlockWithTransactions(Iterable<TransactionMessage> transactions) {
+    List<TransactionMessage> messageList = Lists.newArrayList(transactions);
```
bullet-tooth: this intermediate list can be replaced by `com.google.common.collect.Streams#stream(java.lang.Iterable<T>)`

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -138,6 +162,104 @@ public static TestKit forService(Class<? extends ServiceModule> serviceModule) {
...
+  }
+
+  /**
+   * Creates a block with the given transaction. In-pool transactions will be ignored.
+   *
+   * @return created block
+   * @see <a href="https://exonum.com/doc/version/0.10/advanced/consensus/specification/#pool-of-unconfirmed-transactions">Pool of Unconfirmed Transactions</a>
+   */
+  public Block createBlockWithTransaction(TransactionMessage transaction) {
+    byte[][] transactionMessageArray = { transaction.toBytes() };
```
bullet-tooth: please reuse the code

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -138,6 +162,104 @@ public static TestKit forService(Class<? extends ServiceModule> serviceModule) {
...
+  /**
+   * Returns a list of in-pool transactions that match the given predicate.
+   *
+   * @see <a href="https://exonum.com/doc/version/0.10/advanced/consensus/specification/#pool-of-unconfirmed-transactions">Pool of Unconfirmed Transactions</a>
+   */
+  public List<TransactionMessage> findTransactionsInPool(Predicate<TransactionMessage> predicate) {
+    return withSnapshot((view) -> {
+      Blockchain blockchain = Blockchain.newInstance(view);
+      MapIndex<HashCode, TransactionMessage> txMessages = blockchain.getTxMessages();
+      List<TransactionMessage> messageList = ImmutableList.copyOf(txMessages.values());
```
bullet-tooth: also, can be replaced by `Stream.stream(txMessages.values())`

dmitry-timofeev:
> Returns null on error

@vitvakatu, 
1. Do these errors have any context (description)? If so, it would be appropriate to include it in the error message of an exception instead of returning `null`. 
2. Under which circumstances can these errors occur? 
3. Are these errors the client’s fault? If not, who is to blame in case of these errors?

dmitry-timofeev:
Simply putting `checkNotNull(ret)` all over the place does not seem like a good idea, neither does `throw new UnknownError("Have no idea what happened, but here you go")`

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
MakarovS: `Streams` are marked as `@Beta`, is it okay for us?

dmitry-timofeev: No, for a library no.

dmitry-timofeev: That's the reason we repackaged HashCode and friends.

dmitry-timofeev: https://jira.bf.local/browse/ECR-1530

MakarovS:
About native methods returning null - @vitvakatu correct me if I'm wrong, but you said that it can only happen on panic in core, in which case we will get a RuntimeException rather than null?

vitvakatu:
@MakarovS yep

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -128,55 +148,184 @@ void createTestKitWithoutServices() {
...
+  void initializationChangesState() {
+    TestKit testKit = TestKit.forService(TestServiceModule.class);
+    Map<HashCode, String> map = testKit.withSnapshot((view) -> {
+      TestService testService = testKit.getService(TestService.SERVICE_ID, TestService.class);
+      TestSchema testSchema = testService.createDataSchema(view);
+      ProofMapIndexProxy<HashCode, String> proofMapIndexProxy = testSchema.testMap();
+      return toMap(proofMapIndexProxy);
+    });
+    assertThat(map).hasSize(1);
+    String initialValue = map.get(TestService.INITIAL_ENTRY_KEY);
```
bullet-tooth: `org.assertj.core.api.AbstractMapAssert#containsEntry`

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/EmulatedNode.java
```diff
@@ -32,8 +32,7 @@
   /**
    * Creates a context of an emulated node.
    *
-   * @param validatorId validator id of the validator node, less or equal to 0 in case of an
-   *     auditor node
+   * @param validatorId validator id of the validator node, less than 0 in case of an auditor node
```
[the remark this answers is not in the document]
bullet-tooth: Also, take a look at #841. `-1` is used as the special value for auditor node

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
dmitry-timofeev: Or 
```java
    Map<HashCode, String> expected = ImmutableMap.of(.., ...);
    assertThat(map).isEqualTo(expected);
```

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -128,55 +148,184 @@ void createTestKitWithoutServices() {
...
 
-  public static final class TestServiceModule2 extends AbstractServiceModule {
+  @Test
+  void afterCommitSubmitsTransaction() {
+    TestKit testKit = TestKit.forService(TestServiceModule.class);
 
-    private static final TransactionConverter THROWING_TX_CONVERTER = (tx) -> {
-      throw new IllegalStateException("No transactions in this service: " + tx);
-    };
+    Block block = testKit.createBlock();
```
dmitry-timofeev: It is unclear why this is invoked.

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -128,55 +148,184 @@ void createTestKitWithoutServices() {
...
   }
 
-  static final class TestService implements Service {
+  @Test
+  void createBlockWithTransactionIgnoresInPoolTransactions() {
+    TestKit testKit = TestKit.forService(TestServiceModule.class);
 
-    static short SERVICE_ID = 46;
-    static String SERVICE_NAME = "Test service";
+    testKit.createBlock();
```
dmitry-timofeev: It is unclear why this is invoked.

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -128,55 +148,184 @@ void createTestKitWithoutServices() {
...
+    assertThat(inPoolTransactions).hasSize(2);
+  }
 
-    @Override
-    public Transaction convertToTransaction(RawTransaction rawTransaction) {
-      throw new UnsupportedOperationException();
-    }
+  @Test
+  void createBlockWithTransaction() {
+    TestKit testKit = TestKit.forService(TestServiceModule.class);
```
dmitry-timofeev: Why do all these tests leak the native object?

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -128,55 +148,184 @@ void createTestKitWithoutServices() {
...
+      assertThat(transactionResult).isEqualTo(TransactionResult.successful());
+      TransactionResult transactionResult2 = transactionResults.get(message2.hash());
+      assertThat(transactionResult2).isEqualTo(TransactionResult.successful());
+      return null;
+    });
+  }
+
+  @Test
+  void createBlockWithTransactionsVarargs() {
+    TestKit testKit = TestKit.forService(TestServiceModule.class);
```
dmitry-timofeev: The code below seems to be the same as above except the function. I'd consider de-duplicating.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -138,7 +166,103 @@ public static TestKit forService(Class<? extends ServiceModule> serviceModule) {
     return serviceClass.cast(service);
   }
 
-  private native long nativeCreateTestKit(UserServiceAdapter[] services, boolean auditor,
+  /**
+   * Creates a block with the given transaction. In-pool transactions will be ignored.
```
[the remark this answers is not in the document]
MakarovS: Currently no, but we could combine some core TestKit methods and add such functionality. Should I add such a method (that commits both input and in-pool transactions) or change this into one?

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
MakarovS: We want to create a block so that `afterCommit` is invoked and we can test it (by verifying that it indeed submits a transaction). Added a comment

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
MakarovS: We want to create a block so that `afterCommit` submits a transaction into the pool and when we create the block with specified transactions, it ignores this in-pool transaction - then we check that it is indeed still in the pool (together with the second `afterCommit` transaction)

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -138,7 +166,103 @@ public static TestKit forService(Class<? extends ServiceModule> serviceModule) {
...
+   * @see <a href="https://exonum.com/doc/version/0.10/advanced/consensus/specification/#pool-of-unconfirmed-transactions">Pool of Unconfirmed Transactions</a>
+   */
+  public List<TransactionMessage> findTransactionsInPool(Predicate<TransactionMessage> predicate) {
+    return withSnapshot((view) -> {
+      Blockchain blockchain = Blockchain.newInstance(view);
+      MapIndex<HashCode, TransactionMessage> txMessages = blockchain.getTxMessages();
+      // As only executed transactions are stored in TxResults, it wouldn't contain in-pool
+      // transactions
+      ProofMapIndexProxy<HashCode, TransactionResult> txResults = blockchain.getTxResults();
+      List<TransactionMessage> messages = ImmutableList.copyOf(txMessages.values());
```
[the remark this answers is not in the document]
MakarovS: > It is somewhat weird (and inefficient) that one shall go over all transactions submitted to blockchain ever to find in-pool ones. I'd submit an issue to consider providing access to transaction_pool in Blockchain (or separately).

I looked at it and core schema only returns KeySetIndex of transaction hashes of in-pool transactions, not transactions themselves. So if we want to look up transactions using predicate, that won't work. However, we could look up transactions by hashes instead.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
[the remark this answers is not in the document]
dmitry-timofeev: What are the implications (in terms of testkit usage to test services) of either way?

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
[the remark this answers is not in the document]
dmitry-timofeev: Why wouldn't it work? It will have to request the messages from the same txMessages map (`hashes.stream().map(txMessages::get)`), but it will do O(num_in_pool) instead of O(num_ever_sumbitted).

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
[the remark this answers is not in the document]
MakarovS: I'd add both ways, it seems useful to be able to create a block that consists of both in-pool and incoming transactions. We could use core `add_tx` method for that - add incoming transactions to the pool and then commit block with all in-pool transactions, but I remember there is some problem with `add_tx` in `0.11` - @vitvakatu is that still so? Does that prevent us from using this method?

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -138,7 +166,103 @@ public static TestKit forService(Class<? extends ServiceModule> serviceModule) {
...
+    return createBlockWithTransactions(transaction);
+  }
+
+  /**
+   * Creates a block with the given transaction(s). In-pool transactions will be ignored.
+   *
+   * @return created block
+   * @see <a href="https://exonum.com/doc/version/0.10/advanced/consensus/specification/#pool-of-unconfirmed-transactions">Pool of Unconfirmed Transactions</a>
+   */
+  public Block createBlockWithTransactions(TransactionMessage transaction,
```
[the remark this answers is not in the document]
MakarovS: Yes, I believe it's better to distinguish with/without in-pool methods. Especially since we might want to commit an empty block without in-pool transactions - `createBlockWithTransactions()`

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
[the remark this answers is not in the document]
MakarovS: Added https://github.com/exonum/exonum-java-binding/pull/850 for that

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -16,31 +16,53 @@
...
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Maps;
 import com.google.inject.Singleton;
 import io.vertx.ext.web.Router;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import org.junit.jupiter.api.Test;
 
 class TestKitTest {
```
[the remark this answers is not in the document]
MakarovS: Added those. Correctness of arguments passed to afterCommit is tested in `afterCommitSubmitsTransaction` - we check that `afterCommitTransaction` has the payload with correct blockchain height

> and is provided with a Node that has the same identifiers as EmulatedNode.

Good that we added that test, it uncovered a bug in native part

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -138,7 +166,103 @@ public static TestKit forService(Class<? extends ServiceModule> serviceModule) {
...
+    return withSnapshot((view) -> {
+      Blockchain blockchain = Blockchain.newInstance(view);
+      MapIndex<HashCode, TransactionMessage> txMessages = blockchain.getTxMessages();
+      // As only executed transactions are stored in TxResults, it wouldn't contain in-pool
+      // transactions
+      ProofMapIndexProxy<HashCode, TransactionResult> txResults = blockchain.getTxResults();
+      List<TransactionMessage> messages = ImmutableList.copyOf(txMessages.values());
+      return messages.stream()
+          .filter(predicate)
+          .filter(tx -> !txResults.containsKey(tx.hash()))
```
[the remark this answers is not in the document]
MakarovS: This is `@since 11`

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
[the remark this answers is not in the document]
dmitry-timofeev: I am not sure that answers the question. Why would anyone writing a test ignore the in-pool transactions?

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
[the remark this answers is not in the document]
dmitry-timofeev: I think we must first determine why would anyone want to ignore the in-pool transactions (basically, going around how Exonum works). If there is a good reason for that — _then_ I'd distringuish such classes of methods, and, probably, with a more verbose name (e.g., `commitBlock` vs `commitBlockWithOnlyTransactions`).

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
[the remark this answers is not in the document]
dmitry-timofeev: ... not the Guava `Predicates#not` though

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -228,7 +354,10 @@ public Builder withTimeService(TimeProvider timeProvider) {
      */
     public TestKit build() {
       checkCorrectServiceNumber(services.size());
-      return new TestKit(services, nodeType, validatorCount, timeProvider);
+      if (nodeType == EmulatedNodeType.VALIDATOR) {
+        validatorCount += 1;
```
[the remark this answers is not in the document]
dmitry-timofeev: I still don't think it is OK as this comment does not explain why this complication is needed.  Why don't we just set the total number of validators, not additional, and assume "1" by default?

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -65,48 +91,126 @@ void createTestKitWithBuilderForMultipleSameServices() {
...
+
+    // Check that TestService API is mounted
+    Node serviceNode = service.getNode();
+    EmulatedNode emulatedTestKitNode = testKit.getEmulatedNode();
+    assertThat(serviceNode.getPublicKey())
+        .isEqualTo(emulatedTestKitNode.getServiceKeyPair().getPublicKey());
+    testKit.withSnapshot((view) -> {
+      // Check that initialization changed database state
+      TestSchema testSchema = service.createDataSchema(view);
+      ProofMapIndexProxy<HashCode, String> proofMapIndexProxy = testSchema.testMap();
```
dmitry-timofeev: testProofMap for this and testMap for `j.u.Map`?

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -128,55 +232,202 @@ void createTestKitWithoutServices() {
...
+    }
+  }
 
-    @Override
-    protected void configure() {
-      bind(Service.class).to(TestService.class).in(Singleton.class);
-      bind(TransactionConverter.class).toInstance(THROWING_TX_CONVERTER);
+  @Test
+  void afterCommitSubmitsTransaction() {
+    Block nextBlock;
```
dmitry-timofeev: (Here and elsewhere) Why don't we declare these variables inside try-catch and initialize immediately where they are needed?

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -128,55 +232,202 @@ void createTestKitWithoutServices() {
...
+
+    TransactionResult transactionResult = transactionResults.get(message.hash());
+    assertThat(transactionResult).isEqualTo(TransactionResult.successful());
+    TransactionResult transactionResult2 = transactionResults.get(message2.hash());
+    assertThat(transactionResult2).isEqualTo(TransactionResult.successful());
+  }
+
+  @Test
+  void createBlockWithTransactionWithWrongServiceId() {
+    Class<RuntimeException> exceptionType = RuntimeException.class;
```
dmitry-timofeev: (1) The thrown exception type is on the wrong abstraction level, RuntimeException communicates nothing. Since it is a testkit,
 I would expect either IllegalArgumentException or https://ota4j-team.github.io/opentest4j/docs/current/api/org/opentest4j/AssertionFailedError.html
(2) As we established in the requirements, the error messages must be user-friendly and tell exactly what
the problem is (or which problems are possible). In this case I think there are two possible problems:
   - A test developer passed an incorrectly serialized transaction (they intended to pass a proper one).
     The implementation of the service is fine.
   - A test dev passed a correctly serialized transaction, but the implementation of the service
     (particularly, TransactionConverter) does not handle that transaction correctly.

In both cases, the actual thrown exception *and* the details about the transaction message are quite useful
to understand what is broken.

Speaking of the implementation, it might be more convenient to check that each tx message is correct and create
a proper exception _in Java_, before passing them to the native code.

dmitry-timofeev: There might be a third case which we do not explicitly support — intentionally passing an invalid transaction. I think we shall either recommend testing that separately from the testkit (just take your TxConverter); or add first-class support (checkCannotDeserialize?). I think the first option is fine for now.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -18,32 +18,51 @@
...
+import java.util.function.Predicate;
 import java.util.function.Supplier;
 import javax.annotation.Nullable;
 
 /**
  * TestKit for testing blockchain services. It offers simple network configuration emulation
  * (with no real network setup). Although it is possible to add several validator nodes to this
- * network, only one node will create the service instances and will execute their operations
- * (e.g., {@link Service#afterCommit(BlockCommittedEvent)} method logic).
+ * network, only one node will create the service instances, provide access to its
```
dmitry-timofeev: this pool thing seems out of place, I'd move it in a separate sentence.

"
... only one node will create ..., execute their operations (...), and provide access to its state [here I mean Snapshots of any state]. 

Only the emulated node [but what in case it is an auditor?] has a pool of unconfirmed transactions where a service can submit new transaction messages through Node#submitTransaction; or the test code through #commitBlock. All transactions from the pool are committed when a new block is created [is that a reasonable default?].

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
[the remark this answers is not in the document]
MakarovS: Ok, I'll change that. I thought it was easier for users to deal with `additional` validators and distinguish these `additional` ones as not really emulated. We would have to deal with a little more complicated internal code, but maybe it would make things a bit more straightforward for users.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
MakarovS: What difference does an auditor node make? I don't see how it matters here, as we don't deal with consensus.

> or the test code through #commitBlock

Could you expand on this - what `commitBlock` do you mean?

dmitry-timofeev: I don't know, ask because it is unclear if an auditor _needs_ to have a pool, because it does not participate in consensus and can process only already committed transaction.

--- 

Any commitBlock in testkit that accepts transactions.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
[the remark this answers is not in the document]
dmitry-timofeev: I see, but the current code does not say that anywhere but in the docs, therefore, gets confusing. Are you saying that people _need_ additional validators quite rarely as it changes almost nothing in testkit behaviour? If this distinction is important, we can leave it as is, but make it clear in names of identifiers.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -206,10 +205,43 @@ public Block createBlockWithTransactions(Iterable<TransactionMessage> transactio
...
+      String message = String.format("Unknown service id (%s) in transaction message (%s)",
+          serviceId, transactionMessage);
+      throw new IllegalArgumentException(message);
+    }
+    Service service = services.get(serviceId);
+    try {
+      RawTransaction rawTransaction = toRawTransaction(transactionMessage);
+      service.convertToTransaction(rawTransaction);
+    } catch (Throwable conversionError) {
+      String message = String.format("Transaction message (%s) is invalid", transactionMessage);
```
dmitry-timofeev: I specifically requested this error message to be _actionable_, and described the possible cases. Please fix the error message to reflect them.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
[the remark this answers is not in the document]
dmitry-timofeev: Why this code still iterates over all messages?

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -138,7 +170,137 @@ public static TestKit forService(Class<? extends ServiceModule> serviceModule) {
...
+      KeySetIndexProxy<HashCode> poolTxsHashes = blockchain.getTransactionPool();
+      Set<HashCode> poolTxsHashesSet = toSet(poolTxsHashes);
+      return poolTxsHashesSet.stream()
+          .map(txMessages::get)
+          .collect(toList());
+    });
+  }
+
+  private <T> Set<T> toSet(KeySetIndexProxy<T> setIndex) {
+    return Sets.newHashSet(setIndex.iterator());
```
dmitry-timofeev: I would suggest using LinkedHashSet to preserve iteration order (= the order in which txs will be committed).
```suggestion
    return Sets.newLinkedHashSet(setIndex.iterator());
```

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -128,55 +237,209 @@ void createTestKitWithoutServices() {
...
+          .serviceId(wrongServiceId)
+          .transactionId(TestTransaction.ID)
+          .payload("Test message".getBytes(BODY_CHARSET))
+          .sign(KEY_PAIR, CRYPTO_FUNCTION);
+      IllegalArgumentException thrownException = assertThrows(IllegalArgumentException.class,
+          () -> testKit.createBlockWithTransactions(message));
+      RawTransaction rawTransaction = TestKit.toRawTransaction(message);
+      String expectedMessage = String.format("Unknown service id (%s) in transaction (%s)",
+          wrongServiceId, rawTransaction);
+      assertTrue(thrownException.getMessage().contains(expectedMessage));
```
dmitry-timofeev: Here and below: assertTrue shall not be used where matchers are available because you won't get expected/actual in case of failure in the output (assertThat ... contains|matches|equals)

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
dmitry-timofeev: Or, better — just stream them directly, with no intermediate copy — I'll push that directly

PR #835, opened 2019-04-11
Add links to EJB modules [ECR-3018]:

dmitry-timofeev:
## Overview

Add links to EJB modules that have dependent modules,
so that the Javadocs for consuming modules (e.g., time-oracle)
have correct links to Javadocs of these modules
(common and core).

This requires a two-step publication of Javadocs (which
we already do to check correctness of other outgoing links).

An alternative of aggregating all Javadocs may be considered
in the future: it works alright, but loses information about
which package belongs to which module.

Websites hosting javadocs of our external dependencies do
not seem to work with javadoc tool (checked guava at javadoc.io
and vertx at github.io).

---
See: 
  - https://jira.bf.local/browse/ECR-3018

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/pom.xml
+++ b/exonum-java-binding/pom.xml
@@ -426,7 +426,29 @@
           <configuration>
             <source>${java.compiler.source}</source>
             <doclint>none</doclint>
+            <!-- Allow fetching links below -->
+            <isOffline>false</isOffline>
             <detectOfflineLinks>false</detectOfflineLinks>
+            <links>
+              <!-- Published Exonum API docs. Only the modules that are dependencies of
+                   other modules are listed here.
+
+                   To generate release Javadocs, a two-step process is required:
+                     0. Bump version.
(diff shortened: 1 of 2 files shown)


exonum-java-binding/pom.xml
```diff
@@ -426,7 +426,29 @@
...
             <source>${java.compiler.source}</source>
             <doclint>none</doclint>
+            <!-- Allow fetching links below -->
+            <isOffline>false</isOffline>
             <detectOfflineLinks>false</detectOfflineLinks>
+            <links>
+              <!-- Published Exonum API docs. Only the modules that are dependencies of
+                   other modules are listed here.
+
+                   To generate release Javadocs, a two-step process is required:
```
MakarovS: Should we add this to release template in wiki?

dmitry-timofeev: Yes, it makes sense.

PR #840, opened 2019-04-15
Always include error message in response [ECR-2992]

dmitry-timofeev:
## Overview

Always include the error message in service response because
it is a QA service that does not have anything to hide.

---
See: https://jira.bf.local/browse/ECR-2992

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/qa-service/src/main/java/com/exonum/binding/qaservice/ApiController.java
+++ b/exonum-java-binding/qa-service/src/main/java/com/exonum/binding/qaservice/ApiController.java
@@ -62,6 +62,7 @@ import com.google.common.io.BaseEncoding;
 import io.vertx.core.Handler;
 import io.vertx.core.MultiMap;
 import io.vertx.core.http.HttpServerRequest;
+import io.vertx.core.http.HttpServerResponse;
 import io.vertx.ext.web.Router;
 import io.vertx.ext.web.RoutingContext;
 import io.vertx.ext.web.handler.BodyHandler;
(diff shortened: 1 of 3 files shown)


exonum-java-binding/qa-service/src/main/java/com/exonum/binding/qaservice/ApiController.java
```diff
@@ -333,15 +335,17 @@ private void failureHandler(RoutingContext rc) {
...
       } else {
-        logger.error("Request error:", requestFailure);
-        rc.response()
-            .setStatusCode(HTTP_INTERNAL_ERROR)
-            .end();
+        logger.error("Internal error", requestFailure);
+        response.setStatusCode(HTTP_INTERNAL_ERROR);
       }
+      String description = Strings.nullToEmpty(requestFailure.getMessage());
+      response.putHeader("Content-Type", "text/plain")
```
bullet-tooth: why do we return plain test for errors? It'll complicate clients because they need to handle json for ok responses and test for errors.

dmitry-timofeev: @dip56 Does it complicate your client? If it doesn't, I would not bother changing that.

dip56: It doesn't matter to me, but I think it is a good idea to return all errors in a single JSON format.

dmitry-timofeev: https://jira.bf.local/browse/ECR-3123

dmitry-timofeev: It is, but as a separate task.

PR #842, opened 2019-04-16
Prevent publishing TestKit module till it is ready [ECR-3118]

MakarovS:
## Overview
Prevent publishing TestKit module to Maven central.

---
See: https://jira.bf.local/browse/ECR-3118

### Definition of Done


exonum-java-binding/testkit/pom.xml
```diff
@@ -69,6 +69,15 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-source-plugin</artifactId>
       </plugin>
+
+      <!-- Skip the deployment of TestKit module as it is not yet ready, fix in ECR-3120 -->
```
dmitry-timofeev: I remember that Ilya had to skip the gpg — haven't you reviewed the patch I referenced?

MakarovS: I did, but thought of it as not very important, doesn't make much difference in my opinion, especially since it's not an internal module, but a temporarily disabled one, for one release cycle or less.

Anyway, added skipping gpg for this and all internal modules as well.

dmitry-timofeev: I am not sure I understand — is it needed or not? If it is then (a) how did we use to get by without this configuration; (b) can the amount of code duplication be reduced?

Or is needed in packaging only and not elsewhere because it does not produce an artifact?

MakarovS: Yes, it is needed in packaging only. But as we don't need to sign other internal modules (as we don't publish those) we can as well skip the signing. If it's not worth cluttering up poms, I can remove these.

dmitry-timofeev: I see now, thanks. I think I'll simply back out the commit

PR #844, opened 2019-04-17
Refactor utils module [ECR-3027]

vitvakatu:
## Overview

Modules moved around, some of them merged, renamed or removed. In general, the list of actions might be seen in commit history.

Some visibility modifiers may be invalid, though I can't see any warnings from the compiler. I wish we will eventually transfer to 2018's module system.

---
See: https://jira.bf.local/browse/ECR-3027

### Definition of Done

- [ ] Public API has Javadoc


exonum-java-binding/core/rust/src/utils/errors.rs
```diff
@@ -125,3 +129,109 @@ pub fn describe_java_exception(env: &JNIEnv, exception: JObject) -> String {
...
+            error!(
+                "Unable to find 'RuntimeException' class: {}",
+                e.description()
+            );
+            return;
+        }
+    };
+    if let Err(e) = env.throw_new(exception, description) {
+        error!(
+            "Unable to find 'RuntimeException' class: {}",
```
dmitry-timofeev: While we are here, I'd fix this incorrect error message to "Failed to throw RuntimeException({}): {}" description, e.description

dmitry-timofeev: To avoid confusion, we might rename the parameter to `message`/`error_message`.

exonum-java-binding/core/rust/src/utils/jni_cache.rs
```diff
@@ -35,6 +41,8 @@ static mut SERVICE_ADAPTER_CONVERT_TRANSACTION: Option<JMethodID> = None;
 static mut JAVA_LANG_ERROR: Option<GlobalRef> = None;
 static mut TRANSACTION_EXECUTION_EXCEPTION: Option<GlobalRef> = None;
 
+/// This function is executed on loading native library by JVM.
+/// It initializes the cache of method and class references.
```
dmitry-timofeev: \+ See: https://docs.oracle.com/en/java/javase/12/docs/specs/jni/invocation.html#jni_onload

exonum-java-binding/core/rust/src/storage/pair_iter.rs
```diff
@@ -19,13 +19,20 @@ use std::mem;
...
+    /// The base `Iterator` instance.
     pub iter: InnerIter,
+    /// Class of the Java representation of the iterator elements.
     pub element_class: GlobalRef,
+    /// Cached method Id of the `element_class` constructor.
     pub constructor_id: JMethodID<'static>,
 }
 
 impl<InnerIter: Iterator> PairIter<InnerIter> {
+    /// Returns new `PairIter` for a given `iter` and `class_name`.
```
dmitry-timofeev: I would add that currently it requires the class to have a constructor
accepting a pair of byte arrays — hence the name.

exonum-java-binding/core/rust/src/runtime/cmd.rs
```diff
@@ -14,14 +14,14 @@
  * limitations under the License.
  */
 
+use super::paths::executable_directory;
 use super::{Config, JvmConfig, RuntimeConfig};
```
skletsun: Nit: Two lines of `super::` imports could be merged into one.

vitvakatu:
> I think that the resource manager used to be in utils because it is an optional dependency of handles-operation code. Shall that code be in a separate module with resource-manager as a sub-module? Or shall rm remain as a sub-module of utils?

Yes, this is a tricky question. Maybe moving the `handle` out of utils would be better than holding `resource manager` inside `utils`. `utils` must be a storage for a small functions that are used in a lot of other modules, but do not have a clear purpose (`error` is a perfect case). So what about moving `handle` and creating new top-level module `handle` with resource manager sub-module? cc @dmitry-timofeev @skletsun

skletsun:
> > I think that the resource manager used to be in utils because it is an optional dependency of handles-operation code. Shall that code be in a separate module with resource-manager as a sub-module? Or shall rm remain as a sub-module of utils?
> 
> Yes, this is a tricky question. Maybe moving the `handle` out of utils would be better than holding `resource manager` inside `utils`. `utils` must be a storage for a small functions that are used in a lot of other modules, but do not have a clear purpose (`error` is a perfect case). So what about moving `handle` and creating new top-level module `handle` with resource manager sub-module? cc @dmitry-timofeev @skletsun

I like the idea of a separate module for `handle` with the `resource manager` as sub-module.

dmitry-timofeev:
Merged as is as there are too many individual refactorings.

PR #850, opened 2019-04-19
Add pool transaction hashes access [ECR-3136]

MakarovS:
## Overview
Add pool transaction hashes access.

---
See: https://jira.bf.local/browse/ECR-3136

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
+++ b/exonum-java-binding/core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
@@ -243,6 +243,8 @@ public final class Blockchain {
 
   /**
    * Returns a set of uncommitted (in-pool) transaction hashes; empty in case of no transactions.
+   *
+   * @see <a href="https://exonum.com/doc/version/0.10/advanced/consensus/specification/#pool-of-unconfirmed-transactions">Pool of Unconfirmed Transactions</a>
    */
   KeySetIndexProxy<HashCode> getPoolTransactionHashes() {
     return schema.getPoolTransactionHashes();
--- a/exonum-java-binding/core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
+++ b/exonum-java-binding/core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
@@ -243,10 +243,12 @@ public final class Blockchain {
 
   /**
    * Returns a set of uncommitted (in-pool) transaction hashes; empty in case of no transactions.
+   * Note that this pool represents the state as of the current snapshot, and its state is volatile
+   * even between block commits.
    *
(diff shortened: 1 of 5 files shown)


exonum-java-binding/core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -239,4 +240,11 @@ public Block getLastBlock() {
   public StoredConfiguration getActualConfiguration() {
     return schema.getActualConfiguration();
   }
+
+  /**
+   * Returns a set of uncommitted (in-pool) transaction hashes.
```
bullet-tooth: What will I get if mempool is empty, empty set or null?

exonum-java-binding/core/src/main/java/com/exonum/binding/blockchain/CoreSchemaProxy.java
```diff
@@ -157,6 +158,14 @@ Block getLastBlock() {
         StandardSerializers.hash(), TRANSACTION_LOCATION_SERIALIZER);
   }
 
+  /**
+   * Returns a set of uncommitted (in-pool) transaction hashes.
+   */
+  KeySetIndexProxy<HashCode> getPoolTransactionHashes() {
```
bullet-tooth: add test to `CoreSchemaProxyIntegrationTest`

exonum-java-binding/core/src/main/java/com/exonum/binding/blockchain/Blockchain.java
```diff
@@ -239,4 +240,13 @@ public Block getLastBlock() {
   public StoredConfiguration getActualConfiguration() {
     return schema.getActualConfiguration();
   }
+
+  /**
+   * Returns a set of uncommitted (in-pool) transaction hashes; empty in case of no transactions.
+   *
+   * @see <a href="https://exonum.com/doc/version/0.10/advanced/consensus/specification/#pool-of-unconfirmed-transactions">Pool of Unconfirmed Transactions</a>
+   */
+  KeySetIndexProxy<HashCode> getPoolTransactionHashes() {
```
dmitry-timofeev: I am not entirely happy with the name, what if it is named `getTransactionPool` and the documentation clarifies that a transaction pool is a set of hashes of uncommitted transaction messages?

dmitry-timofeev: Is that correct that it is intended to be internal (hence package-private and, supposedly, invocation through reflection)? If it is not I would also add a warning that this pool represents the state as of the current snapshot, and its state constantly changes even between block commits.

MakarovS: Yes, I was inclining to make it public

PR #852, opened 2019-04-19
Add java app module & check time oracle is not included [ECR-3114, ECR-3115]

dmitry-timofeev:
### Add a Java app module

Add an internal Java app module that has all dependencies of the native
application (currently, core + time-oracle and their transitives).

Move ServiceRuntimeBootstrap to app module to be able to configure
it using types from any _application_ dependency (i.e., time oracle).

### Move dependency reference classes to app module

Moves the dependency reference classes to app module to be able to
add time oracle as forbidden to be included in the service artifact
(and adds it as such).

### Use JAR dependency type in `packaging`

Specify the dependencies on app as JAR (default) to simplify the configuration.

---
See:
 - https://jira.bf.local/browse/ECR-3114
 - https://jira.bf.local/browse/ECR-3115

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/app/pom.xml
+++ b/exonum-java-binding/app/pom.xml
@@ -1,4 +1,20 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2019 The Exonum Team
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


exonum-java-binding/fakes/pom.xml
```diff
@@ -21,6 +21,12 @@
...
   <dependencies>
+    <dependency>
+      <groupId>com.exonum.binding</groupId>
+      <artifactId>exonum-java-app</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
     <dependency>
       <groupId>com.exonum.binding</groupId>
       <artifactId>exonum-java-binding-core</artifactId>
```
bullet-tooth: Core can be removed such as it comes from the app as transitive?

dmitry-timofeev: Thanks, probably yes, we can rely on app having core.

dmitry-timofeev:
Rebased to keep these changes separate.

PR #857, opened 2019-04-24
Fix build after #844

vitvakatu:
## Overview

---

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/pom.xml
+++ b/exonum-java-binding/core/pom.xml
@@ -25,6 +25,7 @@
     <rust.compiler.version>1.32.0</rust.compiler.version>
     <rust.compiler.features>resource-manager</rust.compiler.features>
     <rust.itSubCrate>integration_tests</rust.itSubCrate>
+    <rust.appSubCrate>exonum-java</rust.appSubCrate>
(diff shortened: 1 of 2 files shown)


dmitry-timofeev:
How did it fail if #844 passed after merge :confused:

vitvakatu:
Yep, that's weird. This module is activate only without resource manager, but we indeed have this testing step (visible in logs)

dmitry-timofeev:
But the parent … wasn't broken: https://travis-ci.com/exonum/exonum-java-binding/builds/109424184

dmitry-timofeev:
What is the issue exactly? Do we have an issue with our build system if some problems are not detected?

vitvakatu:
Investigating, please do not merge. Looks like we're not testing no-resource-manager with `run_all_tests`

dmitry-timofeev:
Might it have something to do with https://github.com/exonum/exonum-java-binding/issues/691 (it only changes Cargo.toml of _app_, not _java-bindings_ library)?

vitvakatu:
@dmitry-timofeev good catch, actually. It took some time, but I found out that this change is indeed the source of the problem.

PR #859, opened 2019-04-25
Add Time Service support for TestKit [ECR-3051]

MakarovS:
## Overview
Add Time Service support for TestKit.

---
See: https://jira.bf.local/browse/ECR-3051

### Definition of Done

- [x] Public API has Javadoc


dmitry-timofeev:
Could it be rebased onto master please?

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/FakeTimeProvider.java
```diff
@@ -16,25 +16,59 @@
...
+
+import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
+import java.time.temporal.TemporalAmount;
 
-// TODO: update Javadocs in P2 [ECR-3051]
+/**
+ * Fake time provider for service testing. Allows to manually manipulate time that is returned
+ * by TestKit time service.
+ */
```
dmitry-timofeev: \+ If you need to set results of different consecutive calls on (link) getTime,
 consider using a mock of TimeProvider instead.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/FakeTimeProvider.java
```diff
@@ -16,25 +16,59 @@
...
 public class FakeTimeProvider implements TimeProvider {
 
-  public FakeTimeProvider create(ZonedDateTime time) {
-    throw new UnsupportedOperationException();
+  private ZonedDateTime time;
```
dmitry-timofeev: I think this class must be thread-safe (more precisely, must ensure that writes made in Java are visible by whatever native thread executes transactions of Time Oracle).

We can either put `volatile` here (with an explanation of _why_ that's enough for our use-case) or make all the operations synchronized (also with an explanation :upside_down_face: )

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/FakeTimeProvider.java
```diff
@@ -16,25 +16,59 @@
...
+  private FakeTimeProvider(ZonedDateTime time) {
+    this.time = time;
   }
 
+  /**
+   * Creates a fake time provider with given time. Note that time should be in UTC time zone.
+   *
+   * @throws IllegalArgumentException if value has time zone other than UTC
+   */
+  public static FakeTimeProvider create(ZonedDateTime time) {
```
dmitry-timofeev: initialTime?

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -222,6 +226,9 @@ private void checkTransactions(List<TransactionMessage> transactionMessages) {
 
   private void checkTransaction(TransactionMessage transactionMessage) {
     short serviceId = transactionMessage.getServiceId();
+    if (serviceId == TIME_SERVICE_ID) {
```
dmitry-timofeev: I would clarify why this check is here

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TimeProvider.java
```diff
@@ -18,7 +18,13 @@
...
 import java.time.ZonedDateTime;
 
-// TODO: update Javadocs in P2 [ECR-3051]
+/**
+ * Time provider for service testing. Used as a time source by TestKit time service.
+ */
 public interface TimeProvider {
+
+  /**
+   * Returns time that is stored in this time provider.
```
dmitry-timofeev:
```suggestion
   * Returns the current time of this time provider in UTC time zone.
```
?

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/UtcZonedDateTimeSerializer.java
```diff
@@ -0,0 +1,69 @@
...
+import static com.google.common.base.Preconditions.checkArgument;
+
+import com.exonum.binding.common.serialization.Serializer;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.time.Instant;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+
+// TODO: duplicate of serializer in time-oracle - move to core? make it public in time-oracle?
```
dmitry-timofeev: I think it is fine to make it public in _time-oracle_ with appropriate documentation (UTC-only, mostly for internal use, etc)

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/FakeTimeProviderTest.java
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  @Test
+  void setTime() {
+    FakeTimeProvider timeProvider = FakeTimeProvider.create(TIME);
+    ZonedDateTime newTime = ZonedDateTime.of(2010, 10, 10, 10, 10, 10, 10, ZoneOffset.UTC);
+    timeProvider.setTime(newTime);
+    assertThat(timeProvider.getTime()).isEqualTo(newTime);
+  }
+
```
dmitry-timofeev: setTimeRejectsInvalidTimeZone?

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -426,6 +445,73 @@ void getAuditorEmulatedNode() {
...
+  @Test
+  void timeServiceWorksInTestKit() {
+    FakeTimeProvider timeProvider = FakeTimeProvider.create(TIME);
+    try (TestKit testKit = TestKit.builder(EmulatedNodeType.VALIDATOR)
+        .withService(TestServiceModule.class)
+        .withTimeService(timeProvider)
+        .build()) {
+      // Commit two blocks for time oracle to prepare consolidated time
+      testKit.createBlock();
+      testKit.createBlock();
```
dmitry-timofeev: But why two blocks (why one isn't enough)? Don't we have a single validator here?

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -426,6 +445,73 @@ void getAuditorEmulatedNode() {
...
+      testKit.withSnapshot((view) -> {
+        TimeSchema timeSchema = TimeSchema.newInstance(view);
+        Optional<ZonedDateTime> consolidatedTime = timeSchema.getTime().toOptional();
+        assertThat(consolidatedTime).contains(TIME);
+
+        // Check that validatorsTimes contains one exactly entry with TestKit emulated node's
+        // public key and time provider's time
+        checkValidatorsTimes(timeSchema, testKit);
+        return null;
```
dmitry-timofeev: I see withSnapshot is often used with Consumers, not Functions (unlike
`Node#withSnapshot`). We can't have them as overloaded parameters if I remember correctly
(there might be ambiguities for the compiler). Can anyone think of a better name?
- `withSnapshot(Function)` vs `verifyWithSnapshot(Consumer)` (we are testkit, aren't we?)
- `apply(Function)` vs `consume(Consumer)` (but won't apply give a false impression we modify
the state?)
- `testkit.snapshot().apply(Function)` vs `testkit.snapshot().consume(Function)` (might be too wordy, but more easily discoverable in testkit api than plain apply/consume).

I'd submit an issue to explore the options.

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -426,6 +445,73 @@ void getAuditorEmulatedNode() {
...
+        TimeSchema timeSchema = TimeSchema.newInstance(view);
+        Optional<ZonedDateTime> consolidatedTime = timeSchema.getTime().toOptional();
+        assertThat(consolidatedTime).contains(newTime);
+        return null;
+      });
+    }
+  }
+
+  @Test
+  void timeServiceWithManyValidatorsDoesNotReturnConsolidatedTime() {
```
dmitry-timofeev: If it is not a bug, it is a very unfortunate peculiarity that, I think, needs to be documented briefly.

@MakarovS , @vitvakatu, Have you managed to identify the cause?

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -426,6 +445,73 @@ void getAuditorEmulatedNode() {
...
+
+        // Check that validatorsTimes contains one exactly entry with TestKit emulated node's
+        // public key and time provider's time
+        checkValidatorsTimes(timeSchema, testKit);
+        return null;
+      });
+    }
+  }
+
+  private void checkValidatorsTimes(TimeSchema timeSchema, TestKit testKit) {
```
dmitry-timofeev: Shan't we pass `time` as an argument, so that it is explicit?

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
MakarovS: Yes, it's because on 4 nodes or more this condition is true - https://github.com/exonum/exonum/blob/master/services/time/src/transactions.rs#L136

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
MakarovS: That's an expected behaviour. After the first block time transactions are generated and after the second one they are processed. @vitvakatu can explain in more detail if needed

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
MakarovS: I'd go with `withSnapshot(Function)` and `verifyWithSnapshot(Consumer)`. I don't think a separate issue is necessary, as there is no need in additional research.

dmitry-timofeev: As it is a separate change in the public API of Testkit and the best option is TBD, I'd add an issue.

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
dmitry-timofeev: I think it must be simply forbidden to create a network with four+ validators in `build` as it makes no sense with the present testkit.

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
dmitry-timofeev: I see, thanks. Please add a comment because it is far from apparent.

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
dmitry-timofeev: … *and* active time-service, obviously. That, however, will make it impossible to have time-service and test a service on 4+ nodes, but that "testing" won't correspond to a real-world situation on 4+ nodes anyway (non-working testkit) :thinking:

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -468,6 +470,9 @@ void timeServiceWorksInTestKit() {
       // Update time in time provider
       ZonedDateTime newTime = TIME.plusDays(1);
       timeProvider.setTime(newTime);
+      // Commit two blocks for time oracle to prepare consolidated time. Two blocks are needed as
```
dmitry-timofeev: to update the time — it is already set by two preceeding transactions.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -81,10 +81,13 @@
  */
 public final class TestKit extends AbstractCloseableNativeProxy {
 
+  // The maximum number of validators supported by TestKit when a time oracle is enabled. The time
```
dmitry-timofeev: It is not a Javadoc.

PR #860, opened 2019-04-25
Ignore flaky service test that fails on Java 12 [ECR-3138].

skletsun:
## Overview
Test is disabled for a while.

---
See: https://jira.bf.local/browse/ECR-3138

### Definition of Done


exonum-java-binding/core/rust/integration_tests/tests/service_runtime_bootstrap.rs
```diff
@@ -23,6 +23,7 @@ use java_bindings::{
 };
 
 #[test]
+#[ignore]
```
dmitry-timofeev: Please add an explanation of why it is ignored and a ticket ref.

PR #879, opened 2019-05-08
Poll events on Snapshot creation in TestKit

vitvakatu:
## Overview

cc @MakarovS 

---

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/rust/src/testkit/mod.rs
+++ b/exonum-java-binding/core/rust/src/testkit/mod.rs
@@ -102,6 +102,24 @@ pub extern "system" fn Java_com_exonum_binding_testkit_TestKit_nativeFreeTestKit
     drop_handle::<TestKit>(&env, handle)
 }
 
+/// Polls events out of event queue. This forces messages handling, e.g. adding transactions
+/// to the pool.
+///
+/// This method is called internally by `createBlock` and `createBlockWithTransactions` methods.
+#[no_mangle]
+pub extern "system" fn Java_com_exonum_binding_testkit_TestKit_nativePollEvents(
+    env: JNIEnv,
+    _: JObject,
+    handle: Handle,
+) {
+    let res = panic::catch_unwind(|| {
+        let testkit = cast_handle::<TestKit>(handle);
+        testkit.poll_events();
+        Ok(())
+    });
+    unwrap_exc_or_default(&env, res)
+}
+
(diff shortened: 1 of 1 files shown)


exonum-java-binding/core/rust/src/testkit/mod.rs
```diff
@@ -102,6 +102,24 @@ pub extern "system" fn Java_com_exonum_binding_testkit_TestKit_nativeFreeTestKit
     drop_handle::<TestKit>(&env, handle)
 }
 
+/// Polls events out of event queue. This forces messages handling, e.g. adding transactions
+/// to the pool.
+///
+/// This method is called internally by `createBlock` and `createBlockWithTransactions` methods.
```
dmitry-timofeev: I don't see it being called. It, being a native Java method, must be suppossed to be called by some Java methods, mustn't it?

@vitvakatu , @MakarovS , 
Also, is it _required_ on the Java side? Is it possible to redefine some existing _native_ testkit methods so that _they_ ensure that these events gets processed when needed? If so, is it always sensible to do in these native methods? Or is it closer to the Java responsibility?

MakarovS: > I don't see it being called. It, being a native Java method, must be suppossed to be called by some Java methods, mustn't it?

I'll add a native method to TestKit after we merge this.

MakarovS: > Also, is it required on the Java side? Is it possible to redefine some existing native testkit methods so that they ensure that these events gets processed when needed? If so, is it always sensible to do in these native methods? Or is it closer to the Java responsibility?

I think we could avoid Java dealing with this and just pull events in `Java_com_exonum_binding_testkit_TestKit_nativeCreateSnapshot`. @vitvakatu what do you think?

dmitry-timofeev: I.e., poll events before creating a snapshot so that the snapshot includes all the submitted transactions?

MakarovS: Yes, exactly

vitvakatu:
Updated, now poll_events is called before every snapshot creation. 🤔 Shouldn't we actually poll events on createBlock methods either?

PR #882, opened 2019-05-08
Check that the loaded native library is compatible [ECR-3148]:

dmitry-timofeev:
## Overview

Check that the loaded native library is compatible, i.e.,
has the same version as the Java library. This check is useful in
_service integration tests_ that can update their dependency
on 'exonum-java-binding-core' but not the application (that
contains the native library), or vice versa.

Requires native code to enable this check: ECR-3172

---
See: https://jira.bf.local/browse/ECR-3148

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
+++ b/exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
@@ -44,8 +44,14 @@ public final class LibraryLoader {
   private static final String JAVA_LIBRARY_PATH_PROPERTY = "java.library.path";
   private static final String DYNAMIC_LIBRARIES_ENV_VAR_WINDOWS = "PATH";
   private static final String DYNAMIC_LIBRARIES_ENV_VAR_UNIX = "LD_LIBRARY_PATH";
-  // TODO: Shall this class read the properties instead?
+
+  /**
+   * The current version of the project. Must be updated on
+   * <a href="https://wiki.bf.local/display/EJB/Java+Binding+Release+Checklist+Template">
+   * each release</a>.
+   */
   private static final String JAVA_BINDING_VERSION = "0.7.0-SNAPSHOT";
+
   // TODO: Remove in ECR-3172
   private static final boolean LIBRARY_VERSION_VERIFICATION_ENABLED = false;
 
@@ -60,9 +66,9 @@ public final class LibraryLoader {
    * Creates a new library loader.
    *
    * @param libraryVersion the version of this library to verify that the native library
-   *     is compatible with it. The compatibility is d
+   *     is compatible with it
    */
-  LibraryLoader(String libraryVersion) {
+  private LibraryLoader(String libraryVersion) {
     this.expectedLibVersion = libraryVersion;
     this.loaded = false;
   }
@@ -77,7 +83,7 @@ public final class LibraryLoader {
     INSTANCE.loadOnce();
   }
 
-  synchronized void loadOnce() {
+  private synchronized void loadOnce() {
     if (loaded) {
       // It has already been attempted to load the library (successfully or not)
       return;
@@ -153,7 +159,7 @@ public final class LibraryLoader {
       return;
     }
     String nativeLibVersion = nativeGetLibraryVersion();
-    if (!nativeLibVersion.equals(expectedLibVersion)) {
+    if (!expectedLibVersion.equals(nativeLibVersion)) {
       String message = String.format(
           "Mismatch between versions of Java library and native '%s' library:%n"
               + "  Java library version:   %s%n"
@@ -167,7 +173,5 @@ public final class LibraryLoader {
     }
   }
 
-  // todo: Shall we extract it as a Supplier<String>, and System.load() as Consumer<String>
-  //    to be able to test?
   private static native String nativeGetLibraryVersion();
 }
--- a/exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
+++ b/exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 The Exonum Team
+ * Copyright 2019 The Exonum Team
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,11 +20,23 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 /**
- * A loader of the native shared library with Exonum framework bindings.
+ * A loader of the native shared library with Exonum framework bindings. It loads the native
+ * library and also verifies that it is compatible with the Java classes. The native library
+ * is compatible iff it has exactly the same version as this Java library. The revision
+ * from which they were built is currently not checked, but may be in the future (see ECR-3173),
+ * because the API between Java and native is considered internal and can be changed
+ * in an incompatible way in any revision.
  *
- * <p>To have library java_bindings available by its name,
- * add a path to the folder containing it to <code>java.library.path</code> property,
- * e.g.: <code>java -Djava.library.path=${EXONUM_HOME}/lib/native …</code>
+ * <p>To enable loading of java_bindings library, add a path to the folder containing it
+ * to <code>java.library.path</code> property, e.g.:
+ * <code>java -Djava.library.path=${EXONUM_HOME}/lib/native …</code>
+ *
+ * <p>This class is thread-safe.
+ *
+ * @see <a href="https://exonum.com/doc/version/0.11/get-started/java-binding/#installation">
+ *   Exonum Java installation instructions</a>
+ * @see <a href="https://exonum.com/doc/version/0.11/get-started/java-binding/#testing">
+ *   Build configuration to enable integration testing of Java services</a>
  */
 public final class LibraryLoader {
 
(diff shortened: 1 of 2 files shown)


exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
```diff
@@ -96,5 +148,26 @@ private static String dynamicLibrariesEnvVar() {
...
+              + "Check that the version of 'exonum-java-binding-core' matches the version of "
+              + "the installed 'Exonum Java' application.%n"
+              + "See https://exonum.com/doc/version/0.11/get-started/java-binding/#installation",
+          BINDING_LIB_NAME, expectedLibVersion, nativeLibVersion);
+      logger.fatal(message);
+      throw new LinkageError(message);
+    }
+  }
+
+  // todo: Shall we extract it as a Supplier<String>, and System.load() as Consumer<String>
```
dmitry-timofeev: We can possibly parameterize this class with this couple of functions (hence the constructor):
```java
private static final LibraryLoader INSTANCE =
    new LibraryLoader(System::loadLibrary, LibraryLoader::nativeGetLibraryVersion,
                      JAVA_BINDING_VERSION);
```

Is this complication worth it? Also, it will still have dependencies on the system properties and environment, unless we also abstract them and parameterize :upside_down_face:

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
```diff
@@ -20,29 +20,81 @@
...
+ * @see <a href="https://exonum.com/doc/version/0.11/get-started/java-binding/#testing">
+ *   Build configuration to enable integration testing of Java services</a>
  */
 public final class LibraryLoader {
 
   private static final String BINDING_LIB_NAME = "java_bindings";
   private static final String JAVA_LIBRARY_PATH_PROPERTY = "java.library.path";
   private static final String DYNAMIC_LIBRARIES_ENV_VAR_WINDOWS = "PATH";
   private static final String DYNAMIC_LIBRARIES_ENV_VAR_UNIX = "LD_LIBRARY_PATH";
+  // TODO: Shall this class read the properties instead?
```
dmitry-timofeev: We could make Maven auto-generate the properties file with the version, but … not sure we need that at this moment (with some code here reading those properties).

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
```diff
@@ -20,29 +20,81 @@
 import org.apache.logging.log4j.Logger;
 
 /**
- * A loader of the native shared library with Exonum framework bindings.
+ * A loader of the native shared library with Exonum framework bindings. It loads the native
+ * library and also verifies that it is compatible with the Java classes. The native library
+ * is compatible iff it has exactly the same version as this Java library. The revision
```
skletsun: Typo: iff -> if

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
```diff
@@ -20,29 +20,81 @@
...
   private static final String DYNAMIC_LIBRARIES_ENV_VAR_WINDOWS = "PATH";
   private static final String DYNAMIC_LIBRARIES_ENV_VAR_UNIX = "LD_LIBRARY_PATH";
+  // TODO: Shall this class read the properties instead?
+  private static final String JAVA_BINDING_VERSION = "0.7.0-SNAPSHOT";
+  // TODO: Remove in ECR-3172
+  private static final boolean LIBRARY_VERSION_VERIFICATION_ENABLED = false;
 
   private static final Logger logger = LogManager.getLogger(LibraryLoader.class);
 
+  private static final LibraryLoader INSTANCE = new LibraryLoader(JAVA_BINDING_VERSION);
```
bullet-tooth: How about to implement the singleton with lazy loading? It doesn't bring a big overhead but anyway.

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
```diff
@@ -20,29 +20,81 @@
...
+   * @throws LinkageError if the native library cannot be loaded; or if it is incompatible
+   *     with this library version
    */
   public static void load() {
-    loadOnce();
+    INSTANCE.loadOnce();
   }
 
-  private static void loadOnce() {
+  synchronized void loadOnce() {
```
bullet-tooth: private?

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
bullet-tooth: I'm not sure about such refactoring for tests because it contains only one `equals` statement.

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
dmitry-timofeev: It isn't: "if and only if"

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
```diff
@@ -20,29 +20,81 @@
...
+  private static final LibraryLoader INSTANCE = new LibraryLoader(JAVA_BINDING_VERSION);
+
+  private final String expectedLibVersion;
+  private boolean loaded;
+
+  /**
+   * Creates a new library loader.
+   *
+   * @param libraryVersion the version of this library to verify that the native library
+   *     is compatible with it. The compatibility is d
```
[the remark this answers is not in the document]
dmitry-timofeev: probably "determined by comparing …" :-D

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
dmitry-timofeev: Yes — update this string each release or make Maven write and this code read the build properties.

skletsun: Would it be possible/better to generate source file with constants before building the "static" source code? (move it to compilation phase instead of reading properties at runtime)

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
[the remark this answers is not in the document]
skletsun: code review gives you great opportunity to discover lots of new words like "iff", "d" 🙃

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
dmitry-timofeev: https://en.oxforddictionaries.com/definition/iff

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
bullet-tooth: It's trivial to pick up maven properties in runtime. So I believe it brings more benefits than drawbacks 🦊 
Also, looks like we can configure RUSTFLAGS in maven and remove `tests_profile`? 🤔

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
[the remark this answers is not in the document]
bullet-tooth: @skletsun OMG YOLO

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
dmitry-timofeev: > It's trivial to pick up maven properties in runtime. So I believe it brings more benefits than drawbacks fox_face

Yes, it is not _difficult_ at all, but it seems like an overkill at this point: this class will gain a responsibility to read (once) a certain file with properties + the build configuration needs to be updated.

dmitry-timofeev: > Also, looks like we can configure RUSTFLAGS in maven and remove tests_profile

How? It invokes rustup and find do determine some various paths to add to RUSTFLAGS.

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
dmitry-timofeev: But it will never be actually "lazy" loaded, if I understand correctly — a class is loaded when its first static symbol is accessed (with some exception for compile-time constants, I believe), and the only publicly accessible symbol is `load`.

https://docs.oracle.com/javase/specs/jls/se11/html/jls-12.html#jls-12.4.1

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
dmitry-timofeev: :ok_man: , let's keep it simple(r) then.

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
bullet-tooth: For example:
```
    <plugins>
      <plugin>
        <artifactId>maven-antrun-plugin</artifactId>
        <version>1.7</version>
        <executions>
          <execution>
            <id>define-property</id>
            <phase>validate</phase>
            <goals>
              <goal>run</goal>
            </goals>
            <configuration>
              <exportAntProperties>true</exportAntProperties>
              <tasks>
                <exec executable="git" outputproperty="testproperty">
                  <arg value="--version"/>
                </exec>
              </tasks>
            </configuration>
          </execution>
          <execution>
            <id>show-property</id>
            <phase>compile</phase>
            <goals>
              <goal>run</goal>
            </goals>
            <configuration>
              <tasks>
                <echo>Displaying value of 'testproperty':</echo>
                <echo>[testproperty] ${testproperty}</echo>
              </tasks>
            </configuration>
          </execution>
        </executions>
      </plugin>
    </plugins>
```
It'll print in the log:
```
     [echo] Displaying value of 'testproperty':
     [echo] [testproperty] git version 2.20.1 (Apple Git-117)
```

dmitry-timofeev: Looks interesting, maybe we can offload that to an Ant script instead of a bash script for ease of use (no source tests_profile) and support of any platform (Windows). Please file an issue if consider that promising.

bullet-tooth: ECR-3184

exonum-java-binding/core/src/main/java/com/exonum/binding/util/LibraryLoader.java
bullet-tooth: Agree, you are right. My doubt was that it'll be loaded in case no integration tests defined.

PR #889, opened 2019-05-15
Light Client: fix blockhein height method

bullet-tooth:
## Overview

---
`$.range.end` has exclusive value and it leads to an unexisting height if request last block in the blockchain.

### Definition of Done

- [ ] Public API has Javadoc


exonum-light-client/src/main/java/com/exonum/client/response/BlocksResponse.java
```diff
@@ -28,12 +28,12 @@
...
-   * The smallest height of the returned blocks that match the search criteria.
+   * The smallest inclusive height of the returned blocks that match the search criteria.
    */
   long blocksRangeStart;
 
   /**
-   * The largest height of the returned blocks that match the search criteria.
+   * The largest exclusive height of the returned blocks that match the search criteria.
    * If some blocks in the range do not match the search criteria then:
    * {@code blocksRangeEnd - blocksRangeStart != blocks.size}.
```
dmitry-timofeev: The blocks API is rather confusing. If they don't match (i.e., skip empty blocks), then what do I get in `end` — `latest + 1` or `maxMatching + 1`? If the first, then why isn't it `lastRequested = end-1`, because it doesn't have a use as `end` as you won't use it in iteration. If the second — then again, why on Earth anyone needs that — just take blocks[-1].height and you are fine (if there is a block, obviously, which, again, indicates that we don't need 'maxMatching + 1').

CC @aleksuss

aleksuss: @dmitry-timofeev ok. I'll fix it.

dmitry-timofeev: Thanks! Which names do you think would be clear on this? firstInRange + latestRequested? firstInRange + lastInRange? first + last?

aleksuss: I suppose `first + last` would be clear enough.

dmitry-timofeev: Yes, with appropriate documentation, warning that `last` is the last requested (= `latest`, if specified, or current blockchain height, if omitted), not necessarily the height of the last block in the returned list, because some (empty) might have been filtered.

exonum-light-client/src/main/java/com/exonum/client/response/BlocksResponse.java
```diff
@@ -28,14 +28,17 @@
...
   /**
-   * The smallest inclusive height of the returned blocks that match the search criteria.
+   * The smallest height of the returned blocks that match the search criteria.
    */
   long blocksRangeStart;
 
   /**
-   * The largest exclusive height of the returned blocks that match the search criteria.
+   * The largest height of the returned blocks that match the search criteria.
+   * The value is equal to {@code heightMax + 1} if heightMax request parameter is passed.
```
dmitry-timofeev: If I pass heightMax, but at that height the block is empty and filtered do I always get `blocks[0].height + 1` or do I get `heightMax + 1`?

dmitry-timofeev: I might have not been clear in the question, but what would happen if

```
Blockchain: 0 - 1 - 2 - 3 - 4 - 5 - <and goes on> - N
                    e       e

Request:  latest=4, count=2, skip_empty

In blocks I would get [B3, B1].

But what do I get for start/end:
 - start=1, end=5 (latest + 1, from request)
 - start=1, end=4 (last matching + 1)
```

---

What if we remove it completely till 0.12? Or, if it is required internally, add a <b>Warning: the value is set mostly randomly — read below if would still like to use.</b>

bullet-tooth: in this case, you will get `blocks[0].heigth + 1` i.e. `end=4` (I believe ).

dmitry-timofeev:
@dip56 Could we please have some tests for that:
1. a `Block` at height `getBlockchainHeight` is accessible immediately after `getBlockchainHeight` request — several iterations must be made to increase the likelihood of catching a potential problem, and
1. (currently failing) a request for blocks at height > `getBlockchainHeight` throws an exception.

PR #900, opened 2019-05-21
Added logging of JVM arguments before JVM start [ECR-3147]

skletsun:
## Overview
Arguments are logged with the `info` level.

---
See: https://jira.bf.local/browse/ECR-3147

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/rust/src/runtime/java_service_runtime.rs
+++ b/exonum-java-binding/core/rust/src/runtime/java_service_runtime.rs
@@ -193,7 +193,7 @@ impl JavaServiceRuntime {
 
         // Log JVM arguments
         let mut jvm_args_line = String::new();
-        for option in args_builder.get_options().iter() {
+        for option in args_builder.options().iter() {
             jvm_args_line.push(' ');
             jvm_args_line.push_str(option);
         }
--- a/exonum-java-binding/core/rust/src/runtime/java_service_runtime.rs
+++ b/exonum-java-binding/core/rust/src/runtime/java_service_runtime.rs
@@ -192,7 +192,7 @@ impl JavaServiceRuntime {
         args_builder = Self::add_user_arguments(args_builder, args_append);
 
         // Log JVM arguments
-        Self::log_jvm_arguments(args_builder);
+        Self::log_jvm_arguments(&args_builder);
 
         args_builder.build()
     }
@@ -250,7 +250,7 @@ impl JavaServiceRuntime {
     }
 
     /// Logs JVM arguments collected by a particular builder.
-    fn log_jvm_arguments(&args_builder: InitArgsBuilder) {
+    fn log_jvm_arguments(args_builder: &InitArgsBuilder) {
         let mut jvm_args_line = String::new();
         for option in args_builder.options().iter() {
             jvm_args_line.push(' ');
--- a/exonum-java-binding/core/rust/src/runtime/java_service_runtime.rs
+++ b/exonum-java-binding/core/rust/src/runtime/java_service_runtime.rs
@@ -191,6 +191,14 @@ impl JavaServiceRuntime {
         // Append extra user arguments
         args_builder = Self::add_user_arguments(args_builder, args_append);
 
+        // Log JVM arguments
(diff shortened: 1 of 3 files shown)


exonum-java-binding/core/rust/src/runtime/java_service_runtime.rs
```diff
@@ -252,6 +251,44 @@ impl JavaServiceRuntime {
...
 
+/// Wrapper for InitArgsBuilder.
+struct JvmArgsBuilder {
+    internal: InitArgsBuilder,
+    jvm_args_line: String,
+}
+
+impl JvmArgsBuilder {
+    /// Create a new JvmArgsBuilder
+    pub(self) fn new() -> Self {
```
vitvakatu: Just wondering: why is it `pub(self)`, not `pub` (as struct is private anyway)?

exonum-java-binding/core/rust/src/runtime/java_service_runtime.rs
```diff
@@ -252,6 +251,44 @@ impl JavaServiceRuntime {
     }
 }
 
+/// Wrapper for InitArgsBuilder.
+struct JvmArgsBuilder {
+    internal: InitArgsBuilder,
+    jvm_args_line: String,
+}
+
+impl JvmArgsBuilder {
```
dmitry-timofeev: Why does this code manually accumulates the options that are available in the builder (and creates a separate structure for that) instead of:
- Adding InitArgsBuilder#toString
- Adding InitArgsBuilder#getOptions?

Probably, InitArgs#toString would be even better, but that structure is rather low-level.

skletsun: Well, I was also thinking about both of these options but:
- InitArgsBuilder#build could fail and toString/Display/Debug would print faulty data in this case.
- InitArgs#toString looks handy but it is really low-level.
- Also, I was hoping to refactor the service runtime code by moving this set of associative functions for arguments processing to new arguments builder which looks more natural but was unable to create more clear code at first attempt.

dmitry-timofeev: > InitArgsBuilder#build could fail and toString/Display/Debug would print faulty data in this case.

Not sure I understand what you mean by that — how could they print faulty data? Also, it seems that the only reason for it to fail is if the passed strings contain 0 byte:

> [from CString::new, used in `#build`] This function will return an error if the supplied bytes contain an internal 0 byte. The NulError returned will contain the bytes as well as the position of the nul byte.

---

From the maintanence perspective, the added builder seems like an overkill, at least in its present form — instead of two simple functions (that might be considered too simple to break) with a clear purpose, we have a structure with unclear purpose, with functions doing several things at once (e.g., building a string repr of the arg line and appending these arguments to the InitArgsBuilder), and no tests. 

If not `InitArgsBuilder#Display/Debug`, would accumulating arguments in a list first, then logging, then adding to the builder be simpler?

exonum-java-binding/core/rust/src/runtime/java_service_runtime.rs
skletsun: It's not used outside module so it's equal to private. Which is ok in case we're not going to put it outside.

exonum-java-binding/core/rust/src/runtime/java_service_runtime.rs
```diff
@@ -191,6 +191,14 @@ impl JavaServiceRuntime {
         // Append extra user arguments
         args_builder = Self::add_user_arguments(args_builder, args_append);
 
+        // Log JVM arguments
+        let mut jvm_args_line = String::new();
```
dmitry-timofeev: I'd suggest to keep this method at the same level of abstraction — i.e., extract this code in a function `log_jvm_arguments(args_builder)` (the steps above have already hidden their implementation details in separate functions).

skletsun: Done

dmitry-timofeev:
build's failing in native

PR #901, opened 2019-05-21
LC: Fix lombok classes

bullet-tooth:
## Overview

---
It's an issue with javadocs generation for classes that use lombok: auto-generated methods (and builders) are missing in javadocs.
See: https://stackoverflow.com/questions/32273060/javadoc-not-generated-for-lombok-getter-and-setter

For fixing this issue the classes must be 'delomboked' first.
By introducing `lombok-maven-plugin` java sources with lombok annotations are placed in `src/main/lombok` (instead of `src/main/java`). During the build process, the `src/main/lombok` code will be 'delomboked' and the generated java code ends up in `target/generated-sources/delombok`.

### Definition of Done

- [ ] Public API has Javadoc


MakarovS:
Does it work now with generated classes excluded?

bullet-tooth:
Yes. Jacoco doesn't check these classes

exonum-light-client/pom.xml
```diff
@@ -291,6 +305,12 @@
             <groupId>org.jacoco</groupId>
             <artifactId>jacoco-maven-plugin</artifactId>
             <version>0.8.3</version>
+            <configuration>
+              <excludes>
+                <!--Exclude auto generated lombok classes-->
+                <exclude>com/exonum/client/response/*</exclude>
```
dmitry-timofeev: Is it the only thing we keep in this package? Why does it fail (and these classes have to be excluded), is there any possible fix?

I think it is fine to exclude them, but I'd add why it cannot be or shall not be fixed.

exonum-light-client/src/main/lombok/com/exonum/client/response/HealthCheckInfo.java
```diff
@@ -22,7 +22,8 @@
 @Value
 public class HealthCheckInfo {
   /**
-   * Shows information about whether it is possible to achieve the consensus between
+   * Consensus status which shows an information about
```
dmitry-timofeev: Could we shorten to "Consensus status shows/indicates whether it it is possible to achieve the consensus between validators in the current network state."?

exonum-light-client/pom.xml
bullet-tooth: There are a few enums in this package as well. If it's not ok - we can specify the exact list of classes.
It fails because `jacoco` is trying to check coverage for lombok generated files and can't find them at `src/main/java`.

dmitry-timofeev: I see, so we exclude the shared package, but the comment says we exclude the auto-generated files (which we do, but also some other things). We can ignore some enums in the package, but if we don't specify the ignored one explicitly we might end up with more unreported classes (and never notice).

> Also, reason to check coverage for auto-generated code.

Lombok-processed classes does not entirely fall into such category (as _entirely_ generated classes). You might put as much code in lombok-annotated classes as you wish, so I don't think such comment is appropriate.

dmitry-timofeev: _Normal_ generated classes are already excluded because they reside in `target/generated-sources` 🙃

dmitry-timofeev:
It finally builds :tada:

PR #904, opened 2019-05-23
Fix default Transaction#info [ECR-3193]:

dmitry-timofeev:
## Overview

Empty string caused panic in the native code,
and, subsequently, failed request processing.
Made #info return an empty object, though, one might
argue it will be more difficult to understand
why `transactions` returns nothing. That, however,
is to be addressed *after* Dynamic Services.

---
See: https://jira.bf.local/browse/ECR-3193

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/CHANGELOG.md
+++ b/exonum-java-binding/CHANGELOG.md
@@ -35,7 +35,7 @@ request. It is modified to return an empty object (no info) by default. (#904)
 
 [tx-info-07]: https://exonum.com/doc/api/java-binding-core/0.7.0/com/exonum/binding/transaction/Transaction.html#info()
 
-## [0.6.0]- 2019-05-08
+## [0.6.0] - 2019-05-08
 
 **If you are upgrading an existing Java service, consult 
 the [migration guide](https://github.com/exonum/exonum-java-binding/blob/ejb/v0.6.0/exonum-java-binding/doc/Migration_guide_0.6.md).**
--- a/exonum-java-binding/CHANGELOG.md
+++ b/exonum-java-binding/CHANGELOG.md
@@ -28,6 +28,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   format instead of the whole message in binary form.
 - RocksDB library is no longer required to be installed on Mac or Linux to run
   the Exonum Java application. (#902)
+  
+### Fixed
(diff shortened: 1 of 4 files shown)


vnermolaev:
Pardon me, but why not return the actual status of the transaction instead of an empty object?
1. this would be aligned with the behaviour exhibited by similar services written in Rust
2. JS light client inherently expects _some status_ in `Transaction.send` method (which polls blockchain until status of the tx changes to `committed`)

For example, the front end for the cryptocurrency Java service uses `Transaction.send`
https://github.com/exonum/exonum-java-binding/blob/5ea07356bf6005d1b7bfe90c83a6539a412d44ae/exonum-java-binding/cryptocurrency-demo/frontend/src/plugins/blockchain.js#L79
This method will fail if the `info` method is not overridden as such

https://github.com/exonum/exonum-java-binding/blob/master/exonum-java-binding/cryptocurrency-demo/src/main/java/com/exonum/binding/cryptocurrency/transactions/CreateWalletTx.java#L83

dmitry-timofeev:
I am not sure I understand — you *will* get everything you see here: https://github.com/exonum/exonum-java-binding/blob/7225ed76b143051a996bab1120de26a15fe2e469/exonum-light-client/src/main/java/com/exonum/client/response/TransactionResponse.java#L29 

in the response, i.e.:
```json
{
  "type": "committed",
  "content": {
    "comment": "That's our default Transaction#info below, everything else is intact:",
    "debug": {}, 
    "message": "41c453a7f45cb0dd73644aa376d3802bb7da4c6797bcf6749211fbcabb5aa8710000800000000a220a20ae740320999e335dd4f5fdc0468f34eb46544aa1995b6cacfcedc82428bd71dd100a1896f7fda8bf8c8af71cfccd80d4c0d5d6f82955cf7c081969282604d3f7e416274a4319484ceea947981b8d337bd170210acd62508f3663acba395bd131456c0b6cd7f09690aec68a05"
  },
  "location": {
    "block_height": 11,
    "position_in_block": 0
  },
  "location_proof": {
    "val": "2f23541b10b258dfc80693ed1bf6ed6f53ccf8908047f7d33e0fec4f29a4a613"
  },
  "status": {
    "type": "success"
  }
}
```

The thing that `Transaction#info` returns is put in an extra field `content.debug`. It isn't even included in `TransactionResponse` in Java light client.

PR #913, opened 2019-05-27
Add TestKit JUnit extension [ECR-3076]

MakarovS:
## Overview
Add TestKit JUnit extension.

---
See: https://jira.bf.local/browse/ECR-3076

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitExtension.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  private final TestKit.Builder testKitBuilder;
+
+  public TestKitExtension(TestKit.Builder testKitBuilder) {
+    this.testKitBuilder = testKitBuilder;
+  }
+
+  @Override
+  public void beforeEach(ExtensionContext extensionContext) {
```
dmitry-timofeev: Why beforeEach is implemented, creating the Testkit even if it is not requested? I think it must be instantiated when requested (= resolved).

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitExtension.java
```diff
@@ -0,0 +1,71 @@
...
+    TestKit testKit = testKitBuilder.build();
+    getStore(extensionContext).put(KEY, testKit);
+  }
+
+  @Override
+  public void afterEach(ExtensionContext extensionContext) {
+    TestKit testKit = getStore(extensionContext).get(KEY, TestKit.class);
+    testKit.disposeInternal();
```
dmitry-timofeev: Why do we need this if the store will destroy the objects put there, if they implement https://junit.org/junit5/docs/current/api/org/junit/jupiter/api/extension/ExtensionContext.Store.CloseableResource.html?

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitExtension.java
```diff
@@ -0,0 +1,71 @@
...
+  @Override
+  public Object resolveParameter(
+      ParameterContext parameterContext, ExtensionContext extensionContext)
+      throws ParameterResolutionException {
+    return getStore(extensionContext).get(KEY, TestKit.class);
+  }
+
+  private ExtensionContext.Store getStore(ExtensionContext context) {
+    return context.getStore(ExtensionContext.Namespace.create(getClass(),
+        context.getRequiredTestMethod()));
```
dmitry-timofeev: Is it really needed to include the test method as the namespace? Can't we just create a single namespace as a constant? I think stores of, for instance, test methods running in parallel isolated, and `TempDirectory` in JUnit 5 suggests they are.

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -68,12 +74,9 @@
   }
 
   @Test
-  void createTestKitForSingleService() {
-    TestService service;
-    try (TestKit testKit = TestKit.forService(TestServiceModule.class)) {
-      service = testKit.getService(TestService.SERVICE_ID, TestService.class);
-      checkTestServiceInitialization(testKit, service);
-    }
+  void createTestKitForSingleService(TestKit testKit) {
```
dmitry-timofeev: It is nice if we can inject Testkit directly, but the present implementation makes it impossible to configure the testkit differently for different tests (without `@Nested` or separating the tests into multiple files).

MakarovS: Yes, I thought about that, but is it a frequent case for users to test their service with different TestKit configuration? It seems that usually they'd use the same configuration in the same test suite or use `@Nested`. Even if they need that, they could just not inject it and create it manually. If not, what do you propose?

dmitry-timofeev: auditor/validator? 

> If not, what do you propose?

Please see in jira.

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/ValidatorCount.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Changes network validator count for injected TestKit.
+ */
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface ValidatorCount {
+
+  /**
+   * Validator count of TestKit network.
```
dmitry-timofeev:
```
Number of validator nodes in the Testkit network. Must be positive.

@see Builder#<apropriate method>
```

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/Validator.java
```diff
@@ -0,0 +1,30 @@
...
+
+package com.exonum.binding.testkit;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Changes main TestKit node type to validator for injected TestKit.
```
dmitry-timofeev: `@see <appropriate enum/Builder method>`?

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitExtension.java
```diff
@@ -61,11 +58,87 @@ public boolean supportsParameter(ParameterContext parameterContext,
...
+      }
+      testKit = testKitBuilder.build();
+      CloseableTestKit newCloseableTestKit = new CloseableTestKit(testKit);
+      getStore(extensionContext).put(KEY, newCloseableTestKit);
+    } else {
+      // Throw an exception if TestKit was already instantiated in this context, but user tries to
+      // reconfigure it
+      if (annotationsUsed) {
+        throw new RuntimeException("TestKit was parameterized with annotations after being"
+            + " instantiated in " + extensionContext.getDisplayName());
```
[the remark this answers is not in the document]
MakarovS: Sorry for the ambiguity - I intended this message to mean "parameterized in `test`", not "instantiated in `test`".
The thing is that we can inject either in `@BeforeEach`, `@Test`, or `@AfterEach`. All of them have the same `ExtensionContext`, so it's not clear what to put as an `origin`.
There are two possible options -

- if a user reconfigures TestKit in `@Test`, then he will get a clear `displayName`
- If he tries to reconfigure in `@AfterEach` then he'll get `displayNames` of all tests he has.

We could improve the exception message though and make clear what could be incorrect.

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitExtension.java
```diff
@@ -61,11 +58,87 @@ public boolean supportsParameter(ParameterContext parameterContext,
...
+    }
+    return testKit;
+  }
+
+  /**
+   * Check the extension context and throw if TestKit is injected in @BeforeAll or @AfterAll.
+   */
+  private void checkExtensionContext(ExtensionContext extensionContext) {
+    Optional<Method> testMethod = extensionContext.getTestMethod();
+    testMethod.orElseThrow(() ->
```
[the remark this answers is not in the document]
MakarovS: > I don't see that the actual condition corresponds to the intended one.

What do you mean? For `@BeforeAll` and `@AfterAll` it will be empty, for `@BeforeEach`, `@Test`, `@AfterEach` it will contain a test method.

> Also, unless the extension is in static field, it won't be able to inject in @BeforeAll — so we may consider dropping this limitation, just documenting the consequences in the extension documentation.

Unless `@TestInstance(Lifecycle.PER_CLASS)` is used, then user is able to declare `@BeforeAll` and `@AfterAll` as non-static.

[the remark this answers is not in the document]
dmitry-timofeev: I meant that the code said `getTestMethod` but the condition referred to `BeforeAll`/`AfterAll`, and this connection was not obvious from the code, as the fact that `getTestMethod` returned non-empty thing in case of `BeforeEach`/`AfterEach`.

I suggest clarifying that concisely.

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitExtension.java
```diff
@@ -61,11 +58,87 @@ public boolean supportsParameter(ParameterContext parameterContext,
...
+        new RuntimeException("TestKit can't be injected in @BeforeAll or @AfterAll"));
```
dmitry-timofeev: On the question of whether we shall or shall not forbid it:

"If you do need the same Testkit instance for all tests in the class, just use the builder directly:

private static final Testkit testkit = Testkit.newBuilder()
    .withService(YourServiceModule.class)
    .build();
"

… but as I wrote this instruction I remembered that the user will also need to write `@AfterAll` that *destroys* the testkit! So maybe we shall support injection in `BeforeAll` after all, shan't we?

MakarovS: Yes, but isn't the idea of `@Test` cases is that they are independent? Why would a user want to use the same TestKit? It's usually used to create some blocks and then review the state of the system, so using the same TestKit would lead to errors.

But we could document that and it would be easier for us - no need to throw an exception in `@Before/AfterAll` injections.

dmitry-timofeev: So the questions are:
1. Could anyone (@bullet-tooth) think of a realistic use-case when one needs a shared testkit for all methods?
2. Shall we restrict such usage (as error-prone) if we can't think of such use-cases?

I can think of a scenario when setup is performed in `@BeforeAll` and all tests perform only read-only operations on testkit — but how realistic that is, and how resistant to future additions of new tests performing write operations? Also, this scenario would also work with `@BeforeEach`, somewhat slower, but way more reliable.

MakarovS: Also I'd prefer to explicitly prohibit injecting in `@AfterAll`, as it doesn't really make sense (it'll just create one more TestKit if injected). Although I agree that users might want to perform only read-only operations in different tests, it seems like a very error-prone approach. I think `@BeforeEach` is more reliable indeed even if slow.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/Auditor.java
```diff
@@ -23,6 +23,9 @@
 
 /**
  * Changes main TestKit node type to auditor for injected TestKit.
+ *
+ * @see TestKit.Builder#withNodeType(EmulatedNodeType)
```
dmitry-timofeev: Is second `@see` missing? Because the method won't show.

exonum-java-binding/testkit/pom.xml
```diff
@@ -33,11 +33,24 @@
       <version>${project.version}</version>
     </dependency>
 
+    <dependency>
+      <groupId>junit</groupId>
```
dmitry-timofeev: It is JUnit 4.

exonum-java-binding/testkit/pom.xml
```diff
@@ -33,11 +33,24 @@
...
     <dependency>
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>org.junit.platform</groupId>
+      <artifactId>junit-platform-testkit</artifactId>
+      <version>${junit-platform-testkit.version}</version>
```
dmitry-timofeev: test-scoped?

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKitExtension.java
```diff
@@ -29,22 +29,88 @@
...
+ * void test(TestKit testKit) {
+ *   // Test logic
+ * }
+ * </code></pre>
+ *
+ * instead of:
+ *
+ * <pre><code>
+ * static {
+ *   LibraryLoader.load();
```
dmitry-timofeev: It is no longer needed (once the pr lands)

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKitExtension.java
```diff
@@ -29,22 +29,88 @@
...
 import org.junit.jupiter.api.extension.ParameterResolver;
 
+/**
+ * Extension that injects TestKit into service tests. Register this extension with TestKit builder
+ * and a TestKit will be injected as a parameter, instantiated from given builder.
+ *
+ * Allows injecting TestKit into test context:
+ *
+ * <pre><code>
+ * @RegisterExtension
```
dmitry-timofeev: @ must be escaped

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKitExtension.java
```diff
@@ -29,22 +29,88 @@
...
+ *
+ * <pre><code>
+ * static {
+ *   LibraryLoader.load();
+ * }
+ *
+ * @Test
+ * void test() {
+ *   try (TestKit testKit = TestKit.forService(TestServiceModule.class)) {
+ *     // Set up logic
```
dmitry-timofeev: That is not an apples-to-apples comparison. If we include @BeforeEach in the first,
than there is absolutely no problem including @BeforeEach in the second example. With @AfterEach, necessarily, to destroy it :upside_down_face:

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKitExtension.java
```diff
@@ -29,22 +29,88 @@
...
+ *     // Test logic
+ *   }
+ * }
+ * </code></pre>
+ *
+ * <p>As different tests might need slightly different TestKit configuration, following
+ * parameterization annotations are available:
+ * <ul>
+ *   <li>{@link Validator} sets main TestKit validator node type to validator</li>
```
dmitry-timofeev: Do we need validator? Won't you usually use validator by default, and configure Auditors if needed? Or keep it for completeness?

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKitExtension.java
```diff
@@ -29,22 +29,88 @@
...
+ * These annotations should be applied on TestKit parameter:
+ *
+ * <pre><code>
+ * @RegisterExtension
+ * static TestKitExtension testKitExtension = new TestKitExtension(
+ *     TestKit.builder()
+ *         .withService(TestServiceModule.class));
+ *
+ * @Test
+ * void test(@Auditor @ValidatorCount(validatorCount = 8) TestKit testKit) {
```
dmitry-timofeev: I suggest using value (IIRC) for this parameter. This way one could write `@ValidatorCount(8)`

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitExtensionTest.java
```diff
@@ -0,0 +1,205 @@
...
+      // Shouldn't be executed
+    }
+
+    @Test
+    void testTestKitReconfiguration(TestKit testKit) {
+      // This test should pass
+    }
+  }
+
+  @TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
dmitry-timofeev: Is it needed if the extension is already in `static` field?

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitParameterizationTest.java
```diff
@@ -0,0 +1,59 @@
...
+class TestKitParameterizationTest {
+
+  @RegisterExtension
+  static TestKitExtension testKitExtension = new TestKitExtension(
+      TestKit.builder()
+          .withService(TestServiceModule.class));
+
+  @Test
+  void testDefaultTestKit(TestKit testKit) {
+    // Check that main TestKit node is an auditor
```
dmitry-timofeev: 'validator`.
     Also, as this test relies on that and the num of validators, please set that explicitly
     on the builder (as TEMPLATE_NODE_TYPE, TEMPLATE_NUM_VALIDATORS).

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitParameterizationTest.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  @Test
+  void testTestKitAuditor(@Auditor TestKit testKit) {
+    // Check that main TestKit node is an auditor
+    assertThat(testKit.getEmulatedNode().getValidatorId()).isEmpty();
+  }
+
+  @Test
+  void testTestKitValidatorCount(@ValidatorCount(validatorCount = 8) TestKit testKit) {
```
dmitry-timofeev: What if 

```java

  @Test
  void testDefaultTestKit(TestKit testKit) {
    // ...
    testKit.withSnapshot(verifyNumValidators(DEFAULT_NUM_VALIDATORS));
  }

  @Test
  void testTestKitValidatorCount(@ValidatorCount(validatorCount = 8) TestKit testKit) {
    testKit.withSnapshot(verifyNumValidators(8));
  }

  private static Function<Snapshot, Void> verifyNumValidators(int expected) {
    return (view) -> {
      Blockchain blockchain = Blockchain.newInstance(view);
      assertThat(blockchain.getActualConfiguration().validatorKeys().size())
          .isEqualTo(expected);
      return null;
    };
  }
```

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -54,14 +54,18 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.TestInstance.Lifecycle;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
+@TestInstance(Lifecycle.PER_CLASS)
```
dmitry-timofeev: Why? It does not seem to be needed.

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -54,14 +54,18 @@
...
 class TestKitTest {
 
   @RegisterExtension
-  TestKitExtension testKitExtension = new TestKitExtension(
-      TestKit.builder(EmulatedNodeType.VALIDATOR)
+  static TestKitExtension testKitExtension = new TestKitExtension(
```
dmitry-timofeev: Shall we use the extension in the test of testkit + its builder?

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -81,7 +85,7 @@ void createTestKitForSingleService(TestKit testKit) {
 
   @Test
   void createTestKitWithBuilderForSingleService() {
```
dmitry-timofeev: It is probably no longer needed, isn't it?

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -199,7 +203,8 @@ void createTestKitWithSeveralValidators() {
   @Test
   void createTestKitWithAuditorAndAdditionalValidators() {
     short validatorCount = 2;
-    try (TestKit testKit = TestKit.builder(EmulatedNodeType.AUDITOR)
+    try (TestKit testKit = TestKit.builder()
+        .withNodeType(EmulatedNodeType.AUDITOR)
```
dmitry-timofeev: Shall we use `with` prefix, or maybe drop it (e.g., `nodeType`) or add usual `set` and `add` depending on the operation? On the other hand, if used consistently, it slightly helps with discoverability (if **all** use `with`, you can type `wi` and see what it could do).

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/Validator.java
```diff
@@ -23,6 +23,9 @@
 
 /**
  * Changes main TestKit node type to validator for injected TestKit.
+ *
+ * @see TestKit.Builder#withNodeType(EmulatedNodeType)
```
dmitry-timofeev: Same as above (two `@sees`?)

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitExtension.java
dmitry-timofeev: > it'll just create one more TestKit if injected

It won't — contexts are nested, everything in the parent context (`@BeforeAll`) is accessible in children (`@Beach`, `@Test`).

dmitry-timofeev: Ok, let's forbid but let's put something along the lines of

```
Testkit cannot be injected in @BeforeAll/AfterAll because it is a stateful, mutable, object and sharing it between all tests is error-prone. Consider injecting it in @BeforeEach instead.

If you *do* need the same instance for all tests — just use `Testkit#builder` directly. Don't forget to destroy it in @AfterEach.
```

MakarovS: > It won't — contexts are nested, everything in the parent context (@BeforeAll) is accessible in children (@Beach, @Test).

I was talking about `@AfterAll`. If we inject in `@BeforeAll`, we'd have same TestKit in all tests - but if we inject in `@AfterAll`, we won't - every Test will have its own instance, including `@AfterAll` method.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKitExtension.java
MakarovS: I think we should have both. It's possible that a user needs most of his tests to work with an auditor TestKit, but a couple to have a validator main node.

We could also go with a single annotation `@NodeType(EmulatedNodeType.VALIDATOR/AUDITOR)`.

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
MakarovS: I like having same prefix for the reason you stated - it's nice to have all methods starting with `with`

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitExtensionTest.java
MakarovS: Yes, still needed

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
MakarovS: I think it's okay to use extension for most of the TestKit tests (not builder tests though).

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
MakarovS: But why? I think it's still a viable test.

exonum-java-binding/cryptocurrency-demo/src/test/java/com/exonum/binding/cryptocurrency/transactions/CreateWalletTxTest.java
```diff
@@ -67,51 +74,47 @@ void constructorRejectsNegativeBalance() {
 
   @Test
   @RequiresNativeLibrary
```
bullet-tooth: I'm not sure that this annotation is still needed to be applied with tests which use `TestKit`. Of course, it adds a description of the test. From another hand, `TestKit` doc says that it initializes Exonum blockchain instance (which requires the native library to be loaded). So I propose to apply `@RequiresNativeLibrary` at `TestKitExtension` and remove the annotation in tests. 
@dmitry-timofeev @MakarovS  WDYT?

exonum-java-binding/cryptocurrency-demo/src/test/java/com/exonum/binding/cryptocurrency/transactions/TransferTxHistoryTest.java
```diff
@@ -32,60 +32,65 @@
...
+  @RegisterExtension
+  TestKitExtension testKitExtension = new TestKitExtension(
+      TestKit.builder()
+          .withService(CryptocurrencyServiceModule.class));
+
   private static final KeyPair ACCOUNT_1 = PredefinedOwnerKeys.FIRST_OWNER_KEY_PAIR;
   private static final KeyPair ACCOUNT_2 = PredefinedOwnerKeys.SECOND_OWNER_KEY_PAIR;
 
   @Test
   @RequiresNativeLibrary
```
bullet-tooth: The same is here. The class and tests are annotated.

exonum-java-binding/pom.xml
```diff
@@ -114,6 +114,7 @@
     <hamcrest.version>2.1</hamcrest.version>
     <junit.version>4.12</junit.version>
     <junit.jupiter.version>5.4.2</junit.jupiter.version>
+    <junit-platform-testkit.version>1.5.0-M1</junit-platform-testkit.version>
```
bullet-tooth: Please move it to the `TestKit` pom (or is it needed for other modules?)

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -78,7 +77,7 @@
  * created.
  *
  * @see <a href="https://exonum.com/doc/version/0.11/get-started/test-service/">TestKit documentation</a>
- *      <a href="https://exonum.com/doc/version/0.11/advanced/consensus/specification/#pool-of-unconfirmed-transactions">Pool of Unconfirmed Transactions</a>
```
bullet-tooth: Please fix `@linkplain` above ⏫ (redundant `}` )

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKitExtension.java
```diff
@@ -0,0 +1,207 @@
...
+import org.junit.jupiter.api.extension.ParameterResolutionException;
+import org.junit.jupiter.api.extension.ParameterResolver;
+
+/**
+ * Extension that injects TestKit into service tests and destroys afterwards. Register this
+ * extension with TestKit builder and a TestKit will be injected as a parameter, instantiated from
+ * given builder. Example usage:
+ *
+ * <pre><code>
+ * &#64;RegisterExtension
```
bullet-tooth: I see neither examples of creating `TestKit` in a declarative way (using `@ExtendWith`) or a Javadoc describing why it's not possible/forbidden.

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitExtension.java
dmitry-timofeev: I see, in this case — yes, but it makes no apparent sense to use that :-)

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
dmitry-timofeev: Sorry, wrong method — the question is for createTestKitForSingleService^.

exonum-java-binding/cryptocurrency-demo/src/test/java/com/exonum/binding/cryptocurrency/transactions/CreateWalletTxTest.java
dmitry-timofeev: `@RequiresNativeLibrary` is our internal annotation for tagging (and, possibly, subsequent filtering — we used to use it to skip such tests on Linux). It has no other effects. 

I don't think the testkit extension (which is suppossed to be used by other) shall bring our tags for client tests (= auto-tag them).

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitExtensionTest.java
dmitry-timofeev: But why?

> Static Fields
If a @RegisterExtension field is static, the extension will be registered after extensions that are registered at the class level via `@ExtendWith`. Such static extensions are not limited in which extension APIs they can implement. **Extensions registered via static fields may therefore implement class-level and instance-level extension APIs such as BeforeAllCallback, AfterAllCallback,** and TestInstancePostProcessor as well as method-level extension APIs such as BeforeEachCallback, etc.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TimeProvider.java
```diff
@@ -27,4 +27,9 @@
    * Returns the current time of this time provider in UTC time zone.
    */
   ZonedDateTime getTime();
+
+  /**
+   * Returns a copy of this time provider.
```
dmitry-timofeev: I do not understand how it is supposed to work. It is apparently required when testkit is
`static` **and** it has a time provider set. In this case the time provider needs to be pre-configured
in a static context (= before `@BeforeAll`). 

Some issues with the current design:
1. Does not make sense with FakeTimeProvider unless it is constant (= never changes the current time),
because the client cannot access the actual time provider used by the testkit. We can probably
make testkit return it, but I think it will only complicate things (pass one provider to the 
builder, but get another one from the teskit). On the other hand, if we do not copy (in a `static` context),
and the client puts it in a `static` field and then reconfigures in tests —
they will end up with broken initial state in case of single-threaded execution of tests or concurrency
issues (and anything broken) in case of multi-threaded execution.
2. Cannot use mocks even without the extension because they cannot be cloned (well, can with a wrapper with a throwing copy implementation).

Unless I'm missing something, copying seems restrictive and not that easy to use or fix.
But it is required only for `static` extensions. What we could do:
1. 'Fix' by adding `Testkit#getTimeProvider` and some obscure documentation.
2. Ignore the concurrency issue (i.e., `static` extension with a shared template builder 
and a shared time provider will use the same instance in all tests). If a client breaks
it by modifying the provider in any test, it is their problem.
3. Forbid `static` extensions — we seem to have found a good reason :-)
4. Anything else?

exonum-java-binding/integration-tests/src/test/java/com/exonum/binding/test/BlockchainIntegrationTest.java
```diff
@@ -343,21 +337,21 @@ void getActualConfiguration() {
...
+    testKitTest(testKit, (blockchain) -> {
       KeySetIndexProxy<HashCode> transactionPool = blockchain.getTransactionPool();
       assertThat(transactionPool.contains(message.hash()))
           .describedAs("pool=%s", transactionPool)
           .isTrue();
     });
   }
 
-  private void testKitTest(Consumer<Blockchain> test) {
+  private void testKitTest(TestKit testKit, Consumer<Blockchain> test) {
```
dmitry-timofeev: Wouldn't it be better to keep testkit as a field if it’s needed to pass it between methods?

exonum-java-binding/integration-tests/src/test/java/com/exonum/binding/test/TimeSchemaProxyIntegrationTest.java
```diff
@@ -88,15 +78,15 @@ void getValidatorsTime() {
...
+  private void setUpConsolidatedTime(TestKit testKit) {
     // Commit two blocks for time oracle to update consolidated time. Two blocks are needed as
     // after the first block time transactions are generated and after the second one they are
     // processed
     testKit.createBlock();
     testKit.createBlock();
   }
 
-  private void testKitTest(Consumer<TimeSchema> test) {
+  private void testKitTest(TestKit testKit, Consumer<TimeSchema> test) {
```
dmitry-timofeev: Same here — it is used by every method in a test class.

exonum-java-binding/testkit/pom.xml
```diff
@@ -33,11 +33,25 @@
       <version>${project.version}</version>
     </dependency>
 
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
```
dmitry-timofeev: I think an extension must not depend on the aggregate thing, but on `junit-jupiter-api`.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKitExtension.java
```diff
@@ -0,0 +1,207 @@
...
+    return testKit;
+  }
+
+  /**
+   * Check the extension context and throw if TestKit is injected in @BeforeAll or @AfterAll.
+   */
+  private void checkExtensionContext(ExtensionContext extensionContext) {
+    Optional<Method> testMethod = extensionContext.getTestMethod();
+    testMethod.orElseThrow(() ->
+        new RuntimeException("TestKit can't be injected in @BeforeAll or @AfterAll because it is"
```
dmitry-timofeev: RuntimeException

exonum-java-binding/integration-tests/src/test/java/com/exonum/binding/test/BlockchainIntegrationTest.java
MakarovS: I agree, thought that it was better the way it was

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TimeProvider.java
MakarovS: I'd forbid (discourage?) `static` extensions. It's easier for us, will prevent any other possible issues with having a static extensions (I can't think of any right now though) and won't limit user in any way.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKitExtension.java
MakarovS: It's not possible to use `@ExtendWith` because that way we can't pass a TestKit builder as an argument to extension constructor. Should I specify that in Javadoc?

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TimeProvider.java
MakarovS: Also I suppose it's better to make a copy of `templateTestKitBuilder` on every parameter resolve anyway, even though we discourage having a `static` extension. Even if someone uses `static`, they are unlikely to have any issues in most cases.

dmitry-timofeev: OK, I agree with discouraging it because apparently there is no reasonable way to support `static`. Shall we also "forbid" (probably, by supporting class-level callbacks, and, if they get invoked — throwing an Exception)? Are there any downsides in doing that (i.e., is it possible that a user is forced to use `static` but it won't cause any problems)?

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKitExtension.java
dmitry-timofeev: Since we provide examples of usage right in the Javaddocs (and, later, on the web), I don't think it is necessary to say that "You can't register it with ExtendWith". If users try and it does not work — we can reasonably assume they will get to the (any) docs.

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitExtensionTest.java
dmitry-timofeev: ❓
On top of that — the extension is no longer static.

dmitry-timofeev: ❓
On top of that — the extension is no longer static.

MakarovS: Sorry, missed your previous question. If we had both an extension and `@BeforeAll` method `static`, then it's all good - we don't need an annotation. We had `@BeforeAll` not static though when you originally commented on that, that's why we had `@TestInstance(TestInstance.Lifecycle.PER_CLASS)` instead.

Now as extension is not `static`, we need this annotation, otherwise we get `"@BeforeAll method <method> must be static unless the test class is annotated with @TestInstance(Lifecycle.PER_CLASS)."`.

bullet-tooth:
I see old way usage of TestKit in `com.exonum.binding.test.BlockchainIntegrationTest`. Or is it in the scope of another task?

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitExtensionTest.java
dmitry-timofeev: Ok, let's keep them non-static.

dmitry-timofeev:
> I see old way usage of TestKit in com.exonum.binding.test.BlockchainIntegrationTest. Or is it in the scope of another task?

It makes little sense to use the extension when you need to inject it in `@BeforeEach` and then use in test and utility methods. It is simpler to put it in a field, to avoid passing explicitly between methods.
Testkit, however, would automatically destroy it (no need for `@AfterEach`), but the extension overhead (in terms of extra lines of code to register it and some complexity) seems higher.

PR #915, opened 2019-05-27
Remove explicit find_class:

dmitry-timofeev:
## Overview

- Remove explicit find_class
- Cache the RuntimeException class
- Catch possible panics in JNI_OnLoad, returning an invalid
version to communicate initialization failure

---

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/rust/src/utils/errors.rs
+++ b/exonum-java-binding/core/rust/src/utils/errors.rs
@@ -14,14 +14,14 @@
  * limitations under the License.
  */
 
+use jni::objects::JObject;
+use jni::JNIEnv;
+
 use std::any::Any;
 use std::cell::Cell;
 use std::error::Error;
 use std::result;
 use std::thread;
-
-use jni::objects::JObject;
-use jni::JNIEnv;
 
 use utils::{get_class_name, get_exception_message, jni_cache::classes_refs};
 use {JniError, JniErrorKind, JniResult};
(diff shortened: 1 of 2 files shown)


exonum-java-binding/core/rust/src/utils/errors.rs
```diff
@@ -14,15 +14,15 @@
...
-use jni::objects::JObject;
-use jni::JNIEnv;
-
 use std::any::Any;
 use std::cell::Cell;
 use std::error::Error;
 use std::result;
 use std::thread;
 
+use jni::objects::JObject;
```
vitvakatu: fyi, we followed https://github.com/exonum/exonum/blob/master/CONTRIBUTING.md in terms of import items order. Though I don't think it's a problem, some kind of uniformity may be useful

exonum-java-binding/core/rust/src/utils/jni_cache.rs
```diff
@@ -27,6 +27,10 @@ use jni::{
 };
 use parking_lot::{Once, ONCE_INIT};
 use std::os::raw::c_void;
+use std::panic::catch_unwind;
```
skletsun: I'd suggest making it nested like the rest of the imports.

dmitry-timofeev: :ok_hand: , but *why* do we use this syntax? It seems to be harder in case of merges — instead of a single line per imports rustfmt may compress several imports into a single line, if a single import disappears and expands than multiple lines would be affected in the diff.

PR #917, opened 2019-05-30
Change Dumb executor to TLS-based one [ECR-3204]

vitvakatu:
## Overview

---
See: https://jira.bf.local/browse/ECR-3204

### Definition of Done

- [ ] Public API has Javadoc


exonum-java-binding/core/rust/src/proxy/executor.rs
```diff
@@ -0,0 +1,68 @@
...
+impl Executor {
+    /// Creates new Executor with specified JVM.
+    pub fn new(vm: Arc<JavaVM>) -> Self {
+        Self {
+            vm,
+        }
+    }
+
+    /// Executes a provided closure, making sure that the current thread
+    /// is attached to the JVM. Additionally ensures that local object references freed after call.
```
dmitry-timofeev: are freed after call?

dmitry-timofeev: also below

exonum-java-binding/core/rust/integration_tests/tests/executor.rs
```diff
@@ -18,37 +18,37 @@ extern crate java_bindings;
...
 lazy_static! {
     pub static ref VM: Arc<JavaVM> = create_vm_for_tests();
-    pub static ref EXECUTOR: DumbExecutor = DumbExecutor::new(VM.clone());
+    pub static ref EXECUTOR: Executor = Executor::new(VM.clone());
 }
 
 #[test]
 fn single_thread() {
-    test_single_thread(&*EXECUTOR);
+    test_single_thread(EXECUTOR.clone());
```
dmitry-timofeev: Probably not single with the guys below in the same file :upside_down_face: 

I think we can leave it as is, but please inspect these tests to see if they need to be updated or extended.

exonum-java-binding/core/rust/src/proxy/executor.rs
```diff
@@ -0,0 +1,68 @@
...
+// limitations under the License.
+
+use jni::objects::JObject;
+use jni::{JNIEnv, JavaVM};
+
+use std::sync::Arc;
+
+use JniResult;
+
+/// The capacity of local frames, allocated for attached threads
```
dmitry-timofeev: … by default. Same as the default value Hotspot uses when calling native Java methods.

exonum-java-binding/core/rust/src/proxy/executor.rs
```diff
@@ -0,0 +1,68 @@
...
+const LOCAL_FRAME_CAPACITY: i32 = 32;
```
dmitry-timofeev: DEFAULT_…?

exonum-java-binding/core/rust/integration_tests/tests/executor.rs
dmitry-timofeev: :question:

exonum-java-binding/core/rust/integration_tests/tests/executor.rs
```diff
@@ -47,8 +47,3 @@ fn concurrent_threads() {
     const THREAD_NUM: usize = 8;
     test_concurrent_threads(EXECUTOR.clone(), THREAD_NUM)
 }
-
-#[test]
-fn nested_attach() {
```
dmitry-timofeev: I think we shall keep it because it verifies the executor implementation allowing recursive `with_attached`, which some implementations might not permit.

exonum-java-binding/core/rust/integration_tests/tests/executor.rs
dmitry-timofeev: At least `nested_attach` cannot reliably live in the same source file.

vitvakatu: Actually it can, but it will be unable to check if the thread is detached correctly after the outer-most scope is finished. Other tests a totally fine to live together.

PR #947, opened 2019-06-05
Rename InternalServerError to InvalidTransactionException [ECR-3258]

dmitry-timofeev:
## Overview

Rename InternalServerError to InvalidTransactionException (as it
more precisely reflects the _most likely_ exception from Node#submit)
and make it unchecked (as _any_ exception is unrecoverable).

There is, arguably, the same issue that the exception type does not
always work — if ApiSender#send fails, than the exception type
is misleading. That is to be addressed in ECR-3190.

---
See: 
- https://jira.bf.local/browse/ECR-3258
- https://jira.bf.local/browse/ECR-3190

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/core/rust/src/proxy/node.rs
```diff
@@ -130,7 +130,12 @@ pub extern "system" fn Java_com_exonum_binding_service_NodeProxy_nativeSubmit(
                 match node.submit(raw_transaction) {
                     Ok(tx_hash) => convert_hash(&env, &tx_hash),
                     Err(err) => {
-                        let error_class = INTERNAL_SERVER_ERROR;
+                        // node#submit can fail for two reasons: unknown transaction id and
+                        // an error in ApiSender#send. The former is the service implementation
+                        // error and is appropriate to communicate with the exception below;
+                        // the second is an internal, unrecoverable error that has nothing
+                        // to do with the service: ECR-3190
```
MakarovS: What is expected from ECR-3190 after the completion of this PR? From its current description one could conclude that this PR also solves ECR-3190

dmitry-timofeev: yes, updated.

PR #949, opened 2019-06-06
Auto locate the libjvm [ECR-3074]

skletsun:
## Overview
Added a wrapper shell script that adds a path to `libjvm` to the `LD_LIBRARY_PATH` variable and executes the App.

---
See: https://jira.bf.local/browse/ECR-3074

### Definition of Done


exonum-java-binding/core/rust/exonum-java/exonum-java
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+# Try to detect the Java home from installed executable. We don't rely on JAVA_HOME that is already set because it may
+# not work in some cases (like `skdman`).
```
dmitry-timofeev: Is it ok to disallow specifying JAVA_HOME?

vitvakatu: Perhaps we should add a backdoor to specify not-default JVM?

dmitry-timofeev: Does sdkman sets the wrong java-home? Can it be switched off? jenv also _can_ do that, but it does not by default (because it sets the wrong java-home).

We can add something like `EXONUM_JAVA_HOME`. But I'd prefer to use `JAVA_HOME`.

skletsun: Well, I used this initially:
```
JAVA_HOME="${JAVA_HOME:-$(java -XshowSettings:properties -version \
  2>&1 > /dev/null |\
  grep 'java.home' |\
  awk '{print $3}')}"
LIBJVM_PATH="$(find ${JAVA_HOME} -type f -name libjvm.* | xargs -n1 dirname)"
```
and it turned out that `JAVA_HOME` on my machine was set to `/Users/skletsun/.sdkman/candidates/java/current` which is a symlink to currently chosen sdk. And obviously, searching over a symlink using current flags did nothing.

For the moment I ended up with a command line arguments for `find` that works on OSX and going to check if there common options for `GNU` and `BSD` versions of `find`.

exonum-java-binding/core/rust/src/runtime/cmd.rs
```diff
@@ -145,7 +145,7 @@ fn get_node_config(context: &Context) -> NodeConfig<PathBuf> {
 fn get_path_to_default_log_config() -> String {
     let log_fallback_path = {
         let mut path = executable_directory();
-        path.push("log4j-fallback.xml");
+        path.push("../log4j-fallback.xml");
```
skletsun: What is the best way to deal with this? Keep it as is supposing that our primary target is the app? And update our `qa` and `cc` scripts (that don't rely on the app) by specifying the path to config explicitly (or moving it to correct location)?

exonum-java-binding/core/rust/exonum-java/exonum-java
skletsun: Well, I've ended up with `find -L` which works fine on both OSX and Linux on my side. And it seems `find` now behaves similar to OSX and Linux which feels a little strange to me because I remember there were differences. Anyways, it seems to work.

skletsun:
> We discussed that only the script shall be on the PATH, but the binary is still available.

Corrected. The binary is put to `EXONUM_HOME/bin`.

exonum-java-binding/core/rust/exonum-java/TUTORIAL.md
```diff
@@ -19,20 +19,6 @@ in the [user guide](https://exonum.com/doc/version/0.11/get-started/java-binding
 
 ### Step 1. Configure Environment
 
-#### `LD_LIBRARY_PATH`
```
vitvakatu: We also need to mention that set `JAVA_HOME` will change the JVM to use. Perhaps in a different, section

exonum-java-binding/core/rust/exonum-java/exonum-java
```diff
@@ -0,0 +1,18 @@
...
+  grep 'java.home' |\
+  awk '{print $3}')}"
+
+# Find the directory containing libjvm.
+LIBJVM_PATH="$(find -L ${JAVA_HOME} -type f -name libjvm.* | xargs -n1 dirname)"
+
+export LD_LIBRARY_PATH="${LIBJVM_PATH}"
+
+# Detect the directory of the installed app.
+EXONUM_HOME="$(dirname $0)"
```
skletsun: It seems `dirname $0` fits our needs despite it sometimes returns the relative path.

exonum-java-binding/core/rust/exonum-java/TUTORIAL.md
skletsun: Just thinking if it's worth mentioning. Is there any other reason to set `JAVA_HOME` except explicit pointing to a different JVM? 🤔

exonum-java-binding/core/rust/src/runtime/paths.rs
```diff
@@ -28,7 +28,7 @@ pub fn system_classpath() -> String {
     let mut jars = Vec::new();
     let jars_directory = {
         let mut executable_directory = executable_directory();
-        executable_directory.push("lib/java");
+        executable_directory.push("../lib/java");
```
dmitry-timofeev: Do scripts launching an app with `cargo run` work with such change? If I understand correctly, the app Java dependencies are always copied to `lib/java` somewhere alongside the application.

exonum-java-binding/core/rust/src/runtime/cmd.rs
dmitry-timofeev: (here and below, with lib/java) I would not make one more distinction between a packaged app and the one we get once we build the project and can run with `cargo run`.

Currently they differ in how they load the libraries: the first requires them to be in `$origin/lib/native`, the second relies on absolute paths to libjvm and rust stdlib (I wonder though how Java loads the java_bindings library in the second case). 

Maybe we can we keep the same layout for a packaged and 'normally built' application (e.g., by moving _the script_ to `bin/`)? Is there anything else we can do?

exonum-java-binding/core/rust/exonum-java/TUTORIAL.md
dmitry-timofeev: Probably in the installation/configuration instructions. I think it is, because it allows to configure the VM to use.

skletsun: https://github.com/exonum/exonum-doc/pull/331

dmitry-timofeev: also https://jira.bf.local/browse/ECR-3048

PR #950, opened 2019-06-06
Rename InvalidTransactionException [ECR-3190]:

dmitry-timofeev:
## Overview

As it can occur for several reasons (not only because of
an invalid tx submitted by user), its type must communicate that.

---
See: https://jira.bf.local/browse/ECR-3190

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/src/main/java/com/exonum/binding/service/TransactionSubmissionException.java
+++ b/exonum-java-binding/core/src/main/java/com/exonum/binding/service/TransactionSubmissionException.java
@@ -16,9 +16,12 @@
 
 package com.exonum.binding.service;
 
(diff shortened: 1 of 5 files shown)


exonum-java-binding/core/src/main/java/com/exonum/binding/service/TransactionSubmissionException.java
```diff
@@ -17,11 +17,12 @@
 package com.exonum.binding.service;
 
 /**
- * Indicates that the submitted transaction is not valid (e.g., belongs to an unknown service).
+ * Indicates a transaction could not be submitted. For example, the submitted transaction
```
MakarovS: `that a transaction could not be submitted`?

PR #953, opened 2019-06-11
Add get blocks in range [ECR-3255]

dmitry-timofeev:
## Overview

---
**:warning: See: https://jira.bf.local/browse/ECR-3255**

### Definition of Done

- [x] Public API has Javadoc


exonum-light-client/src/main/java/com/exonum/client/ExonumClient.java
```diff
@@ -108,42 +108,45 @@
 
   /**
    * Returns blockchain blocks information for the requested range. The blocks are returned
-   * in reverse order, starting from the {@code heightMax}.
-   * @param count Number of blocks to return.
-   *        It should be in range [1, {@linkplain ExonumApi#MAX_BLOCKS_PER_REQUEST}]
-   * @param blockFilter controls whether to skip blocks with no transactions
-   * @param heightMax maximum height of the returned blocks.
+   * in ascending order by their height.
```
bullet-tooth: Should we describe that it can lead to performance degradation in case of large ranges?

exonum-light-client/src/main/java/com/exonum/client/ExonumClient.java
```diff
@@ -108,42 +108,45 @@
...
+   *
+   * @param fromHeight the height of the first block to include. Must be non-negative
+   * @param toHeight the height of the last block to include. Must be greater than
+   *        or equal to {@code fromHeight} and less than or equal to the blockchain height.
    *        If the {@code heightMax} is greater than actual blockchain height then
```
bullet-tooth: old param name in the doc.

exonum-light-client/src/main/java/com/exonum/client/ExonumHttpClient.java
```diff
@@ -141,19 +149,76 @@ public BlockResponse getBlockByHeight(long height) {
...
+    while (remainingBlocks > 0
+        && (nextHeight == null || nextHeight >= GENESIS_BLOCK_HEIGHT)) {
+      int numBlocks = min(remainingBlocks, MAX_BLOCKS_PER_REQUEST);
+      BlocksResponse blocksResponse = doGetBlocks(numBlocks, blockFilter, nextHeight, timeOption);
+
+      blocks.addAll(blocksResponse.getBlocks());
+
+      nextHeight = blocksResponse.getBlocksRangeStart() - 1;
+      blockchainHeight = max(blockchainHeight, blocksResponse.getBlocksRangeEnd() - 1);
+      remainingBlocks = Math.toIntExact(count - (blockchainHeight - nextHeight));
```
bullet-tooth: looks like we can remove `blockchainHeight` and use `blocks.size()` instead

exonum-light-client/src/main/java/com/exonum/client/ExonumHttpClient.java
```diff
@@ -141,19 +149,76 @@ public BlockResponse getBlockByHeight(long height) {
...
+    }
+
+    long fromHeight = max(blockchainHeight - count + 1, GENESIS_BLOCK_HEIGHT);
```
bullet-tooth: looks like it's the wrong statement. For example, we need 10 non-empty blocks in the range 50..100 height, but blocks 91..95 are empty.
`fromHeight` in this case will be 100-10 + 1 = 91.
But the correct one is 86

exonum-light-client/src/test/java/com/exonum/client/ExonumHttpClientBlocksIntegrationTest.java
```diff
@@ -45,28 +53,40 @@
...
 class ExonumHttpClientBlocksIntegrationTest {
-  private static MockWebServer server;
-  private static ExonumClient exonumClient;
 
-  @BeforeAll
-  static void start() throws IOException {
+  private MockWebServer server;
+  private ExonumClient exonumClient;
+
+  @BeforeEach
```
bullet-tooth: it takes some time to start the server...

exonum-light-client/src/main/java/com/exonum/client/ExplorerApiHelper.java
```diff
@@ -139,17 +146,15 @@ private static TransactionResult getTransactionResult(
    */
   @Value
   static class SubmitTxRequest {
-    @SerializedName("tx_body")
-    TransactionMessage body;
+    TransactionMessage txBody;
```
bullet-tooth: it will not work because used for serialization `com.exonum.binding.common.serialization.json.JsonSerializer#json` instance is not configured for names with an underscore

exonum-light-client/src/test/java/com/exonum/client/ExonumHttpClientBlocksIntegrationTest.java
```diff
@@ -115,117 +137,412 @@ void getBlockWrongHeight(long height) {
...
+    long end = last.getHeight() + 1;
+    checkArgument(start < end);
+
+    return createGetBlocksResponse(start, end, blocks);
+  }
+
+  @SuppressWarnings("UnstableApiUsage")
+  private static String createGetBlocksResponse(long start, long end, List<Block> blocks) {
+    // Self-check the blocks to be in ascending order by height
+    assertTrue(isInStrictOrder(blocks, comparing(Block::getHeight)));
```
dmitry-timofeev: I wonder if we shall also add such check to the BlocksRange?

exonum-light-client/src/main/java/com/exonum/client/ExplorerApiHelper.java
dmitry-timofeev: It will — see the patch to this very file configuring Gson.

exonum-light-client/src/test/java/com/exonum/client/ExonumHttpClientBlocksIntegrationTest.java
dmitry-timofeev: First, it is quick enough (tens of ms even for tests requesting thousands of blocks), second — it does not matter at all when one breaking test breaks the other, unrelated, **that otherwise works** (it happened when I started coding with the tests and broken impl)

exonum-light-client/src/main/java/com/exonum/client/ExonumHttpClient.java
dmitry-timofeev: No, we can't, because (a) blockchainHeight is not known until the first request completes; (b) it is needed to infer the number of remaining blocks and/or fromHeight.

exonum-light-client/src/main/java/com/exonum/client/ExonumHttpClient.java
dmitry-timofeev: It is correct indeed — please see the updated spec of this method. It is specified to return _up to_ min(count, blockchainHeight + 1) blocks. Filtering no longer affects the range. E.g., if you request top 10 excluding empty, you will get top 10 excluding empty.

exonum-light-client/src/main/java/com/exonum/client/ExonumClient.java
dmitry-timofeev: Clarified that such behaviour will be fixed in subsequent release (though it was tempting to fix it now with an additional request)

exonum-light-client/src/main/java/com/exonum/client/ExonumHttpClient.java
dmitry-timofeev: > The loop is finished at this point, but blocks.size() is 1000, not 1100.

That is correct by the spec — because it says it will return the correctly sized range of blocks, it will not extend it arbitrarily.

There are two kind of things one might expect from this method: return top N blocks _filtering_ empty; or return top N _non-empty_ blocks. These are different requests. This method allows to request a sized _range_ of blocks, not a certain number of (non-empty) blocks.

The fact that we are having this confusion probably means this API is a monstrosity and we will have to discuss it in a more detail in jira.

dmitry-timofeev: @bullet-tooth @MakarovS @DENMROOT , please see https://jira.bf.local/browse/ECR-3255?focusedCommentId=48525&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-48525

exonum-light-client/src/main/java/com/exonum/client/ExonumClient.java
dmitry-timofeev: I am actually not sure what we shall convey here. The thing that it takes space proportional to the number of blocks sounds obvious and reasonable. No one would reasonably expect anything different. Same goes for the time it takes to fetch them :thinking:

exonum-light-client/src/main/lombok/com/exonum/client/response/BlocksRange.java
```diff
@@ -0,0 +1,90 @@
...
+   */
+  public BlocksRange(long fromHeight, long toHeight, List<Block> blocks) {
+    checkArgument(0 <= fromHeight, "fromHeight (%s) is negative");
+    checkArgument(fromHeight <= toHeight, "fromHeight (%s) > toHeight (%s)", fromHeight, toHeight);
+    this.fromHeight = fromHeight;
+    this.toHeight = toHeight;
+    this.blocks = checkBlocks(fromHeight, toHeight, blocks);
+  }
+
+  private static List<Block> checkBlocks(long fromHeight, long toHeight, List<Block> blocks) {
```
bullet-tooth: do we need to check there are no duplicates in the list? 🤔

dmitry-timofeev: That's a good question how paranoid we shall get here :smile:

exonum-light-client/src/main/java/com/exonum/client/ExplorerApiHelper.java
```diff
@@ -44,18 +47,23 @@
  */
 final class ExplorerApiHelper {
 
+  @VisibleForTesting
+  static final Gson JSON = JsonSerializer.builder()
```
MakarovS: Either create another `Gson` object in `SystemApiHelper` or remove `@VisibleForTesting`.

exonum-light-client/src/main/java/com/exonum/client/request/BlockFilteringOption.java
```diff
@@ -21,8 +21,15 @@
...
   /**
-   * Skip empty blocks (containing no transactions).
-   * Only non-empty blocks will be returned in a response.
+   * Skip empty blocks (containing no transactions). Only non-empty blocks will be returned
+   * in a response.
+   *
+   * <p>Requesting the <em>server</em> to skip empty blocks when the client does not need them
+   * allows to save bandwidth and complete the request faster if a large number of blocks
+   * is requested. However, some blockchains never have empty blocks, in which case this option
+   * has no effect (e.g., an active <a href="https://exonum.com/doc/version/latest/advanced/time/">
```
MakarovS: Specify a version here?

exonum-light-client/src/main/lombok/com/exonum/client/response/Block.java
```diff
@@ -80,8 +80,11 @@
   ZonedDateTime commitTime;
 
   /**
-   * Returns time when the block was committed to the blockchain.
-   * Can be empty if include time parameter is not specified in the request.
+   * Returns the time when the block was committed to the blockchain.
```
MakarovS: `when the block is committed`?

dmitry-timofeev: Why present, aren't they all committed in the past?

exonum-light-client/src/main/java/com/exonum/client/request/BlockFilteringOption.java
dmitry-timofeev: I left it out because the light client is not based on a particular version of Exonum, and may be compatible with several versions (on the other hand, not necessarily the latest :thinking: ). Ideally, it must be the latest version that the _light client_ supports, but I doubt the trouble to keep it in sync is worth it. In EJB it is simpler because it is based on a particular version.

exonum-light-client/src/main/java/com/exonum/client/ExplorerApiHelper.java
dmitry-timofeev: Yes, ideally it must be in some static utility class… Will see if there is a good candidate.

exonum-light-client/src/main/java/com/exonum/client/ExonumClient.java
```diff
@@ -107,45 +109,65 @@
...
-   *        It should be in range [1, {@linkplain ExonumApi#MAX_BLOCKS_PER_REQUEST}]
+   * Returns blockchain blocks in the requested <em>closed</em> range. The blocks are returned
+   * in ascending order by their height.
+   *
+   * @param fromHeight the height of the first block to include. Must be non-negative
+   * @param toHeight the height of the last block to include. Must be greater than
+   *        or equal to {@code fromHeight} and less than or equal to the blockchain height.
+   *        If the {@code toHeight} is greater than actual blockchain height then
+   *        the actual height will be used — such error-prone behaviour will be fixed
+   *        in Exonum 0.12, see ECR-3188)
```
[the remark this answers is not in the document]
dmitry-timofeev: I would say no, added to ECR-3299

exonum-light-client/src/main/java/com/exonum/client/ExplorerApiHelper.java
dmitry-timofeev: Decided to move to ExonumApi instead of adding a separate class to hold that. @bullet-tooth

exonum-light-client/src/main/lombok/com/exonum/client/response/Block.java
MakarovS: You are right, I misunderstood the sentence.

dmitry-timofeev:
@bullet-tooth @MakarovS is the new name better (d882f95), what do you think?

MakarovS:
It's good. What do you think about renaming the method to `getLastBlocksRange`? Then `size` naming would work better and the difference between this method and `findNonEmptyBlocks` is clear. But the current one is good as well.

bullet-tooth:
@dmitry-timofeev it's awesome :)

dmitry-timofeev:
I like `getLastBlocksRange`, but isn't it too long? @bullet-tooth WDYT? I've also updated the Javadoc of `getLastBlocks`.

dmitry-timofeev:
OK, I think we can settle for `getLastBlocks`, and reconsider its name/spec later if they cause confusion.

PR #965, opened 2019-06-20
Fix cryptocurrency and QA scripts

skletsun:
## Overview
Fixed these scripts (that don't rely on packaged app) by introducing extra command line argument for java library path. If not specified - _executable/lib/native_ is used.

---
See: https://jira.bf.local/browse/ECR-3271

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/core/rust/src/runtime/cmd.rs
```diff
@@ -62,6 +63,15 @@ impl CommandExtension for Run {
                 "ejb-log-config-path",
                 false,
             ),
+            Argument::new_named(
+                EJB_LIBRARY_PATH,
+                false,
+                "Path to the directory containing the java_bindings shared library that is passed \
+                to JVM via -Djava.library.path. Mostly for internal usage.",
```
dmitry-timofeev: \+ "Overrides the standard path to native libraries, enabling running the non-packaged exonum-java application. Mostly for internal usage."

exonum-java-binding/core/rust/src/runtime/cmd.rs
```diff
@@ -62,6 +63,15 @@ impl CommandExtension for Run {
...
+                None,
+                "ejb-java-library-path",
```
dmitry-timofeev: I'd pick the longer name `ejb-override-java-library-path` — because it overrides the normal mechanism.

exonum-java-binding/core/rust/src/runtime/config.rs
```diff
@@ -55,15 +55,6 @@ pub struct RuntimeConfig {
...
-
-/// Internal EJB configuration.
-///
-/// Not visible by user, used internally while initializing runtime.
-#[doc(hidden)] // For testing purposes only.
-pub struct InternalConfig {
-    /// EJB system classpath.
-    pub system_class_path: String,
     /// EJB library path.
     pub system_lib_path: Option<String>,
```
dmitry-timofeev: It either has the wrong name: override_system_lib_path/custom_system_lib_path
or it must not be an Option.

I am a little confused by which code shall set up the library path to the
default (application) value of "lib/native" — it used to happen in ServiceFactoryAdapter,
currently ServiceRuntime does it itself. Is it the right responsibility for ServiceRuntime?

As it looks rather confusing, I'd improve the thing.

exonum-java-binding/core/rust/integration_tests/tests/service_runtime_bootstrap.rs
```diff
@@ -35,19 +35,15 @@ fn bootstrap() {
         // Pass log4j path to avoid error messages of mis-configuration
         log_config_path: log4j_path(),
         port: 6300,
+        system_lib_path: None,
```
dmitry-timofeev: This confirms the confusing nature of this parameter and its handling: `None` is passed, but that will make JavaServiceRuntime pick 'lib/native' under the hood
(it didn't use to do that).

I don't think though we shall ever skip 'java.library.path' — I don't know why it was made conditional (@vitvakatu :question: ).

exonum-java-binding/core/rust/exonum-java/start_cryptocurrency_node.sh
```diff
@@ -45,9 +45,7 @@ echo "${SERVICE_NAME} = '${ARTIFACT_PATH}'" >> ${SERVICES_CONFIG_FILE}
 header "PREPARE PATHS"
 
 EJB_LOG_CONFIG_PATH="${EJB_APP_DIR}/log4j-fallback.xml"
-
-export LD_LIBRARY_PATH="$JAVA_LIB_DIR"
-echo "LD_LIBRARY_PATH=${LD_LIBRARY_PATH}"
+JAVA_LIBRARY_PATH="${EJB_ROOT}/core/rust/target/debug"
```
dmitry-timofeev: Is it always `debug` when cargo run is used (it’s probably fine for these scripts if it unconditionally is)?

skletsun: Yeap, `cargo run` without `--release` uses the `dev` profile that always works with `debug` artifacts.

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
```diff
@@ -53,9 +54,20 @@ impl JavaServiceFactoryAdapter {
     fn get_or_create_java_service_runtime(config: Config) -> JavaServiceRuntime {
         // Initialize runtime if it wasn't created before.
         JAVA_SERVICE_RUNTIME_INIT.call_once(|| {
+            // If path is not provided use the standard library path for packaged app.
+            let system_lib_path = if config.runtime_config.override_system_lib_path.is_some() {
```
dmitry-timofeev: I'd consider extracting that in a separate function.

PR #969, opened 2019-06-21
Add clean Rust target Maven command [ECR-3287].

dmitry-timofeev:
## Overview

Run `mvn clean -DcleanRust` and you are clean :shower: 

---
See: https://jira.bf.local/browse/ECR-3287

### Definition of Done


dmitry-timofeev:
Again, it is rarely needed to clean Rust target unlike Maven target, but when it is, it saves on `cd core/rust && cargo clean && cd - ` / `rm -rf core/rust/target`.

PR #972, opened 2019-06-26
Add TestKit#getSnapshot [ECR-3216]

MakarovS:
## Overview
Add `TestKit#getSnapshot`, replace `TestKit#withSnapshot` with `TestKit#applySnapshot` and make `TestKit#withSnapshot` accept `Consumer<Snapshot> snapshotFunction` as a parameter.

---
See: https://jira.bf.local/browse/ECR-3216

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/qa-service/src/test/java/com/exonum/binding/qaservice/QaServiceImplIntegrationTest.java
```diff
@@ -68,33 +69,30 @@
...
+    MapIndex<HashCode, String> counterNames = schema.counterNames();
+
+    HashCode defaultCounterId = sha256().hashString(DEFAULT_COUNTER_NAME, UTF_8);
+    HashCode afterCommitCounterId = sha256().hashString(AFTER_COMMIT_COUNTER_NAME, UTF_8);
+
+    assertThat(counters.get(defaultCounterId)).isEqualTo(0L);
+    assertThat(counterNames.get(defaultCounterId)).isEqualTo(DEFAULT_COUNTER_NAME);
+    assertThat(counters.get(afterCommitCounterId)).isEqualTo(0L);
+    assertThat(counterNames.get(afterCommitCounterId)).isEqualTo(AFTER_COMMIT_COUNTER_NAME);
+;
```
dmitry-timofeev: ! Here and below

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -98,6 +99,8 @@
   private static final short TIME_SERVICE_ID = 4;
 
   private final Map<Short, Service> services = new HashMap<>();
+  @VisibleForTesting
+  final List<Cleaner> snapshotCleaners = new ArrayList<>();
```
dmitry-timofeev: Why is there a List and not a single Cleaner? If the snapshots are destroyed at the same time, a single Cleaner will work (and will be more reliable in case of errors).

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -289,23 +292,58 @@ private void checkTransaction(TransactionMessage transactionMessage) {
...
    *
    * @param snapshotFunction a function to execute
    * @param <ResultT> a type the function returns
    * @return the result of applying the given function to the database state
    */
-  public <ResultT> ResultT withSnapshot(Function<Snapshot, ResultT> snapshotFunction) {
-    try (Cleaner cleaner = new Cleaner("TestKit#withSnapshot")) {
-      long snapshotHandle = nativeCreateSnapshot(nativeHandle.get());
-      Snapshot snapshot = Snapshot.newInstance(snapshotHandle, cleaner);
+  public <ResultT> ResultT applySnapshot(Function<Snapshot, ResultT> snapshotFunction) {
```
dmitry-timofeev: The docs must mention why these methods exist on top of #getSnapshot and explain when they shall be used instead of #getSnapshot.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -289,23 +292,58 @@ private void checkTransaction(TransactionMessage transactionMessage) {
...
 
+  /**
+   * Performs a given function with a snapshot of the current database state (i.e., the one that
+   * corresponds to the latest committed block). In-pool (not yet processed) transactions are also
+   * accessible with it in {@linkplain Blockchain#getTxMessages() blockchain}.
+   *
+   * @param snapshotFunction a function to execute
+   */
+  public void withSnapshot(Consumer<Snapshot> snapshotFunction) {
+    try (Cleaner cleaner = new Cleaner("TestKit#withSnapshot")) {
```
dmitry-timofeev:
```java
    applySnapshot(s -> {
      snapshotConsumer.accept(s);
      return null;
    });
```

And move :arrow_up_small:

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -289,23 +292,58 @@ private void checkTransaction(TransactionMessage transactionMessage) {
...
    * Performs a given function with a snapshot of the current database state (i.e., the one that
-   * corresponds to the latest committed block). In-pool (not yet processed) transactions are also
-   * accessible with it in {@linkplain Blockchain#getTxMessages() blockchain}.
+   * corresponds to the latest committed block) and returns a result of its execution. In-pool
+   * (not yet processed) transactions are also accessible with it in
+   * {@linkplain Blockchain#getTxMessages() blockchain}.
    *
    * @param snapshotFunction a function to execute
    * @param <ResultT> a type the function returns
    * @return the result of applying the given function to the database state
```
dmitry-timofeev: \+ (here and elsewhere) @link alternatives: #withSnapshot and #getSnapshot

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -289,23 +292,58 @@ private void checkTransaction(TransactionMessage transactionMessage) {
...
+      throw new RuntimeException(e);
+    }
+  }
+
+  /**
+   * Returns a snapshot of the current database state (i.e., the one that
+   * corresponds to the latest committed block). In-pool (not yet processed) transactions are also
+   * accessible with it in {@linkplain Blockchain#getTxMessages() blockchain}.
+   *
+   * <p>All created snapshots are deleted when corresponding TestKit is disposed.
```
dmitry-timofeev:
```java
   * <p>All created snapshots are deleted when this TestKit is {@linkplain #close() closed}.
   * It is forbidden to access the snapshots once the TestKit is closed.
   *
   * <p>If you create many snapshots (more than a hundred), it is recommended
   * to use {@link #withSnapshot(Consumer)} or {@link #applySnapshot(Function)},
   * which destroy the snapshots once the passed closure completes.
```

dmitry-timofeev: Not sure about a hundred, actually, any number will not _generally_ work because you can create five, but each will allocate a million indexes. Explaining that, however, would require explaining how our proxies and cleaners work — but their point is to isolate the users from this knowledge.

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTest.java
```diff
@@ -536,6 +522,21 @@ void createTestKitWithTimeServiceAndTooManyValidators() {
...
   }
 
+  @Test
+  void getSnapshot(TestKit testKit) {
+    Snapshot view1 = testKit.getSnapshot();
+    Snapshot view2 = testKit.getSnapshot();
+    assertThat(testKit.snapshotCleaners).hasSize(2);
+
+    testKit.close();
+    Cleaner cleaner1 = view1.getCleaner();
```
dmitry-timofeev: Snapshots have com.exonum.binding.core.storage.database.View.getViewNativeHandle
which is specified to throw ISE once it is closed. I'd use that instead of relying on
testkit internals.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
MakarovS: Should we rename Cleaner description in `applySnapshot`? Now it'll be `TestKit#applySnapshot`, even when users call `TestKit#withSnapshot`, which might be confusing in some cases

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -316,9 +366,18 @@ public EmulatedNode getEmulatedNode() {
...
   protected void disposeInternal() {
+    closeSnapshotCleaners();
     nativeFreeTestKit(nativeHandle.get());
   }
 
+  private void closeSnapshotCleaners() {
+    try {
+      snapshotCleaner.close();
+    } catch (CloseFailuresException e) {
+      throw new RuntimeException(e);
```
dmitry-timofeev: (1) Inappropriate exception type: IllegalStateException would be more suitable;
(2) The testkit itself will be leaked if such exception occurs. We might add it as a CleanAction
to ensure reliable destruction, but I think using cleaner here (in explicitly closeable classes
that control that Cleaner) would add unnecessary indirection, and try/finally will work better:

```java
 protected void disposeInternal() {
    try {
      snapshotCleaner.close();
    } catch (CloseFailuresException e) {
      throw …
    } finally {
      nativeFreeTestKit(nativeHandle.get());
    }
```

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -292,20 +295,67 @@ private void checkTransaction(TransactionMessage transactionMessage) {
...
-  public <ResultT> ResultT withSnapshot(Function<Snapshot, ResultT> snapshotFunction) {
-    try (Cleaner cleaner = new Cleaner("TestKit#withSnapshot")) {
-      long snapshotHandle = nativeCreateSnapshot(nativeHandle.get());
-      Snapshot snapshot = Snapshot.newInstance(snapshotHandle, cleaner);
+  public <ResultT> ResultT applySnapshot(Function<Snapshot, ResultT> snapshotFunction) {
+    try (Cleaner cleaner = new Cleaner("TestKit#applySnapshot")) {
+      Snapshot snapshot = createSnapshot(cleaner);
       return snapshotFunction.apply(snapshot);
     } catch (CloseFailuresException e) {
       throw new RuntimeException(e);
```
dmitry-timofeev: Exception type.

exonum-java-binding/testkit/src/main/java/com/exonum/binding/testkit/TestKit.java
```diff
@@ -292,20 +295,67 @@ private void checkTransaction(TransactionMessage transactionMessage) {
    * corresponds to the latest committed block). In-pool (not yet processed) transactions are also
    * accessible with it in {@linkplain Blockchain#getTxMessages() blockchain}.
    *
+   * <p>This method destroys the snapshot once the passed closure completes, compared to
+   * {@link #getSnapshot()}, which disposes created snapshots only when closing the TestKit.
```
dmitry-timofeev: Nit: I think it must say "when TestKit is closed" [by something else] because getSnapshot does not close it itself.

PR #989, opened 2019-07-02
Add TestKit to BOM and archetype modules [ECR-3056]

MakarovS:
## Overview
Add TestKit to BOM and archetype modules.

---
See: https://jira.bf.local/browse/ECR-3056

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/service-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -97,6 +98,11 @@
       <version>${mockito.version}</version>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>com.exonum.binding</groupId>
+      <artifactId>exonum-testkit</artifactId>
```
dmitry-timofeev: test scope?

dmitry-timofeev:
The archetype project must depend on the testkit as well (as it does on core in runtime scope), as documented in the issue.

exonum-java-binding/service-archetype/pom.xml
```diff
@@ -43,6 +51,11 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-archetype-plugin</artifactId>
           <version>${maven-archetype.version}</version>
+          <configuration>
+            <properties>
+              <nativeLibPath>${project.parent.basedir}/core/rust/target/debug</nativeLibPath>
```
dmitry-timofeev: @dip56 Would you please test that release builds continue to work (we didn't use to use the native library in archetype ITs), and the generated project works with the installed app (the current 0.7.0-SNAPSHOT)?

exonum-java-binding/service-archetype/src/main/resources/archetype-resources/src/test/java/ServiceModuleTest.java
```diff
@@ -45,6 +48,19 @@ void testTransactionConverterBinding() {
     assertThat(s, instanceOf(MyTransactionConverter.class));
   }
 
+  @Test
+  void testServiceInstantiation() {
```
bullet-tooth: wouldn't it be better to create a separate class with junit5 extension?

dmitry-timofeev: Not sure. This particular test is only useful at project initialization to verify the app is installed properly + as an example of testkit configuration :thinking: We can extend some comments here and there/modify the test display name to make their purpose more apparent.

dmitry-timofeev: I removed the existing tests as they do not bring much value on top of this new one.

dmitry-timofeev: and documented the likely failure

PR #991, opened 2019-07-04
Use system sodium in the crypto function by default: [ECR-3288]

dmitry-timofeev:
## Overview

That allows the users to benefit from the system libsodium
whenever it is available (it always is when the app is run,
but might not be in other usages).

---
See: https://jira.bf.local/browse/ECR-3288

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/CHANGELOG.md
+++ b/exonum-java-binding/CHANGELOG.md
@@ -15,6 +15,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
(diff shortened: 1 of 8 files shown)


bullet-tooth:
I believe that `com.exonum.binding.common.crypto.CryptoFunctions#ed25519` javadoc should be updated as well

dmitry-timofeev:
> javadoc should be updated as well

How exactly? The system library thing?

bullet-tooth:
```
  /**
   * Returns a ED25519 public-key signature system crypto function.
   */
```

PR #997, opened 2019-07-10
[ECR-3223] LC: port is optional; added prefix

bullet-tooth:
## Overview

---
See: https://jira.bf.local/browse/ECR-3223

- now the port is optional in the host URL
- a prefix can be set to all requests

### Definition of Done

- [x] Public API has Javadoc


exonum-light-client/src/main/java/com/exonum/client/HttpUrlHelper.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.net.URL;
+import java.util.Map;
+import okhttp3.HttpUrl;
+import okhttp3.HttpUrl.Builder;
+
+final class HttpUrlHelper {
+
+  static HttpUrl getFullUrl(URL host, String prefix, String relativeUrl,
+      Map<String, String> query) {
```
dmitry-timofeev: (encoded)QueryParams/Parameters?

exonum-light-client/src/test/java/com/exonum/client/HttpUrlHelperTest.java
```diff
@@ -0,0 +1,70 @@
...
+   * Provides a combination of parameters by the following rules:
+   * - port is optional
+   * - prefix is optional
+   * - paths can start either with or without heading slash
+   * - query params is optional.
+   */
+  private static List<Arguments> source() {
+    Map<String, String> noQuery = emptyMap();
+    return ImmutableList.of(
+        of("http://localhost/path/to/source",
```
dmitry-timofeev: Lone `of` is off 🙃 , please replace with `arguments`.

exonum-light-client/src/main/java/com/exonum/client/HttpUrlHelper.java
```diff
@@ -0,0 +1,65 @@
...
+        .host(host.getHost());
+    if (host.getPort() != -1) {
+      builder.port(host.getPort());
+    }
+    return builder;
+  }
+
+  /**
+   * Removes heading slash from the path.
+   * Useful because underlying OkHttp applies slashes when constructing paths.
```
dmitry-timofeev: Doesn't okhttp do that itself? Also, double slashes must not cause any problems, mustn't they?

exonum-light-client/src/test/java/com/exonum/client/HttpUrlHelperTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    assertThat(url.toString(), is(expectedUrl));
+  }
+
+  /**
+   * Provides a combination of parameters by the following rules:
+   * - port is optional
+   * - prefix is optional
+   * - paths can start either with or without heading slash
+   * - query params is optional.
```
MakarovS: `are optional`?

exonum-light-client/src/main/java/com/exonum/client/HttpUrlHelper.java
bullet-tooth: No, okhttp doesn't do that :( 
Urls with double stashes are valid. But such urls semantically identical but not equal. It may cause issues when storing them in collections and potentially cause some issues on a server side, for example caching (where the cache key is a path to the resource).
I see this question is more complex: such as we allow users to specify host and prefix in a string form may be do we need to have a full normalization urls support, for example as described [here](https://github.com/Norconex/commons-lang/blob/master/norconex-commons-lang/src/main/java/com/norconex/commons/lang/url/URLNormalizer.java#L56) ?

dmitry-timofeev: I see, do I understand it correctly that we do the normalization to aid caching on a (possibly, intermediate) server in cases where the user has optional leading and/or trailing slashes? Isn't the naive approach used here enough?

exonum-light-client/src/main/java/com/exonum/client/HttpUrlHelper.java
```diff
@@ -48,16 +49,13 @@ static HttpUrl getFullUrl(URL host, String prefix, String relativeUrl,
...
    */
-  private static String normalize(String path) {
-    requireNonNull(path);
-    if (path.startsWith("/")) {
-      return path.substring(1);
-    } else {
-      return path;
-    }
+  private static HttpUrl normalize(HttpUrl url) {
+    URI normalized = url.uri().normalize();
```
dmitry-timofeev: It is no longer a naïve approach 🙃 Shall we add more tests specific to this normalization algorithm to exercise that?

exonum-light-client/src/main/java/com/exonum/client/HttpUrlHelper.java
bullet-tooth: Yes, right.
But, I believe, we can simplify this and remove normalization and rely fully on OkHttp.

dmitry-timofeev: I must admit that the URLs look better normalized, but agree it is not strictly needed (if a user cares, they can omit the slashes) :-)

dmitry-timofeev:
Would you please also add a short changelog entry?

The build currently fails due unused imports.

exonum-light-client/src/test/java/com/exonum/client/TestUtils.java
```diff
@@ -36,4 +40,12 @@ static String toHex(TransactionMessage message) {
     return HEX_ENCODER.encode(message.toBytes());
   }
 
+  static void assertExactPath(RecordedRequest request, String url) {
+    assertThat(request.getPath(), is("/" + url));
```
bullet-tooth: 🤷‍♀️

dmitry-timofeev: I'd try with matchers, as these are unclear.

PR #1007, opened 2019-07-12
Add the benchmarking module [ECR-2383]

dmitry-timofeev:
## Overview

This module is intended for:
- Faster, easier, _immediate_ experiments, that do not require installing the SuT into the local Maven repository to re-run benchmarks; and allow to re-build them in one command: `mvn package -pl benchmarks -am`
- Some benchmarks of Exonum Java operations that we would like to always keep updated and maintain with Exonum Java evolution.

It replaces neither exonum/exonum-benchmarking nor exonum/exonum-java-benchmarks.

---
See: https://jira.bf.local/browse/ECR-2383

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/benchmarks/README.md
+++ b/exonum-java-binding/benchmarks/README.md
@@ -6,7 +6,7 @@ It is configured to use [Java Microbenchmark Harness][jmh-home] (JMH).
 
 [jmh-home]: https://openjdk.java.net/projects/code-tools/jmh/
 
-*This module is included in this project for easier prototyping and experimentation.
+*This module is included in the project for easier prototyping and experimentation.
 It does* not *contain system-level benchmarks of Exonum services, which currently 
 reside in an internal project:* 
 [exonum/exonum-benchmarking](https://github.com/exonum/exonum-benchmarking)
@@ -17,7 +17,7 @@ project.
 
 ## Usage
 
-Build the `benchmarks.jar` (from exonum-java-binding directory):
+Build the `benchmarks.jar` (from [exonum-java-binding](..) directory):
 
 ```
 # Build the benchmarks and the modules it depends upon 
--- a/exonum-java-binding/benchmarks/README.md
+++ b/exonum-java-binding/benchmarks/README.md
@@ -1,0 +1,31 @@
+# Exonum Java Benchmarks
+
+This module contains benchmarks of various Exonum Java components.
+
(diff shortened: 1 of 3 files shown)


exonum-java-binding/benchmarks/pom.xml
```diff
@@ -0,0 +1,124 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Copyright (c) 2014, Oracle America, Inc.
```
MakarovS: Why Oracle copyright?

dmitry-timofeev: Because the archetype, with which the project is generated, includes this header into the pom. IANAL, but it does not seem to restrict the license we can apply to the module. Also, we do not distribute this module in binary.

exonum-java-binding/benchmarks/README.md
```diff
@@ -0,0 +1,31 @@
+# Exonum Java Benchmarks
+
+This module contains benchmarks of various Exonum Java components.
+
+It is configured to use [Java Microbenchmark Harness][jmh-home] (JMH).
+
+[jmh-home]: https://openjdk.java.net/projects/code-tools/jmh/
+
+*This module is included in this project for easier prototyping and experimentation.
```
bullet-tooth: To avoid the tautology: 
```suggestion
*The module is included in this project for easier prototyping and experimentation.
```

exonum-java-binding/benchmarks/README.md
```diff
@@ -0,0 +1,31 @@
...
+in [exonum/exonum-java-benchmarks](https://github.com/exonum/exonum-java-benchmarks)
+project.
+
+## Usage
+
+Build the `benchmarks.jar` (from exonum-java-binding directory):
+
+```
+# Build the benchmarks and the modules it depends upon 
+mvn package -pl benchmarks -am
```
bullet-tooth: looks like it will fail due to `tests_profile` because this check comes from `exonum-java-binding-parent`

dmitry-timofeev: I think it is safe to assume anyone running mvn commands inside `ejb` has the tests_profile already set and not repeat it here.

exonum-java-binding/benchmarks/README.md
dmitry-timofeev: :+1: , flipped them

PR #1011, opened 2019-07-15
Use static rocksdb linkage on Mac [ECR-3324]

vitvakatu:
## Overview

RocksDB is linked statically on Mac to avoid problems with dynamic library resolving.. It is required to set `ROCKSDB_LIB_DIR`  prior to running `package_app.sh`.

---
See: https://jira.bf.local/browse/ECR-3324

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/package_app.sh
+++ b/exonum-java-binding/package_app.sh
@@ -52,6 +52,11 @@ function build-exonum-java-macos() {
       echo "Please set ROCKSDB_LIB_DIR"
       exit 1
     fi
+    # Check if SNAPPY_LIB_DIR is set
+    if [ -z "${SNAPPY_LIB_DIR:-}" ]; then
+      echo "Please set SNAPPY_LIB_DIR"
+      exit 1
+    fi
     build-exonum-java-for-platform "@loader_path" "libjava_bindings.dylib"
 }
 
--- a/exonum-java-binding/package_app.sh
+++ b/exonum-java-binding/package_app.sh
@@ -52,11 +52,6 @@ function build-exonum-java-macos() {
       echo "Please set ROCKSDB_LIB_DIR"
       exit 1
     fi
-    # Check if SNAPPY_LIB_DIR is set
-    if [ -z "${SNAPPY_LIB_DIR:-}" ]; then
-      echo "Please set SNAPPY_LIB_DIR"
-      exit 1
-    fi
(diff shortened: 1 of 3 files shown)


exonum-java-binding/package_app.sh
```diff
@@ -44,6 +44,13 @@ function build-exonum-java-for-platform() {
 }
 
 function build-exonum-java-macos() {
+    # We use static linkage for RocksDB on Mac
```
dmitry-timofeev: I’d explain briefly why.

exonum-java-binding/package_app.sh
```diff
@@ -44,6 +44,19 @@ function build-exonum-java-for-platform() {
 }
 
 function build-exonum-java-macos() {
+    # We use static linkage for RocksDB on Mac because we depend on RocksDB 5.18.3
+    # which is not available via Homebrew
+    export ROCKSDB_STATIC=1
+    # Check if ROCKSDB_LIB_DIR is set
+    if [ -z "${ROCKSDB_LIB_DIR:-}" ]; then
```
bullet-tooth: why do we need these variables? It's always the same `/usr/local/lib`. Can we set them here instead of requiring that from users?

vitvakatu: We're not requiring this from users, because our users will use the Homebrew package :wink:
Setting this variable explicitly will allow you to choose what RocksDB version to use. Besides, it is also mentioned in [Exonum install guide](https://exonum.com/doc/version/latest/get-started/install/)

bullet-tooth: Mb it'll be better to replace by the following (in pseudo-code):
```
export ROCKSDB_LIB_DIR={ROCKSDB_LIB_DIR:-/usr/local/lib}
IF $ROCKSDB_LIB_DIR/librocksdb.dylib not found THEN fail(RocksDB not installed)
```

dip56: actually, you can use a specific version of the rocksdb and use general path for snappy:
```
export ROCKSDB_LIB_DIR=/usr/local/Cellar/rocksdb/6.1.2/lib
or
export ROCKSDB_LIB_DIR=/usr/local/Cellar/rocksdb5/5.18.3/lib
and
export SNAPPY_LIB_DIR=/usr/local/lib
```

dmitry-timofeev:
Rocksdb 6 is fine.

PR #1014, opened 2019-07-16
Improve the script generating Javadocs:

dmitry-timofeev:
## Overview

Clean and install the modules so that it works when
the local repo does not contain the artifacts with a newly set
version.

Also rename `doNotBuildRustLib` property to more usual
skip\<action>: `skipRustLibBuild`

---

### Definition of Done


dmitry-timofeev:
@dip56 , the `doNotBuildRustLib` property needs to be updated if used in any internal scripts.

PR #1015, opened 2019-07-16
[ECR-3344] LC: Fixed block path. Specified exact path in tests

bullet-tooth:
## Overview

---
See: https://jira.bf.local/browse/ECR-3344
- Fixed path in `getBlockByHeight` method
- Specified exact string paths in tests

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-light-client/src/test/java/com/exonum/client/ExonumHttpClientBlocksIntegrationTest.java
+++ b/exonum-light-client/src/test/java/com/exonum/client/ExonumHttpClientBlocksIntegrationTest.java
@@ -31,8 +31,7 @@ import static com.exonum.client.Blocks.BLOCK_3_WITHOUT_TIME;
 import static com.exonum.client.Blocks.aBlock;
 import static com.exonum.client.ExonumApi.JSON;
 import static com.exonum.client.ExonumApi.MAX_BLOCKS_PER_REQUEST;
-import static com.exonum.client.ExonumUrls.BLOCK;
-import static com.exonum.client.ExonumUrls.BLOCKS;
+import static com.exonum.client.RecordedRequestMatchers.hasPath;
 import static com.exonum.client.RecordedRequestMatchers.hasPathStartingWith;
 import static com.exonum.client.request.BlockFilteringOption.INCLUDE_EMPTY;
 import static com.exonum.client.request.BlockFilteringOption.SKIP_EMPTY;
@@ -82,6 +81,8 @@ import org.junit.jupiter.params.provider.ValueSource;
 
 class ExonumHttpClientBlocksIntegrationTest {
 
+  private final String expectedBlocksPath = "api/explorer/v1/blocks";
+
   private MockWebServer server;
   private ExonumClient exonumClient;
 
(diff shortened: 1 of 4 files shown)


exonum-light-client/src/test/java/com/exonum/client/ExonumHttpClientBlocksIntegrationTest.java
```diff
@@ -82,6 +81,8 @@
 
 class ExonumHttpClientBlocksIntegrationTest {
 
+  private final String expectedBlocksPath = "api/explorer/v1/blocks";
```
dmitry-timofeev: Isn't it equal to `BLOCKS`? Or did you inline it everywhere on purpose?

exonum-light-client/src/test/java/com/exonum/client/ExonumHttpClientIntegrationTest.java
```diff
@@ -209,8 +203,7 @@ void getTransactionNotFound() throws InterruptedException {
     // Assert request params
     RecordedRequest recordedRequest = server.takeRequest();
     assertThat(recordedRequest.getMethod(), is("GET"));
-    assertThat(recordedRequest, hasPathStartingWith(TRANSACTIONS));
-    assertThat(recordedRequest.getRequestUrl().queryParameter("hash"), is(id.toString()));
+    assertThat(recordedRequest, hasPath("api/explorer/v1/transactions?hash=" + id.toString()));
```
dmitry-timofeev: If we use `recordedRequest.getRequestUrl().encodedPath()` in the
matcher, we can make an assertion on the path only (that do not require constructing a full request or comparing a prefix only), and keep separate assertions on the request parameters. Will it make sense?

dmitry-timofeev: And, importantly, will also make `hasPathStartingWith` obsolete.

exonum-light-client/src/test/java/com/exonum/client/ExonumHttpClientBlocksIntegrationTest.java
bullet-tooth: Yes, now it's equal. But what if an error in the BLOCKS path? So I inlined it to the variable for removing duplicates

exonum-light-client/src/test/java/com/exonum/client/ExonumHttpClientIntegrationTest.java
bullet-tooth: It's used in `com.exonum.client.ExonumHttpClientWithPrefixIntegrationTest`.

dmitry-timofeev: I see, well, obsolete for all tests but one :-)

bullet-tooth:
I had a quick overview of Truth and it looks promising and beauty.
Another question is how many customs types do we have, such as it's designed for easy extension. Probably, quite a lot i.e. `HashCode`, `ListIndex`, `MapIndex` etc.

dmitry-timofeev:
I haven't pondered on it enough, but I think we might consider employing a trick we already use with our indices: converting them into Java Collections to enjoy already made matchers. With Truth it can be done with their 'extractors':

```java
assertThat(fooIndex).asMap() // -> MapSubject from Truth, 
                             // we don't have to write MapIndexSubject
  .contains(foo);
```

It might have some performance implications, though.

PR #1024, opened 2019-07-18
Implement transformation of Fork into patch [ECR-3330]

vitvakatu:
## Overview

~~Rust side invalidates provided fork by replacing it with another one (created in place). Java side must clean this fork to avoid memory leaks and must not allow any operation involving this fork.~~

See review comments for questions.

---
See: https://jira.bf.local/browse/ECR-3330

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/rust/src/storage/db.rs
+++ b/exonum-java-binding/core/rust/src/storage/db.rs
@@ -91,6 +91,18 @@ impl View {
     pub fn get(&mut self) -> &mut ViewRef {
         &mut self.reference
     }
+
+    /// Unwraps the stored Fork from the View, panics if it's not possible.
+    pub fn into_fork(self) -> Box<Fork> {
+        if let Some(view_owned) = self._owned {
+            match view_owned {
+                ViewOwned::Snapshot(_) => panic!("`into_fork` called on Snapshot"),
+                ViewOwned::Fork(fork) => fork,
+            }
+        } else {
+            panic!("`into_fork` called on non-owning View");
+        }
+    }
 }
 
 impl ViewRef {
--- a/exonum-java-binding/core/rust/src/storage/db.rs
+++ b/exonum-java-binding/core/rust/src/storage/db.rs
@@ -34,10 +34,10 @@ pub(crate) type Value = Vec<u8>;
 /// should be placed in the heap to prevent its movement after creating a reference to it.
 
 pub(crate) struct View {
-    /// Allows to drop owned view if needed
-    owned: Option<ViewOwned>,
     /// Unsafe reference to the owned View (owned either by Rust or by Java)
     reference: ViewRef,
+    /// Allows to drop owned view if needed
+    owned: Option<ViewOwned>,
 }
 
 enum ViewOwned {
@@ -53,9 +53,9 @@ pub(crate) enum ViewRef {
 impl View {
     pub fn from_owned_snapshot(snapshot: Box<Snapshot>) -> Self {
         View {
-            owned: Some(ViewOwned::Snapshot(snapshot)),
             // Make a "self-reference" to a value stored in the `owned` field.
             reference: unsafe { ViewRef::from_snapshot(&*snapshot) },
+            owned: Some(ViewOwned::Snapshot(snapshot)),
         }
     }
 
@@ -64,9 +64,9 @@ impl View {
         // and will not break the `reference` field.
         let fork = Box::new(fork);
         View {
-            owned: Some(ViewOwned::Fork(fork)),
             // Make a "self-reference" to a value stored in the `owned` field.
             reference: unsafe { ViewRef::from_fork(&*fork) },
+            owned: Some(ViewOwned::Fork(fork)),
         }
     }
 
@@ -74,8 +74,8 @@ impl View {
     #[allow(dead_code)]
     pub fn from_ref_snapshot(snapshot: &Snapshot) -> Self {
         View {
-            owned: None,
             reference: unsafe { ViewRef::from_snapshot(snapshot) },
+            owned: None,
         }
     }
 
@@ -83,8 +83,8 @@ impl View {
     #[allow(dead_code)]
     pub fn from_ref_fork(fork: &Fork) -> Self {
         View {
-            owned: None,
             reference: unsafe { ViewRef::from_fork(fork) },
+            owned: None,
         }
     }
 
--- a/exonum-java-binding/core/rust/src/storage/db.rs
+++ b/exonum-java-binding/core/rust/src/storage/db.rs
@@ -34,9 +34,9 @@ pub(crate) type Value = Vec<u8>;
 /// should be placed in the heap to prevent its movement after creating a reference to it.
 
 pub(crate) struct View {
-    // The `owned` field is used, but its value only needed for the drop stage,
-    // so `Box<Fork>`/`Box<Snapshot>` will be dropped when an instance of `View` leaves the scope.
-    _owned: Option<ViewOwned>,
+    /// Owned allows to drop owned view if needed
+    owned: Option<ViewOwned>,
+    /// Unsafe reference to the owned View (owned either by Rust or by Java)
     reference: ViewRef,
 }
 
@@ -55,7 +55,7 @@ impl View {
         View {
             // Make a "self-reference" to a value stored in the `owned` field.
             reference: unsafe { ViewRef::from_snapshot(&*snapshot) },
-            _owned: Some(ViewOwned::Snapshot(snapshot)),
+            owned: Some(ViewOwned::Snapshot(snapshot)),
         }
     }
 
@@ -66,7 +66,7 @@ impl View {
         View {
             // Make a "self-reference" to a value stored in the `owned` field.
             reference: unsafe { ViewRef::from_fork(&*fork) },
-            _owned: Some(ViewOwned::Fork(fork)),
+            owned: Some(ViewOwned::Fork(fork)),
         }
     }
 
@@ -75,7 +75,7 @@ impl View {
     pub fn from_ref_snapshot(snapshot: &Snapshot) -> Self {
         View {
             reference: unsafe { ViewRef::from_snapshot(snapshot) },
-            _owned: None,
+            owned: None,
         }
     }
 
@@ -84,7 +84,7 @@ impl View {
     pub fn from_ref_fork(fork: &Fork) -> Self {
         View {
             reference: unsafe { ViewRef::from_fork(fork) },
-            _owned: None,
+            owned: None,
         }
     }
 
@@ -94,7 +94,7 @@ impl View {
 
     /// Unwraps the stored Fork from the View, panics if it's not possible.
     pub fn into_fork(self) -> Box<Fork> {
-        if let Some(view_owned) = self._owned {
+        if let Some(view_owned) = self.owned {
             match view_owned {
                 ViewOwned::Snapshot(_) => panic!("`into_fork` called on Snapshot"),
                 ViewOwned::Fork(fork) => fork,
@@ -106,7 +106,7 @@ impl View {
 
     /// Returns `true` iff `into_fork` conversion is possible.
     pub fn can_convert_into_fork(&self) -> bool {
-        match self._owned {
+        match self.owned {
             None => false,
             Some(ViewOwned::Snapshot(_)) => false,
             Some(ViewOwned::Fork(_)) => true,
@@ -167,7 +167,7 @@ mod tests {
         let fork = db.fork();
         let mut view = View::from_ref_fork(&fork);
         check_ref_fork(&mut view);
-        assert!(view._owned.is_none());
+        assert!(view.owned.is_none());
     }
 
     #[test]
@@ -176,7 +176,7 @@ mod tests {
         let snapshot = db.snapshot();
         let mut view = View::from_ref_snapshot(&*snapshot);
         check_ref_snapshot(&mut view);
-        assert!(view._owned.is_none());
+        assert!(view.owned.is_none());
     }
 
     // Creates database with a prepared state.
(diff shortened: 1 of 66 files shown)


exonum-java-binding/core/rust/src/storage/db.rs
```diff
@@ -91,6 +91,18 @@ impl View {
     pub fn get(&mut self) -> &mut ViewRef {
         &mut self.reference
     }
+
+    /// Unwraps the stored Fork from the View, panics if it's not possible.
+    pub fn into_fork(self) -> Box<Fork> {
```
vitvakatu: Looks like this method is used only once, but it's the simplest way.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/database/View.java
```diff
@@ -70,8 +72,9 @@ public boolean canModify() {
    *
    *  @throws IllegalStateException if the view is invalid (closed or nullptr)
    */
-  public long getViewNativeHandle() {
-    return super.getNativeHandle();
+  public NativeHandle getViewNativeHandle() {
+    checkState(nativeHandle.isValid(), "This handle is not valid: %s", this);
```
vitvakatu: This is code duplication and requires public `isValid` (which is also used by tests). Is there better solution?

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/database/View.java
```diff
@@ -70,8 +72,9 @@ public boolean canModify() {
    *
    *  @throws IllegalStateException if the view is invalid (closed or nullptr)
    */
-  public long getViewNativeHandle() {
-    return super.getNativeHandle();
+  public NativeHandle getViewNativeHandle() {
```
vitvakatu: Technically this is a breaking change. As an alternative I can add a separate method for returning `NativeHandle`. Would it be better?

exonum-java-binding/core/rust/src/storage/temporarydb.rs
```diff
@@ -73,6 +73,8 @@ pub extern "system" fn Java_com_exonum_binding_core_storage_database_MemoryDb_na
 }
 
 /// Merges the given fork into the database.
+/// Invalidates provided `view_handle` by replacing it with new Fork. `view_handle` must be
+/// freed immediately by Java side.
```
dmitry-timofeev: Does 'immediately' matter?

exonum-java-binding/core/rust/src/storage/temporarydb.rs
```diff
@@ -81,14 +83,12 @@ pub extern "system" fn Java_com_exonum_binding_core_storage_database_MemoryDb_na
...
-            ViewRef::Fork(fork) => fork,
-        };
-        //FIXME: Implement merging via db.merge(fork.into_patch())
-        //        db.merge(fork.patch().clone())
-        //            .expect("Unable to merge fork");
+        let db = handle::cast_handle::<TemporaryDB>(db_handle);
+        let view_ref = handle::cast_handle::<View>(view_handle);
+        let view = std::mem::replace(view_ref, View::from_owned_fork(db.fork()));
+        let fork = view.into_fork();
+        db.merge(fork.into_patch())
```
dmitry-timofeev: Does Java destroy all collections created with the fork prior to this operation? Or it isn't necessary?

exonum-java-binding/core/rust/src/storage/temporarydb.rs
```diff
@@ -81,14 +83,12 @@ pub extern "system" fn Java_com_exonum_binding_core_storage_database_MemoryDb_na
...
-            ViewRef::Snapshot(_) => panic!("Attempt to merge snapshot instead of fork."),
-            ViewRef::Fork(fork) => fork,
-        };
-        //FIXME: Implement merging via db.merge(fork.into_patch())
-        //        db.merge(fork.patch().clone())
-        //            .expect("Unable to merge fork");
+        let db = handle::cast_handle::<TemporaryDB>(db_handle);
+        let view_ref = handle::cast_handle::<View>(view_handle);
+        let view = std::mem::replace(view_ref, View::from_owned_fork(db.fork()));
+        let fork = view.into_fork();
```
dmitry-timofeev: In what kind of state does the system end up if we panic in `into_fork`?

exonum-java-binding/core/rust/src/storage/temporarydb.rs
```diff
@@ -81,14 +83,12 @@ pub extern "system" fn Java_com_exonum_binding_core_storage_database_MemoryDb_na
...
-        let _fork = match *handle::cast_handle::<View>(view_handle).get() {
-            ViewRef::Snapshot(_) => panic!("Attempt to merge snapshot instead of fork."),
-            ViewRef::Fork(fork) => fork,
-        };
-        //FIXME: Implement merging via db.merge(fork.into_patch())
-        //        db.merge(fork.patch().clone())
-        //            .expect("Unable to merge fork");
+        let db = handle::cast_handle::<TemporaryDB>(db_handle);
+        let view_ref = handle::cast_handle::<View>(view_handle);
+        let view = std::mem::replace(view_ref, View::from_owned_fork(db.fork()));
```
dmitry-timofeev: Is it guaranteed not to cause cast exceptions?

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/database/MemoryDb.java
```diff
@@ -71,11 +72,15 @@ public Fork createFork(Cleaner cleaner) {
 
   /**
    * Applies the changes from the given fork to the database state.
+   * Invalidates provided fork, as it can't be used after merge.
    *
    * @param fork a fork to get changes from
    */
   public void merge(Fork fork) {
-    nativeMerge(getNativeHandle(), fork.getViewNativeHandle());
+    NativeHandle forkNativeHandle = fork.getViewNativeHandle();
```
dmitry-timofeev: I think it would be better if the fork itself performed the operation of converting itself
into a patch.

The problems with the code:
  - It frees the fork handle using unrelated free method for the MemoryDb, which will likely crash.
  - There is unclear interference with the Fork Cleaner.
  - The indices created with this fork remain live:
     - They are not destroyed before Fork (they will be after, which might cause a crash)
     - They are not invalidated — allowing subsequent operations (which will definitely cause a crash).
  - I think each type must free its own handles, and destroy other proxies through their public API.

vitvakatu:
Some garbage commits because of tricky rebase, but the changes are ok

exonum-java-binding/core/rust/src/storage/fork.rs
```diff
@@ -0,0 +1,52 @@
...
+use handle::{self, acquire_handle_ownership, Handle};
+use storage::db::View;
+use {to_handle, utils};
+
+/// Returns true if this View can be converted into patch.
+#[no_mangle]
+pub extern "system" fn Java_com_exonum_binding_core_storage_database_Fork_nativeCanConvertIntoPatch(
+    env: JNIEnv,
+    _: JObject,
+    view_handle: Handle,
```
dmitry-timofeev: I wonder if we shall actually move these methods to `Views` in Java :thinking:

exonum-java-binding/core/rust/src/storage/fork.rs
```diff
@@ -0,0 +1,52 @@
...
+/// Converts View into patch and returns the handle to this patch.
+/// Provided `view_handle` will be cleared and can no longer be used.
+#[no_mangle]
+pub extern "system" fn Java_com_exonum_binding_core_storage_database_Fork_nativeIntoPatch(
+    env: JNIEnv,
+    _: JObject,
+    view_handle: Handle,
+) -> Handle {
+    let res = panic::catch_unwind(|| {
+        let view: Box<View> = acquire_handle_ownership(view_handle);
```
dmitry-timofeev: Given this is OK, are there any failures that might realistically occur?

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/database/Fork.java
```diff
@@ -151,7 +144,9 @@ NativeHandle intoPatch() {
   /**
    * Converts this fork into patch, consuming the object, and returns the native handle
    * to the patch.
-   * TODO(@bogdanov) Document the clean-up guarantees of this method once ^ is clarified.
+   *
+   * <p>In case of failure RuntimeException is thrown and provided nativeHandle is
```
dmitry-timofeev: :arrow_up_small:

exonum-java-binding/core/rust/src/storage/temporarydb.rs
```diff
@@ -72,23 +72,20 @@ pub extern "system" fn Java_com_exonum_binding_core_storage_database_MemoryDb_na
...
-        let _fork = match *handle::cast_handle::<View>(view_handle).get() {
-            ViewRef::Snapshot(_) => panic!("Attempt to merge snapshot instead of fork."),
-            ViewRef::Fork(fork) => fork,
-        };
-        //FIXME: Implement merging via db.merge(fork.into_patch())
-        //        db.merge(fork.patch().clone())
-        //            .expect("Unable to merge fork");
+        let db = handle::cast_handle::<TemporaryDB>(db_handle);
+        let patch = handle::acquire_handle_ownership::<Patch>(patch_handle);
+        db.merge(*patch).expect("Unable to merge fork");
```
dmitry-timofeev: Will the context from the error be included into the exception message? In which realistic cases can it fail?

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/database/MemoryDb.java
```diff
@@ -78,7 +78,8 @@ public Fork createFork(Cleaner cleaner) {
    * are closed and cannot be used anymore. Any subsequent operations on these objects will result
    * in {@link IllegalStateException}.
    *
-   * <p>TODO(@bogdanov): If the fork cannot be applied to the database …
+   * <p>If the fork cannot be applied to the database {@link RuntimeException} is
```
dmitry-timofeev: I am thinking of any realistic scenarious that might trigger this to understand how to document it and if we need to translate RuntimeException into anything more appropriate. Fork from other database?

exonum-java-binding/core/rust/src/storage/db.rs
```diff
@@ -91,6 +91,27 @@ impl View {
...
+        if let Some(view_owned) = self._owned {
```
skletsun: I'm guessing if we should remove the underscore for the `_owned` field and update its description...

exonum-java-binding/core/rust/src/storage/temporarydb.rs
skletsun: "fork"->"patch"?

exonum-java-binding/core/rust/src/storage/fork.rs
vitvakatu: Two possible ways to fail here:
1) `into_fork` conversion fails, because Java does not provide valid owned Fork handle
2) `into_patch` conversion fails because of some internal MerkleDB/RocksDB error. We can not say exactly, because it depends on their implementation. Realistically I don't think any errors are possible at this stage in MerkleDB.

exonum-java-binding/core/rust/src/storage/temporarydb.rs
vitvakatu: The error message will be included, as well as stacktrace on debug builds. It can fail because of some internal RocksDB error, so we can't prevent and recover from this.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/database/MemoryDb.java
vitvakatu: Merging fork from other database is definitely possible (however, I can't say anything about the result of such action). I think `RuntimeException` is good here given low possibility of any errors and unability to recover from them.

exonum-java-binding/core/rust/src/storage/db.rs
```diff
@@ -55,7 +55,7 @@ impl View {
         View {
             // Make a "self-reference" to a value stored in the `owned` field.
             reference: unsafe { ViewRef::from_snapshot(&*snapshot) },
-            _owned: Some(ViewOwned::Snapshot(snapshot)),
+            owned: Some(ViewOwned::Snapshot(snapshot)),
```
dmitry-timofeev: Why initialization order is not consistent with the declaration order?

exonum-java-binding/core/rust/src/storage/fork.rs
dmitry-timofeev: Good then. This method has rather limited use for Java ITs, so I don't think we have to handle errors of such small likelihood.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/database/TemporaryDb.java
```diff
@@ -81,7 +81,8 @@ public Fork createFork(Cleaner cleaner) {
    * are closed and cannot be used anymore. Any subsequent operations on these objects will result
    * in {@link IllegalStateException}.
    *
-   * <p>TODO(@bogdanov): If the fork cannot be applied to the database …
+   * <p>If the fork cannot be applied to the database {@link RuntimeException} is
+   * thrown and the provided fork is closed.
```
dmitry-timofeev:
```java
   * @throws RuntimeException if the fork cannot be applied to the database state.
   *    The provided fork will be closed
```

exonum-java-binding/core/rust/src/storage/db.rs
dmitry-timofeev: Also, see the comment in #1048 regarding the non-owned views and service runtimes.

vitvakatu: This initialization order is needed for correct borrowing usage, because we need to take pointer before moving snapshot into the `owned`

dmitry-timofeev: I see, that was unexpected :-)

PR #1026, opened 2019-07-19
Support conversion of Fork into Patch [ECR-3364]

dmitry-timofeev:
## Overview

Support conversion of Fork into Patch:

Add Fork into Patch conversion so that the changes can be applied
to the MerkleDB database. This conversion consumes the original
fork object, therefore, it is needed to support 'native peer ownership
transfer' from Java to native.

---
See:
- https://jira.bf.local/browse/ECR-3364
- #1024 

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/src/main/java/com/exonum/binding/core/proxy/CancellableCleanAction.java
+++ b/exonum-java-binding/core/src/main/java/com/exonum/binding/core/proxy/CancellableCleanAction.java
@@ -19,8 +19,6 @@ package com.exonum.binding.core.proxy;
 /**
  * A cancellable clean action can be cancelled. That is useful, for instance, when
  * a Java native proxy transfers ownership over the native peer back to the native code.
- *
- * @param <ResourceDescriptionT> the type of the resource
  */
 public interface CancellableCleanAction<ResourceDescriptionT>
     extends CleanAction<ResourceDescriptionT> {
--- a/exonum-java-binding/core/src/main/java/com/exonum/binding/core/proxy/CancellableCleanAction.java
+++ b/exonum-java-binding/core/src/main/java/com/exonum/binding/core/proxy/CancellableCleanAction.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2019 The Exonum Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License. 
+ */
+
+package com.exonum.binding.core.proxy;
+
+/**
+ * A cancellable clean action can be cancelled. That is useful, for instance, when
+ * a Java native proxy transfers ownership over the native peer back to the native code.
+ *
+ * @param <ResourceDescriptionT> the type of the resource
+ */
+public interface CancellableCleanAction<ResourceDescriptionT>
+    extends CleanAction<ResourceDescriptionT> {
+
+  /**
+   * Cancels this clean action, making {@link #clean()} a no-op. This operation cannot be reversed.
+   */
+  void cancel();
+}
--- a/exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/database/Fork.java
+++ b/exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/database/Fork.java
@@ -75,4 +75,16 @@ public final class Fork extends View {
   private Fork(NativeHandle nativeHandle, Cleaner cleaner) {
     super(nativeHandle, cleaner, new IncrementalModificationCounter(), true);
   }
+
+  /**
+   * Converts this fork into a patch that can be merged into the database.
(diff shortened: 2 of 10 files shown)


exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/database/MemoryDb.java
```diff
@@ -70,12 +71,20 @@ public Fork createFork(Cleaner cleaner) {
   }
 
   /**
-   * Applies the changes from the given fork to the database state.
+   * Applies the changes from the given fork to the database state. MemoryDb can only
+   * merge forks that {@linkplain #createFork(Cleaner) it created itself}.
+   *
+   * <p>Once this method completes, any indexes created with the fork and the fork itself
```
bullet-tooth: can we have a test that calls `merge` (or `fork.intoPatch()`) twice?

dmitry-timofeev: Added. Ideally it must throw IllegalArgument, but IllegalState in the case of rarely used method will do.

dmitry-timofeev: It is not possible to check if a Fork is valid without adding an extra package-private method.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/database/MemoryDb.java
```diff
@@ -70,12 +71,20 @@ public Fork createFork(Cleaner cleaner) {
   }
 
   /**
-   * Applies the changes from the given fork to the database state.
+   * Applies the changes from the given fork to the database state. MemoryDb can only
+   * merge forks that {@linkplain #createFork(Cleaner) it created itself}.
```
bullet-tooth:
```suggestion
   * merge forks which have been {@linkplain #createFork(Cleaner) created by it}.
```

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/database/MemoryDb.java
```diff
@@ -70,12 +71,20 @@ public Fork createFork(Cleaner cleaner) {
...
+   *
+   * <p>Once this method completes, any indexes created with the fork and the fork itself
+   * are closed and cannot be used. Any subsequent operations on these objects will result
```
bullet-tooth:
```suggestion
   * are closed and cannot be used anymore. Any subsequent operations on these objects will result
```

exonum-java-binding/core/src/main/java/com/exonum/binding/core/proxy/CancellableCleanAction.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License. 
+ */
+
+package com.exonum.binding.core.proxy;
+
+/**
+ * A cancellable clean action can be cancelled. That is useful, for instance, when
+ * a Java native proxy transfers ownership over the native peer back to the native code.
+ *
+ * @param <ResourceDescriptionT> the type of the resource
```
bullet-tooth: Can be copied from the parent as more detailed. Also, found if no params specified - they are available in IDEA quick doc by copying from the parent. But they don't appear in the generated javadocs.

As we are here: tailing comma in the parent should be deleted:  _"...of {@link java.lang.Class}, {@link String}, {@link Enum})~~,~~"_

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/database/Fork.java
```diff
@@ -58,21 +66,90 @@ public static Fork newInstance(long nativeHandle, boolean owningHandle, Cleaner
...
+  private Fork(NativeHandle nativeHandle, ProxyDestructor destructor, Cleaner forkCleaner) {
+    super(nativeHandle, forkCleaner, new IncrementalModificationCounter(), true);
+    this.destructor = destructor;
+  }
+
+  /**
+   * Converts this fork into a patch that can be merged into the database.
+   * This method will close any resources registered with {@linkplain #getCleaner() its cleaner}
+   * (typically, indexes, iterators), convert this fork into a patch, invalidate this fork,
+   * and return a handle to the patch. The caller is responsible to properly destroy
```
MakarovS: `responsible for properly destroying`?

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/database/Fork.java
```diff
@@ -58,21 +66,90 @@ public static Fork newInstance(long nativeHandle, boolean owningHandle, Cleaner
...
+   * @param destructor a destructor of the native peer, registered with the parent cleaner
+   * @param forkCleaner a cleaner for objects depending on the fork
+   */
+  private Fork(NativeHandle nativeHandle, ProxyDestructor destructor, Cleaner forkCleaner) {
+    super(nativeHandle, forkCleaner, new IncrementalModificationCounter(), true);
+    this.destructor = destructor;
+  }
+
+  /**
+   * Converts this fork into a patch that can be merged into the database.
```
MakarovS: Provide a link to patch in documentation when it's ready? https://exonum.com/doc/version/latest/architecture/merkledb/#patches

dmitry-timofeev: It is an internal method that depends on the native method producing the patch, therefore, this documentation is for us. It is probably faster to navigate to the Patch declaration in the native code which might have better documentation that the website, but the link won't hurt.

PR #1032, opened 2019-07-22
New proofs hashing rules for ProofMaps [ECR-3320]

MakarovS:
## Overview
New proofs hashing rules for ProofMaps.

---
See: https://jira.bf.local/browse/ECR-3320

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/map/UncheckedFlatMapProof.java
```diff
@@ -250,22 +256,35 @@ private CheckedMapProof checkProof() {
   }
 
   private static HashCode getEmptyProofListHash() {
```
dmitry-timofeev: not a list

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/map/UncheckedFlatMapProof.java
```diff
@@ -250,22 +256,35 @@ private CheckedMapProof checkProof() {
...
+  }
+
+  private static HashCode getMapProofRootHash(HashCode merkleRoot) {
+    return HASH_FUNCTION.newHasher()
+        .putByte(MAP_NODE_PREFIX)
+        .putObject(merkleRoot, hashCodeFunnel())
+        .hash();
   }
 
   private static HashCode getSingleEntryRootHash(DbKey key, HashCode valueHash) {
```
dmitry-timofeev: This patch changes the meaning of the two previous and this method: the former calculate one thing, this one — another. Therefore, they no longer can have the same name.

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/map/UncheckedFlatMapProof.java
```diff
@@ -250,22 +256,35 @@ private CheckedMapProof checkProof() {
...
     assert key.getNodeType() == Type.LEAF;
     return HASH_FUNCTION.newHasher()
+        .putByte(MAP_BRANCH_NODE_PREFIX)
```
dmitry-timofeev: Is it correct (= a leaf node having a **branch** prefix)?

pavel-mukhanov: No, as we discussed in wiki, for leaf we use **blob** prefix.

MakarovS:
>I think the terminology must be improved:
What is now considered a 'root hash'?
What is considered to be a 'Merkle root'?
Are these terms well-defined in this context?

I'd use `rootHash` for the root hashes of the trees and then `proofMapHash` for the resulting hash (`proofMapHash = MapNode || rootHash`)

dmitry-timofeev:
> I'd use rootHash for the root hashes of the trees and then proofMapHash for the resulting hash (proofMapHash = MapNode || rootHash)

Why don't we make this code consistent with the terminology we established in the wiki (merkle_root/object_hash): https://wiki.bf.local/display/EXN/MerkleDB+changelog?focusedCommentId=31294978#comment-31294978

MakarovS:
> Why don't we make this code consistent with the terminology we established in the wiki (merkle_root/object_hash): https://wiki.bf.local/display/EXN/MerkleDB+changelog?focusedCommentId=31294978#comment-31294978

I actually prefer `proofMapHash` to `objectHash` (and `rootHash` to `merkleRoot` too). But if you think we should change that, I will.

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/map/UncheckedFlatMapProof.java
```diff
@@ -51,6 +52,13 @@
 
   private final List<ByteString> missingKeys;
 
+  @VisibleForTesting
+  static final byte BLOB_PREFIX = 0x00;
+  @VisibleForTesting
+  static final byte MAP_NODE_PREFIX = 0x03;
+  @VisibleForTesting
+  static final byte MAP_BRANCH_NODE_PREFIX = 0x04;
```
dmitry-timofeev: These names are misleading: _branch_ prefix is applied to both _leaf_ and branch nodes; whilst _node_ prefix is actually applied to compute the map hash (= it is not applied to nodes, but to 'merkle root').

MAP_NODE_PREFIX -> MAP_OBJECT_PREFIX/MAP_ROOT_PREFIX and 
MAP_BRANCH_NODE_PREFIX -> MAP_NODE_PREFIX?

I also suggest improving the naming upstream (cc @pavel-mukhanov , @aleksuss ).

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/map/UncheckedFlatMapProof.java
```diff
@@ -249,35 +257,53 @@ private CheckedMapProof checkProof() {
     return commonPrefix;
   }
 
-  private static HashCode getEmptyProofListHash() {
-    return HashCode.fromBytes(new byte[Hashing.DEFAULT_HASH_SIZE_BYTES]);
+  private static HashCode getEmptyProofMapHash() {
+    return HASH_FUNCTION.newHasher()
```
dmitry-timofeev:
```java
    HashCode merkleRoot = HashCode.fromBytes(new byte[Hashing.DEFAULT_HASH_SIZE_BYTES]);
    return getProofMapHash(merkleRoot);
```

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/map/UncheckedFlatMapProof.java
```diff
@@ -249,35 +257,53 @@ private CheckedMapProof checkProof() {
...
 
-  private static HashCode getSingleEntryRootHash(DbKey key, HashCode valueHash) {
+  private static HashCode getProofMapHash(HashCode rootHash) {
+    return HASH_FUNCTION.newHasher()
+        .putByte(MAP_NODE_PREFIX)
+        .putObject(rootHash, hashCodeFunnel())
+        .hash();
+  }
+
+  private static HashCode getSingleEntryRootProofMapHash(DbKey key, HashCode valueHash) {
```
dmitry-timofeev: Too long name. `getSingleEntryRootHash/MerkleRoot` (since we are inside MapProof class)?

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/map/UncheckedFlatMapProof.java
dmitry-timofeev: @pavel-mukhanov @MakarovS Is the wiki currently wrong or this code? Shall leafNodeHash be:

(1) as this code does:
```
valueHash = sha256(BLOB || value)
leafNodeHash = sha256(BRANCH || key || valueHash)
```

or (2) as the comment and wiki suggest:
```
leafNodeHash = sha256(BLOB || value)
```

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/map/UncheckedFlatMapProof.java
dmitry-timofeev: Also, why doesn't the code use the newer 'merkleRoot' term?

dmitry-timofeev:
> I actually prefer proofMapHash to objectHash (and rootHash to merkleRoot too). But if you think we should change that, I will.

We can consider the naming separately. objectHash might be a little confusing for Java users (because there is `Object` a superclass, any instance is also an 'object'), + MerkleDB has indexes (aka collections), not objects.

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/map/UncheckedFlatMapProof.java
pavel-mukhanov: Sorry, I'm confused you, **leaf** value should be hashed with blob prefix, **leaf node** with branch prefix. The code is right. I should change wiki once again.

dmitry-timofeev: Good, thanks!

dmitry-timofeev: @MakarovS , @pavel-mukhanov , 

Currently the code seems to compute the leafNodeHash as (2), is that correct? See https://github.com/exonum/exonum-java-binding/blob/f9f15066923a644ebcd00c3056abc45d0264f8df/exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/map/UncheckedFlatMapProof.java#L231

MakarovS: It seems correct, according to original code - https://github.com/exonum/exonum/blob/master/components/merkledb/src/proof_map_index/proof.rs#L561
@pavel-mukhanov please correct me if I'm wrong

pavel-mukhanov: It's correct.

dmitry-timofeev:
As #1044 has landed, I think the ITs for ProofMap must be re-enabled.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxy.java
```diff
@@ -285,15 +285,15 @@ private UncheckedMapProof getMultiKeyProof(Collection<? extends K> keys) {
   private native UncheckedMapProof nativeGetMultiProof(long nativeHandle, byte[] keys);
 
   /**
-   * Returns the root hash of the underlying Merkle-Patricia tree.
+   * Returns the index hash of this map.
```
dmitry-timofeev: I'd clarify what an index hash is (it represents the complete state of this map. Any modifications to the stored entries affect the index hash).

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxy.java
```diff
@@ -285,15 +285,15 @@ private UncheckedMapProof getMultiKeyProof(Collection<? extends K> keys) {
...
    *
    * @throws IllegalStateException  if this map is not valid
    */
-  public HashCode getRootHash() {
-    return HashCode.fromBytes(nativeGetRootHash(getNativeHandle()));
+  public HashCode getIndexHash() {
```
dmitry-timofeev: On top of 'object hash' and 'index/collection hash', we might name the thing 'state hash' — we already use the term in Service/Schema APIs.

exonum-java-binding/core/src/test/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxyIntegrationTest.java
```diff
@@ -208,20 +208,20 @@ void get() {
   }
 
   @Test
-  void getRootHash_EmptyMap() {
+  void getIndexHash_EmptyMap() {
     runTestWithView(database::createSnapshot,
-        (map) -> assertThat(map.getRootHash(), equalTo(EMPTY_MAP_ROOT_HASH)));
+        (map) -> assertThat(map.getIndexHash(), equalTo(EMPTY_MAP_ROOT_HASH)));
```
dmitry-timofeev: Shan't the constant name be updated?

MakarovS:
@pavel-mukhanov please review https://github.com/exonum/exonum-java-binding/pull/1032/commits/47b7e3e6e839fc4019f54728494b8f55c9b3534e
It's based on core code - https://github.com/exonum/exonum/blob/master/components/merkledb/src/proof_map_index/key.rs#L430

pavel-mukhanov:
@MakarovS looks good

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/map/UncheckedFlatMapProof.java
```diff
@@ -249,42 +257,114 @@ private CheckedMapProof checkProof() {
...
-        .putObject(rightChild.getDbKey(), dbKeyFunnel())
+        .putBytes(leftChildKey)
+        .putBytes(rightChildKey)
         .hash();
   }
 
+  /**
+   * Returns a LEB128 compressed binary representation of the given db key.
+   */
+  public static byte[] compress(DbKey dbKey) {
```
dmitry-timofeev: Why isn't it in a funnel so that it is more easily testable?

On top of that, it will provide easier API to write the things (putByte/putBytes, similar to ByteBuffer, instead of numerous array copies and keeping track of the number of written bytes manually).

Please also test this.

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/map/UncheckedFlatMapProof.java
```diff
@@ -249,42 +257,114 @@ private CheckedMapProof checkProof() {
...
+    byte[] buffer = new byte[Hashing.DEFAULT_HASH_SIZE_BYTES * 2];
```
dmitry-timofeev: (not needed with a funnel) The choice of the constant looks random: it seems to write up to `<Hash_size> + (?) 2 bytes = 34` for the 'compressed' size (if 2 bytes are required to represent 256 significant bits).

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/map/UncheckedFlatMapProof.java
```diff
@@ -249,42 +257,114 @@ private CheckedMapProof checkProof() {
...
+    int bytesWritten = writeCompressed(buffer, dbKey);
+    byte[] result = new byte[bytesWritten];
```
dmitry-timofeev: (not needed with a funnel) `return Arrays.copyOf(buffer, bytesWritten);`

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/map/UncheckedFlatMapProof.java
```diff
@@ -249,42 +257,114 @@ private CheckedMapProof checkProof() {
...
+   * Returns a number of bytes written into buffer which contains LEB128 compressed binary
+   * representation of the given db key.
+   */
+  private static int writeCompressed(byte[] buffer, DbKey dbKey) {
+    int bitsLength = dbKey.getNumSignificantBits();
+    // Perform division, rounding the result up
+    int wholeBytesLength = (bitsLength + Byte.SIZE - 1) / Byte.SIZE;
+    byte[] key = dbKey.getKeySlice();
+
+    int bytesWritten = writeUnsignedLeb128(buffer, bitsLength);
```
dmitry-timofeev: I find this code/documentation rather confusing. We write up to 9 bits of information (because bitsLength is from 0 to 256) using writeUnsignedLeb128 with a fancy loop.
 Does this method only compress the number of significant bits, as it then just copies the key as is?

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/map/UncheckedFlatMapProof.java
```diff
@@ -249,42 +257,114 @@ private CheckedMapProof checkProof() {
...
+    System.arraycopy(key, 0, buffer, bytesWritten, key.length);
```
dmitry-timofeev: Why copy all `key.length = 32 bytes` if wholeBytesLength is anywhere between 0 bytes and 32 bytes?

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/map/UncheckedFlatMapProof.java
```diff
@@ -249,42 +257,114 @@ private CheckedMapProof checkProof() {
...
+    bytesWritten += wholeBytesLength;
+    int bitsInLastByte = bitsLength % Byte.SIZE;
+    if (wholeBytesLength > 0 && bitsInLastByte != 0) {
+      resetBits(buffer, bytesWritten, bitsInLastByte);
```
dmitry-timofeev: Why is this needed at all if you copied proper key bit slice byte-by-byte? DbKey constructor ensures that no bits are set after numSignificantBits.

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/map/UncheckedFlatMapProof.java
MakarovS: I believe so, yes. I think @alekseysidorov could provide more details as the author of the original algorithm

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/DbKeyCompressedFunnel.java
```diff
@@ -0,0 +1,55 @@
...
+    byte[] key = from.getKeySlice();
+    into.putBytes(key, 0, wholeBytesLength);
+  }
+
+  public static Funnel<DbKey> dbKeyCompressedFunnel() {
+    return INSTANCE;
+  }
+
+  private static void writeUnsignedLeb128(PrimitiveSink into, int value) {
+    int remaining = value >>> 7;
```
bullet-tooth: some serious stuff is going here 🧐

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/CheckedProof.java
```diff
@@ -33,11 +33,11 @@
...
 
   /**
-   * Returns the calculated root hash of the proof.
-   * Must be equal to the Merkle root hash of the collection, providing this proof.
+   * Returns the calculated index hash of the proof.
+   * Must be equal to the index hash of the collection, providing this proof.
    * @throws IllegalStateException if the proof is not valid
    */
-  HashCode getRootHash();
+  HashCode getIndexHash();
```
bullet-tooth: I believe it would be great to notice such changes somewhere for the future "migration guide".

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/DbKeyCompressedFunnel.java
dmitry-timofeev: Nah, it seems to be almost identical to the varint encoding used in protobuf and some of our serializers.

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/DbKeyCompressedFunnel.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+package com.exonum.binding.common.proofs;
+
+import com.exonum.binding.common.hash.Funnel;
+import com.exonum.binding.common.hash.PrimitiveSink;
+import com.exonum.binding.common.proofs.map.DbKey;
+
+/**
+ * A funnel for a database key. Puts the LEB128 compressed binary representation of the given
```
dmitry-timofeev: The wording is misleading — LEB128 is not applied to the whole key, but to the key slice size only (or what is the current term of the size of the significant part of the key).

dmitry-timofeev: Also we might say 'encodes' instead of compresses, as it 'compresses' anything greater than 127 into two bytes.

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/map/UncheckedFlatMapProof.java
```diff
@@ -249,39 +258,55 @@ private CheckedMapProof checkProof() {
...
+        .putByte(MAP_ROOT_PREFIX)
+        .putObject(merkleRoot, hashCodeFunnel())
+        .hash();
+  }
+
+  private static HashCode getSingleEntryMerkleRoot(DbKey key, HashCode valueHash) {
     assert key.getNodeType() == Type.LEAF;
     return HASH_FUNCTION.newHasher()
+        .putByte(MAP_NODE_PREFIX)
         .putObject(key, dbKeyFunnel())
```
dmitry-timofeev: Is that correct that it does not compress here? If yes, is this inconsistency worth it (in the core and all the clients), is it somehow justified?

MakarovS: Yes, I also noticed this inconsistency, but it's like that in core - https://github.com/exonum/exonum/blob/master/components/merkledb/src/hash.rs#L143
@pavel-mukhanov is there a reason why database key path isn't compressed in case of a single entry proof?

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/CheckedProof.java
MakarovS: Updated the changelog

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/DbKeyCompressedFunnel.java
```diff
@@ -0,0 +1,73 @@
...
+    return INSTANCE;
+  }
+
+  @VisibleForTesting
+  static int getWholeBytesKeyLength(int bitsLength) {
+    return (bitsLength + Byte.SIZE - 1) / Byte.SIZE;
+  }
+
+  private static void writeUnsignedLeb128(PrimitiveSink into, int value) {
+    // As we encode number of significant bits in a database key which is [0; 256], three bytes
```
dmitry-timofeev: Why? 256 takes 9 bits: 100000000. We encode 7 bits of the original value in a byte, therefore, it cannot take more than _ceil(9/7) = two bytes_.

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/DbKeyCompressedFunnel.java
```diff
@@ -0,0 +1,73 @@
...
+    // are needed
+    byte[] buffer = new byte[3];
```
dmitry-timofeev: The previous version writing to a sink seemed to be more straightforward (it did not need an extra array, or counting the number of bytes written). Why adding an extra array?

exonum-java-binding/common/src/test/java/com/exonum/binding/common/proofs/DbKeyCompressedFunnelTest.java
```diff
@@ -0,0 +1,55 @@
...
+  void funnelTest(DbKey dbKey) {
+    PrimitiveSink primitiveSink = mock(PrimitiveSink.class);
+    DbKeyCompressedFunnel.dbKeyCompressedFunnel().funnel(dbKey, primitiveSink);
+    int wholeBytesKeyLength = getWholeBytesKeyLength(dbKey.getNumSignificantBits());
+    byte[] encodedSignificantBitsNum = new byte[3];
+    int bytesWritten = writeUnsignedLeb128(encodedSignificantBitsNum,
+        dbKey.getNumSignificantBits());
+    byte[] key = dbKey.getKeySlice();
+
+    verify(primitiveSink).putBytes(encodedSignificantBitsNum, 0, bytesWritten);
```
dmitry-timofeev: This test does not seem to test much. 

It uses the private methods *of the SuT* to get the *expected* values. How can they be different from what the `#funnel` produces if it uses the same methods? It is basically equivalent, if testing is considered, to the following:

```java
verify(primitiveSink).putBytes(any(byte[].class), 0, anyInt());
verify(primitiveSink).putBytes(key, 0, anyInt());
```

exonum-java-binding/common/src/test/java/com/exonum/binding/common/proofs/DbKeyCompressedFunnelTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+
+class DbKeyCompressedFunnelTest {
+
+  @ParameterizedTest
+  @MethodSource("testSource")
+  void funnelTest(DbKey dbKey, byte[] encodedSignificantBitsNum) {
+    PrimitiveSink primitiveSink = mock(PrimitiveSink.class);
+    DbKeyCompressedFunnel.dbKeyCompressedFunnel().funnel(dbKey, primitiveSink);
+    int wholeBytesKeyLength = getWholeBytesKeyLength(dbKey.getNumSignificantBits());
```
dmitry-timofeev: Why getWholeBytesKeyLength is still used in the test to get the **expected** value or not tested separately?

exonum-java-binding/common/src/test/java/com/exonum/binding/common/proofs/DbKeyCompressedFunnelTest.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  private static Stream<Arguments> testSource() {
+    return Stream.of(
+        Arguments.of(DbKey.newBranchKey(keyFromString(""), 0), new byte[]{0}),
+        Arguments.of(DbKey.newBranchKey(keyFromString("1"), 1), new byte[]{1}),
+        Arguments.of(DbKey.newBranchKey(keyFromString("111"), 7), new byte[]{7}),
+        Arguments.of(DbKey.newBranchKey(keyFromString("0001"), 8), new byte[]{8}),
+        Arguments.of(DbKey.newBranchKey(keyFromString("1111"), 127), new byte[]{127}),
+        Arguments.of(DbKey.newBranchKey(keyFromString("1001 1001"), 128), new byte[]{-128, 1}),
```
dmitry-timofeev: As it is a test of LEB128, which is a binary encoding, I suggest making it easier to verify the correctness of test data by using binary grouped by 7 bits (as LEB128 uses):
 * `128:  0b1_0000000 -> 1_0000000, 0_0000001`;
 * `256: 0b10_0000000 -> 1_0000000, 0_0000010`

exonum-java-binding/common/src/test/java/com/exonum/binding/common/proofs/DbKeyCompressedFunnelTest.java
```diff
@@ -0,0 +1,60 @@
...
+  @MethodSource("testSource")
+  void funnelTest(DbKey dbKey, byte[] encodedSignificantBitsNum) {
+    PrimitiveSink primitiveSink = mock(PrimitiveSink.class);
+    DbKeyCompressedFunnel.dbKeyCompressedFunnel().funnel(dbKey, primitiveSink);
+    int wholeBytesKeyLength = getWholeBytesKeyLength(dbKey.getNumSignificantBits());
+    byte[] key = dbKey.getKeySlice();
+
+    verify(primitiveSink, times(encodedSignificantBitsNum.length)).putByte(anyByte());
+    for (byte encodedByte : encodedSignificantBitsNum) {
+      verify(primitiveSink).putByte(encodedByte);
```
dmitry-timofeev: Please use `InOrder` for all interactions. Usual verifications do not verify the order of operations (i.e., the test will pass if the order is wrong). `times` is redundant then.

PR #1036, opened 2019-07-22
LC: fix code smells

bullet-tooth:
which actually should never happen

## Overview

---
Added check for preventing forever loop which actually should never happen

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-light-client/src/main/java/com/exonum/client/ExonumHttpClient.java
+++ b/exonum-light-client/src/main/java/com/exonum/client/ExonumHttpClient.java
@@ -32,6 +32,7 @@ import static com.exonum.client.request.BlockTimeOption.INCLUDE_COMMIT_TIME;
 import static com.exonum.client.request.BlockTimeOption.NO_COMMIT_TIME;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
 import static java.lang.Math.max;
 import static java.lang.Math.min;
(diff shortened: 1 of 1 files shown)


exonum-light-client/src/main/java/com/exonum/client/ExonumHttpClient.java
```diff
@@ -166,7 +167,10 @@ public BlockResponse getBlockByHeight(long height) {
 
       blocks.addAll(blocksResponse.getBlocks());
 
-      rangeLast = blocksResponse.getBlocksRangeStart() - 1;
+      long blocksRangeStart = blocksResponse.getBlocksRangeStart();
+      checkState(blocksRangeStart <= rangeLast,
```
dmitry-timofeev: This code distrusts the Exonum API, doesn't it?

bullet-tooth: Why does it distract? It checks theoretically impossible case to prevent forever loop.

dmitry-timofeev: If API is hacked this library might have even bigger problems, isn't it?

bullet-tooth: definitely yes. Do you think this check is redundant?

dmitry-timofeev: If we assume we can rely on Exonum returning proper responses, then, probably, yes.

PR #1037, opened 2019-07-22
Evaluate Truth assertion framework

dmitry-timofeev:
## Overview

The PR is better viewed commit-by-commit. It starts with a singe test migrated to Truth, adds a minimal `Subject` (~ Hamcrest Matcher), then extends that subject with a new operation, and finally adds some assertions on common library types like `String`s and `Throwable`s and various collections.

---
See: https://jira.bf.local/browse/ECR-3339

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-light-client/pom.xml
+++ b/exonum-light-client/pom.xml
@@ -152,6 +152,18 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>com.google.truth</groupId>
+      <artifactId>truth</artifactId>
+      <version>1.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.truth.extensions</groupId>
+      <artifactId>truth-java8-extension</artifactId>
+      <version>1.0</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/exonum-light-client/src/test/java/com/exonum/client/CollectionSubjectsTest.java
+++ b/exonum-light-client/src/test/java/com/exonum/client/CollectionSubjectsTest.java
@@ -1,0 +1,132 @@
+/*
+ * Copyright 2019 The Exonum Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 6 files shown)


bullet-tooth:
Looks nice.
Will it bring benefits for testing our collections? 
For example, is it hard to make such assertions (or similar):
```
assertThat(listIndexProxy).isEmpty();
assertThat(listIndexProxy).hasSize(1);
assertThat(listIndexProxy).containsExactly("one", "two");
```

dmitry-timofeev:
Thanks for the review,

> Will it bring benefits for testing our collections?

The good thing is, [`IterableSubject`](https://github.com/google/truth/blob/master/core/src/main/java/com/google/common/truth/IterableSubject.java#L91) is already applicable to our List and Set indexes, and it is non-final, therefore, can be extended to add some specific assertions (proofs, hashes, etc.). The assertions you've seen in the PR for `List`s use IterableSubject (= there is no `ListSubject` in Truth).

Regarding Maps, I currently see several ways how to do that:
1. A full-blown implementation of a custom Subject for our Map indexes (which are *not* Maps from the Java Collection Framework).
    - For a proxy of the efforts, required to implement them, one can see the appropriate Truth subjects: for example, [MapSubject](https://github.com/google/truth/blob/master/core/src/main/java/com/google/common/truth/MapSubject.java).
2. Use a hack and rely on the MapSubject. It can be done in several ways: 
    - either by having an explicit method converting MapIndex to a Map: `assertThat(exonumMapIndex).asMap().containsEntry(k, v)`;
    - or by extending the MapSubject and performing this conversion in the constructor: `assertThat(exonumMapIndex).containsEntry(k, v)`.

The big advantage of the second approach is that there is way less operations to implement (test, document) and maintain. There are a couple of disadvantages too: (a) unless we create an MapIndexToMapAdapter, they will always convert the MapIndex into Map in O(N) even for assertions using O(1) operations; (b) MapIndex is not a Map, therefore, some operations have different signatures (e.g., `size -> long` instead of `int`). I think (a) is negligible in usual tests; (b), though a valid concern and must be evaluated more carefully by inspecting operations in the MapSubject and Map indexes, does not seem to be a big issue with the present interface of our map indexes being very similar to Maps.

bullet-tooth:
Looks nice.
>  they will always convert the MapIndex into Map in O(N) even for assertions using O(1) operations

I believe it's not a problem because usually there're not so many elements in test collections.

dmitry-timofeev:
Exactly, that's why I considered the overhead as "negligible" 🙃

dmitry-timofeev:
There is something for a Testkit Px for sure.

dmitry-timofeev:
Added https://jira.bf.local/browse/ECR-3377

PR #1047, opened 2019-07-30
Relax index modification restrictions [ECR-3322]:

dmitry-timofeev:
## Overview

Allow to iterate over one index and modify any other independent
index as this is determined to be safe with MerkleDB.
Modification of the index during iteration over it (even if it
happens through separate proxies) is still prohibited and must
be reliably detected. As indexes are de-duplicated now (#1039), no extra
effort is required to ensure such reliable detection (except the tests).

---
See: 
- https://jira.bf.local/browse/ECR-3322
- https://jira.bf.local/browse/ECR-3359 (de-duplication)

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/database/Fork.java
+++ b/exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/database/Fork.java
@@ -88,7 +88,7 @@ public final class Fork extends View {
    * @param forkCleaner a cleaner for objects depending on the fork
    */
   private Fork(NativeHandle nativeHandle, ProxyDestructor destructor, Cleaner forkCleaner) {
-    super(nativeHandle, forkCleaner, new IncrementalModificationCounter(), true);
+    super(nativeHandle, forkCleaner, true);
     this.destructor = destructor;
   }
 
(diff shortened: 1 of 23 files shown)


bullet-tooth:
`ListSpliterator` doc says that "It will fail if any collection associated with a given Fork is modified".
But looks like it should work like Iterator after the changes.
Won't the following test succeed?
```
  @Test
  void spliteratorTest() throws CloseFailuresException {
    try (Cleaner cleaner = new Cleaner()) {
      Fork view = database.createFork(cleaner);
      List<String> elements = TestStorageItems.values;
      // Create two independent lists
      ListIndex<String> l1 = create("List1", view);
      ListIndex<String> l2 = create("List123", view);
      l1.addAll(elements);

      // Take the spliterator from the first list and modify the other list
      l1.stream().forEach(l2::add);

      assertThat(l1.size(), equalTo(l2.size()));
    }
  }
```

dmitry-timofeev:
Thanks, yes, it must work as the iterator (your test must succeed), I'll update the docs appropriately.

bullet-tooth:
> Thanks, yes, it must work as the iterator (your test must succeed), I'll update the docs appropriately.

And add such test for Spliterator as well?

dmitry-timofeev:
I don't think it is needed as we effectively test that each collection has its own `ModificationCounter` with the existing test) Spliterator is unit-tested to interact with ModificationCounter properly.

dmitry-timofeev:
Hope that's it with the docs :-)

bullet-tooth:
there are still 2 points:
- changelog
- `ModificationCounter` now can be a package-private?

dmitry-timofeev:
Yes, it can be, thanks. The changelog on MerkleDB — separately as I've written in other issue.

PR #1053, opened 2019-08-01
Update ListProof hashing rules [ECR-3318]

MakarovS:
## Overview
Update ListProof hashing rules.
Should we add `length` field to `ListProofBranch` as a way to pass length of the index?
WIP: length is always `1` when calculating hash right now

---
See: https://jira.bf.local/browse/ECR-3318

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/list/ListMerkleRootCalculator.java
+++ b/exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/list/ListMerkleRootCalculator.java
@@ -55,28 +55,26 @@ final class ListMerkleRootCalculator<E> implements ListProofVisitor {
 
   private HashCode merkleRoot;
 
-  private long length;
-
   /**
    * Creates a new ListMerkleRootCalculator.
    *
    * @param serializer a serializer of list elements
    */
-  ListMerkleRootCalculator(ListProofNode listProof, Serializer<E> serializer) {
+  ListMerkleRootCalculator(ListProofRoot listProof, Serializer<E> serializer) {
     this(listProof, serializer, Hashing.defaultHashFunction());
   }
 
-  private ListMerkleRootCalculator(ListProofNode listProof, Serializer<E> serializer,
+  private ListMerkleRootCalculator(ListProofRoot listProof, Serializer<E> serializer,
                                    HashFunction hashFunction) {
     this.serializer = CheckingSerializerDecorator.from(serializer);
     this.hashFunction = checkNotNull(hashFunction);
     elements = new TreeMap<>();
     index = 0;
-    // TODO: fix (get length from native)
-    length = 1;
+    long length = listProof.getLength();
     merkleRoot = null;
 
-    listProof.accept(this);
+    ListProofNode rootNode = listProof.getRootNode();
+    rootNode.accept(this);
 
     merkleRoot = hashFunction.newHasher()
         .putByte(LIST_ROOT_PREFIX)
--- a/exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/list/ListMerkleRootCalculator.java
+++ b/exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/list/ListMerkleRootCalculator.java
@@ -25,6 +25,8 @@ import com.exonum.binding.common.hash.Hashing;
 import com.exonum.binding.common.hash.PrimitiveSink;
 import com.exonum.binding.common.serialization.CheckingSerializerDecorator;
 import com.exonum.binding.common.serialization.Serializer;
+import com.google.common.annotations.VisibleForTesting;
+
 import java.util.NavigableMap;
 import java.util.Optional;
 import java.util.TreeMap;
@@ -34,7 +36,14 @@ import java.util.TreeMap;
  *
  * @param <E> the type of elements in the corresponding list
  */
-final class ListProofRootHashCalculator<E> implements ListProofVisitor {
+final class ListMerkleRootCalculator<E> implements ListProofVisitor {
+
+  @VisibleForTesting
+  static final byte BLOB_PREFIX = 0x00;
+  @VisibleForTesting
+  static final byte LIST_BRANCH_PREFIX = 0x01;
+  @VisibleForTesting
+  static final byte LIST_ROOT_PREFIX = 0x02;
 
   private final CheckingSerializerDecorator<E> serializer;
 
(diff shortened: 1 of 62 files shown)


exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/list/ListMerkleRootCalculator.java
```diff
@@ -34,7 +36,14 @@
  *
  * @param <E> the type of elements in the corresponding list
  */
-final class ListProofRootHashCalculator<E> implements ListProofVisitor {
+final class ListMerkleRootCalculator<E> implements ListProofVisitor {
```
dmitry-timofeev: Update the docs ^

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/list/ListMerkleRootCalculator.java
```diff
@@ -44,26 +53,33 @@
...
     elements = new TreeMap<>();
     index = 0;
-    calculatedRootHash = null;
+    merkleRoot = null;
+
+    ListProofNode rootNode = listProof.getRootNode();
+    rootNode.accept(this);
 
-    listProof.accept(this);
+    merkleRoot = hashFunction.newHasher()
```
dmitry-timofeev: It turns out to be **not** the Merkle root, but the hash of the collection (what is the term for that?).

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/list/ListMerkleRootCalculator.java
```diff
@@ -120,14 +140,14 @@ private HashCode visitLeft(ListProofBranch branch, long parentIndex) {
    *
    * @return hash code
```
dmitry-timofeev: The documentation is not complete + contains redundant @return

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/list/ListProof.java
```diff
@@ -0,0 +1,33 @@
+package com.exonum.binding.common.proofs.list;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Represents a proof that some elements exist in a ProofList at certain positions and the length
+ * of corresponding index.
+ */
+public class ListProof {
```
dmitry-timofeev: final

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/list/UncheckedListProof.java
```diff
@@ -30,5 +30,5 @@
   /**
    * Returns raw source proof of this UncheckedListProof.
    */
-  ListProofNode getRootProofNode();
+  ListProof getListProof();
```
dmitry-timofeev: I don't quite get what is going on here. Why isn't ListProof UncheckedListProof,
  but is contained in it?

exonum-java-binding/common/src/test/java/com/exonum/binding/common/proofs/list/ListMerkleRootCalculatorTest.java
```diff
@@ -56,14 +60,16 @@ void visit_FullProof2elements() {
     ListProofElement right = leafOf(V2);
     ListProofBranch root = new ListProofBranch(left, right);
 
-    calculator = createListProofCalculator(root);
+    ListProof listProof = new ListProof(root, 2);
+    calculator = createListProofCalculator(listProof);
 
-    //calculate expected root hash
+    // Calculate expected merkle root
```
dmitry-timofeev: Here and elsewhere, please use consistent terms: merkle root for the hash of the tree, and (? — don't remember) hash for the hash of the collection.

exonum-java-binding/core/rust/src/storage/proof_list_index.rs
```diff
@@ -369,6 +369,19 @@ pub extern "system" fn Java_com_exonum_binding_core_storage_indices_ProofListInd
     handle::drop_handle::<ProofListIndexIter<Value>>(&env, iter_handle);
 }
 
+fn make_java_proof_root<'a>(
+    env: &JNIEnv<'a>,
+    proof: &ListProof<Value>,
```
dmitry-timofeev: @vitvakatu  Why doesn't ListProof in Rust contain the length?

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/list/ListMerkleRootCalculator.java
pavel-mukhanov: It's `hash` of the ProofList object in MerkleDB.

exonum-java-binding/core/rust/src/storage/proof_list_index.rs
vitvakatu: I have no idea https://github.com/exonum/exonum/blob/master/components/merkledb/src/proof_list_index/proof.rs#L144

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/list/UncheckedListProof.java
MakarovS: `ListProof` is an entity that contains `ListProofNode` and `length`. We could go with a different name for that entity, that's the best one I could come up with. I thought about `ListProofRoot` naming, but it doesn't reflect the fact that we have length in there.

dmitry-timofeev: ListProof name is perfect — the question is why we have a ListProof which seems to be a complete proof from the list — and also an UnchekedListProof interface that contains the ListProof.

exonum-java-binding/core/rust/src/storage/proof_list_index.rs
dmitry-timofeev: That is really strange — why force a user provide the size when we assume they don't know _anything_ about the list (including its size!), don't trust the system, hence need a proof. How do they know the size (without trusting the system)? If they do trust it, then the proof is not needed at all.

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/list/ListProofHashCalculator.java
```diff
@@ -116,18 +136,16 @@ private HashCode visitLeft(ListProofBranch branch, long parentIndex) {
   }
 
   /**
-   * Returns calculated root hash of a list proof tree.
-   *
-   * @return hash code
+   * Returns calculated hash of a list proof tree.
```
dmitry-timofeev: The documentation is still incomplete: is it a merkle root hash of the tree, or a hash of the ProofList collection?

exonum-java-binding/core/rust/src/storage/proof_list_index.rs
vitvakatu: Yep, looks like it violates our privacy guarantees. Created https://jira.bf.local/browse/ECR-3414

pavel-mukhanov: Maybe I missing something, but in rust, we force the user to provide list length only when we validating the proof. We don't need list size to construct the proof.

dmitry-timofeev: > we force the user to provide list length only when we validating the proof.

Why does the core exclude the essential part of the proof? The proof cannot be verified without this piece of data. Why is the user forced to obtain that missing piece of the proof manually?

exonum-java-binding/common/src/main/java/com/exonum/binding/common/proofs/list/UncheckedListProof.java
MakarovS: Created a task for that - https://jira.bf.local/browse/ECR-3429

dmitry-timofeev:
It doesn't build

MakarovS:
Fixed that.
Although now it gets 405 from coveralls -_- restarted the job, should work now.

dmitry-timofeev:
Good, thanks, coveralls had been having some issues, failing all reporting jobs.

PR #1061, opened 2019-08-07
Fix Mac OS build [ECR-3382]

vitvakatu:
## Overview

Fixes static linkage on Mac and reenables Travis CI.

---
See: https://jira.bf.local/browse/ECR-3382

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ env:
   global:
     - RUST_COMPILER_VERSION=stable
     - OSX_PACKAGES="libsodium rocksdb pkg-config protobuf"
-    - ROCKSDB_STATIC=1
+    - ROCKSDB_LIB_DIR=/usr/lib
     - SNAPPY_LIB_DIR=/usr/lib/x86_64-linux-gnu
     # REPO_TOKEN used for integration with coveralls is encoded here
     - secure: Fp22iaJpttsIArAyWmdCGNtljIALTYRVKO7O+H2hgBkwHHqrU7+15sbaq3xzhz4YNWNfuFMIkFUBgd/KYHgAuNDDrtm2agib13C0lQT1NFQO9ccmNCJNsXQrYrXGwpnNqPKp0YmfBfgNwzEpBerlbtvzV/T/RZukT/403XxwxU9y5tHfQokwVLibqP2jJsxdihTfCKIOs+o6hBfArmsn+e+panEv17ZrCjOmBIM/W70Rf2rEM26wFnYsfnAUTCkpl4Ong0SYNpZZxNMtw61W8ApDY8bpz7cKUxCv7SmD3kO7Y+TTHWfWYx6FNXtUpE1vCi6I7fZAY16rViTWOX55NCeFQz56XER7ArJQZtC/nC1lZ9tGKtcofu2Rq7WUoRuTwvLTaf6VzAP/CUj0DUxkV+8WUggl3s/Im7Y9rn8Aqvh8LReZmqzTY+dJ0hFG4DLoLtl71eTEnNoumi5UleBhJPaei3wPNPHg1WlOmhFyhRCsbIIGiyFtSj/faLmdc7tN/sBFANb0g4Exl0mRNvB0IfS1gM6XouEGUTlVree68p11PnsGJGs/QaUB9F9AAGVKTZ2kz7sqkCDdGmLxzbdidYDHZtYWfOIYSJCQsA09n2Txi0fwNByKfl/spdyMmtI1uGeT803rhN9vu0NGrQFG3mU7mqO33fUDEStIQ6/xn0A=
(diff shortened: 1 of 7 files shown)


exonum-java-binding/core/rust/src/lib.rs
```diff
@@ -38,12 +38,13 @@ pub extern crate serde_json;
...
 extern crate lazy_static;
 
-pub extern crate exonum_merkledb;
 extern crate exonum_testkit;
 extern crate exonum_time;
 #[cfg(test)]
 extern crate tempfile;
 
+pub use exonum::exonum_merkledb;
+
```
skletsun: Can't we reference exonum as a git dependency then?

exonum-java-binding/core/rust/Cargo.toml
```diff
@@ -29,6 +28,13 @@ serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0.40"
 
+[dependencies.exonum]
+# FIXME: using git dependency until new Exonum with MerkleDB will be released
```
skletsun: This is a good idea, but I'm afraid we cannot deal with that version error without the `[patch.crates-io]` section. I'd suggest moving underlying lines there.

exonum-java-binding/package_app.sh
```diff
@@ -44,11 +44,10 @@ function build-exonum-java-for-platform() {
...
 
 function build-exonum-java-macos() {
-    # We use static linkage for RocksDB on Mac because dynamic linking
-    # on Mac does not work: the resulting app has a dependency on a _particular_
+    # We use static linkage for RocksDB on Mac because in case of dynamic linking
+    # the resulting app has a dependency on a _particular_
     # version of the RocksDB library, hence, is incompatible with any updates
-    # to the library, even the patch ones. It is believed to be a Cargo issue
-    # or RocksDB build configuration issue, see ECR-3324.
+    # to the library, even the patch ones.
```
dmitry-timofeev: Please clarify this comment to reflect the actual situation.

exonum-java-binding/core/rust/build.rs
```diff
@@ -0,0 +1,11 @@
+use std::env::var_os;
+
+fn main() {
+    // We need to link to libc++.dylib on Mac if using static linkage with RocksDB
+    // This is because `librocksdb.a` provided with Homebrew package depends on
+    // `libc++.dylib` and in case of static linkage the resulting `libjava_bindigs.dylib`
+    // also needs to be linked against `libc++.dylib`.
+    if cfg!(target_os = "macos") && var_os("ROCKSDB_STATIC").is_some() {
+        println!("cargo:rustc-link-lib=dylib=c++");
```
dmitry-timofeev: What are the implications to the app runtime requirements?

.travis.yml
```diff
@@ -28,17 +28,13 @@ addons:
...
       - protobuf-compiler
       - libprotobuf-dev
-      - liblz4-dev
-      - libzstd-dev
 
 env:
   global:
     - RUST_COMPILER_VERSION=stable
     - OSX_PACKAGES="libsodium rocksdb pkg-config protobuf"
     - ROCKSDB_STATIC=1
```
dmitry-timofeev: You said on a call we link on Linux dynamically. But this env variable is shared between all jobs.

vitvakatu: Good catch, thanks. The configuration was invalid and this variable didn't work because no `ROCKSDB_LIB_DIR` was provided.

exonum-java-binding/package_app.sh
```diff
@@ -46,8 +46,8 @@ function build-exonum-java-for-platform() {
 function build-exonum-java-macos() {
     # We use static linkage for RocksDB on Mac because in case of dynamic linking
     # the resulting app has a dependency on a _particular_
```
dmitry-timofeev: @vitvakatu it is not true — haven't you clarified that soname/install_name do not contain the same version string as the library? Please fix misleading statements in the docs.

exonum-java-binding/core/rust/build.rs
dip56: macos command-line tools use the clang in dependencies. but I think without command-line tools you can't build anything :)

exonum-java-binding/package_app.sh
vitvakatu: It usually contains the minor version of the library (but not in case of libsodium). Updated, how about current comment?

dmitry-timofeev: It's good, thanks. I've updated it slightly.

vitvakatu:
Using system snappy on Linux leads to a compilation error for unknown reason. Removed it from travis dependencies, now Snappy will be compiled from sources

.travis.yml
```diff
@@ -23,22 +23,16 @@ addons:
       - build-essential
       - libsodium-dev
       - librocksdb5.17
-      - libsnappy-dev
```
dmitry-timofeev: Y? What is then Snappy_lib_dir?

exonum-java-binding/core/rust/build.rs
dmitry-timofeev: The question is not on building, but on the required runtime dependencies on the user box.

.travis.yml
vitvakatu: See above comment, system snappy is unusable on Linux now. Updated wiki page as well.

dmitry-timofeev: **What is then Snappy_lib_dir?**

vitvakatu: It is used only on Mac now, where the snappy is installed via homebrew. This is the list of Linux packages only

dmitry-timofeev: Good, sorry, I missed it remains in macos-only section only, thanks.

vitvakatu:
I will update installation docs to mention `libc++.dylib` dependency in a separate PR

PR #1068, opened 2019-08-14
Fixed index type integration tests [ECR-3415]

skletsun:
## Overview
Fixed integration tests that check the type of index.
---
See: https://jira.bf.local/browse/ECR-3415

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/src/test/java/com/exonum/binding/core/storage/indices/BaseIndexProxyTestable.java
+++ b/exonum-java-binding/core/src/test/java/com/exonum/binding/core/storage/indices/BaseIndexProxyTestable.java
@@ -19,6 +19,7 @@ package com.exonum.binding.core.storage.indices;
 import static com.exonum.binding.test.Bytes.bytes;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.containsStringIgnoringCase;
 import static org.hamcrest.Matchers.sameInstance;
 import static org.hamcrest.core.IsEqual.equalTo;
 import static org.junit.internal.matchers.ThrowableMessageMatcher.hasMessage;
@@ -36,7 +37,6 @@ import com.exonum.binding.test.RequiresNativeLibrary;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
@@ -179,7 +179,6 @@ abstract class BaseIndexProxyTestable<IndexT extends StorageIndex> {
    * - Constructor of the other type checks it, preventing illegal access to the internals.
    */
   @Test
-  @Disabled("Fails for EntryIndexProxy ECR-3415")
   void indexConstructorPersistsIndexTypeInfo() throws CloseFailuresException {
     try (Cleaner cleaner = new Cleaner()) {
       String name = "test_index";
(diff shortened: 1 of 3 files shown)


exonum-java-binding/core/src/test/java/com/exonum/binding/core/storage/indices/BaseIndexProxyTestable.java
```diff
@@ -197,7 +196,8 @@ void indexConstructorPersistsIndexTypeInfo() throws CloseFailuresException {
       Exception e = assertThrows(RuntimeException.class, () -> createOfOtherType(name, snapshot));
 
       // TODO: Change message after https://jira.bf.local/browse/ECR-3354
-      assertThat(e, hasMessage(containsString("Index type doesn't match specified")));
+      assertThat(e, hasMessage(containsStringIgnoringCase(
+              "Index type does not match specified one")));
```
skletsun: After ECR-3418 we have some changes in error messages. For the Entry case it is now _"Error while reading the index state. Possibly the index type does not match specified one"_, for the rest of indices it is _"Index type does not match specified one"_. @dmitry-timofeev is this solution acceptable?

vitvakatu: I think it is not the solution of the initial problem, as it does not show even the actual type of the index, does it?

skletsun: Well, at line 198 there is a //TODO with link to the task I created already for this problem - https://jira.bf.local/browse/ECR-3354

vitvakatu: Ah, sorry, I thought that this PR is actually aimed to fix ECR-3354 :D

skletsun: What we can see as an error output for tests on indices other than `Entry`:
```
"assertion failed: `(left == right)`
  left: `KeySet`,
 right: `List`: Index type does not match specified one"
```
So, in fact one can see the index type, but as part of a wrong assertion.

dmitry-timofeev: I think yes, the alternative is too complicated to bother (add an assumption here that index type is not an entry to exclude it + an entry-only test).

dmitry-timofeev: Do those left and right get into the message?

skletsun: Yes, it is the message we get from assertion in the `metadata::index_metadata()`:
```
        assert_eq!(
            metadata.index_type, index_type,
            "Index type does not match specified one"
        );
```

dmitry-timofeev: and into the message of a Java Exception?

skletsun: Yes. `Index::new()` obviously panics on native assert -> we catch it and throw as a `RuntimeException`.

dmitry-timofeev: So ECR-3354 is to make it clear which type is expected?

skletsun: I don't remember the details of the discussion that caused this ticket, but in general - yes.

PR #1071, opened 2019-08-15
Enable a test verifying independence of collection groups

dmitry-timofeev:
## Overview

It seems the collection groups became independent in MerkleDB,
and modifications of a collection in one group cannot affect
collections in other groups, even if first group id is a sub-prefix
of the other group id.

---

### Definition of Done


dmitry-timofeev:
@aleksuss are they really independent now? Is such (or similar) situation impossible in MerkleDB?

MakarovS:
CI fails though :(

aleksuss:
@dmitry-timofeev Yes. They are really independent. Each index instance has its own identifier which is a part of the key.

dmitry-timofeev:
> CI fails though :(

Weird, Exonum no longer compiles (this test passed, but I ran it through IDE, so no native lib compilation happened). 

However, there is a new release of Rust, so we must absolutely not be surprised that something broke (@vitvakatu , @skletsun ): https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1370-2019-08-15

PR #1079, opened 2019-08-20
Update to exonum 0.12 [ECR-3450]

skletsun:
## Overview

Pointing branch to latest exonum 0.12. BTC anchoring service is disabled because it breaks the dynamic binding on the app start.

---
See: https://jira.bf.local/browse/ECR-3450

### Definition of Done

### DIFF
--- a/exonum-java-binding/core/rust/Cargo.lock
+++ b/exonum-java-binding/core/rust/Cargo.lock
@@ -176,7 +176,7 @@ name = "backtrace"
 version = "0.3.34"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "backtrace-sys 0.1.31 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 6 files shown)


exonum-java-binding/core/rust/Cargo.lock
```diff
@@ -270,7 +270,7 @@ name = "bzip2-sys"
 version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "cc 1.0.38 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cc 1.0.26 (registry+https://github.com/rust-lang/crates.io-index)",
```
dmitry-timofeev: Are cc and backtrace downgraded on purpose?

skletsun: No, it's probably an accidental result of merging. Thanks, going to correct this.

PR #1080, opened 2019-08-20
Add service configuration parameters [ECR-3464] 

dmitry-timofeev:
## Overview

Currently they can only be accessed as a protobuf message.
If needed, built-in support for some pre-defined configuration
formats may be added (e.g., a protobuf message to pass Java Properties
as service configuration + Configuration#getAsProperties -> Properties).

---
See:
* https://jira.bf.local/browse/ECR-3437
* https://jira.bf.local/browse/ECR-3464 
* https://jira.bf.local/browse/ECR-3469

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Service.java
```diff
@@ -35,13 +34,23 @@
 
   /**
    * Configures the service instance. This method is called <em>once</em> after the service
```
dmitry-timofeev: @alekseysidorov That's our take on the configuration documentation (+ Configuration.java)

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Configuration.java
```diff
@@ -0,0 +1,48 @@
...
+ * the configuration parameters as a service-specific protobuf message in the transactions
+ * which add that service instance to the network. After Exonum starts the service, it
+ * {@linkplain Service#configure(Fork, Configuration) passes the configuration parameters}
+ * to the newly created service instance.
+ *
+ * <p>Service reconfiguration is not currently supported, but will be added soon.
+ *
+ * @see Service#configure(Fork, Configuration)
+ */
+public interface Configuration {
```
MakarovS: Shouldn't this be `ServiceConfiguration`? And implementation `ServiceConfigurationImpl`?
It's good as it is as well though, being in `service` package.

dmitry-timofeev: I went with a shorter name (`Configuration`) inside `service` package, so that `Service` is not repeated where configuration is used in the Service#configure signature. 

`ServiceConfiguration` name can probably be improved, but I don't currently see anything particularly specific about that implementation.

PR #1085, opened 2019-08-23
Add new message format support [ECR-3473]

dmitry-timofeev:
## Overview

Support the new, protobuf-based, message format.

*There are a couple of questions for discussion marked as todos.*

The definitions are as of exonum/exonum@84ead6191173138f96feabadd3033b3f9aa42368
and will be subsequently patched to work well with Java codegen. The corresponding commit
may be skipped (b0448c7 )

---
See: https://jira.bf.local/browse/ECR-3473

### Definition of Done

- [ ] Public API has Javadoc


exonum-java-binding/common/src/main/java/com/exonum/binding/common/message/ParsedTransactionMessage.java
```diff
@@ -0,0 +1,136 @@
...
+import com.google.protobuf.InvalidProtocolBufferException;
+import java.util.Objects;
+
+/**
+ * A parsed transaction message. On instantiation, it decodes the signed message payload
+ * as an Exonum transaction.
+ */
+final class ParsedTransactionMessage implements TransactionMessage {
+
+  private final Consensus.SignedMessage signedMessage;
```
bullet-tooth: It's little confusing that we need references to `Consensus.SignedMessage` and `SignedMessage`.
As I can see `Consensus.SignedMessage`  is required here for convert to bytes array?

exonum-java-binding/common/src/main/java/com/exonum/binding/common/message/SignedMessage.java
```diff
@@ -0,0 +1,108 @@
...
+   * Parses the signed message bytes. The parsing does not involve the signature verification —
+   * do it separately if needed.
+   *
+   * @param messageBytes the serialized message to parse
+   * @return a signed message with exonum message as its payload
+   * @throws InvalidProtocolBufferException if the given bytes are not a serialized
+   *     {@link Consensus.SignedMessage}; or if the payload of the message is not
+   *     {@link Consensus.ExonumMessage}
+   */
+  public static SignedMessage parseFrom(byte[] messageBytes) throws InvalidProtocolBufferException {
```
bullet-tooth: this class has package private level but methods are public

exonum-java-binding/common/src/main/java/com/exonum/binding/common/message/SignedMessage.java
```diff
@@ -0,0 +1,108 @@
...
+package com.exonum.binding.common.message;
+
+import com.exonum.binding.common.crypto.PublicKey;
+import com.exonum.binding.messages.Consensus;
+import com.exonum.binding.messages.Consensus.ExonumMessage;
+import com.google.protobuf.ByteString;
+import com.google.protobuf.InvalidProtocolBufferException;
+
+/**
+ * A wrapper around {@link Consensus.SignedMessage} protobuf message containing
```
bullet-tooth: It's not clear what is the purpose of this class and where it can be helpful. As I can see in `PassedMessage` we can get everything we need from the protobuf message.

dmitry-timofeev: 🤷‍♂ The line below says " which converts protobuf types into internal types." That is its main purpose.

You do need such a class, unless you are fine with such client code:
```
PublicKey authorPk = PublicKey.fromBytes(message.getAuthor()
        .toByteArray());
```

Can you spot a mistake in this snippet — it compiles, will not throw in runtime, but is incorrect?

dmitry-timofeev: On top of that — it stores ExonumMessage as ByteString, which is even less intuitive to use.

exonum-java-binding/common/src/main/proto/consensus.proto
```diff
@@ -0,0 +1,142 @@
...
+message SignedMessage {
+  // Message payload.
+  bytes payload = 1;
+  // Message author.
+  exonum.PublicKey author = 2;
+  // Digital signature.
+  exonum.Signature signature = 3;
+}
+
+// List of consensus messages
```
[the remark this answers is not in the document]
vitvakatu: See https://jira.bf.local/browse/ECR-3472

exonum-java-binding/common/src/main/java/com/exonum/binding/common/message/ParsedTransactionMessage.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  @Override
+  public int getTransactionId() {
+    return tx.getCallInfo().getMethodId();
+  }
+
+  @Override
+  public byte[] getPayload() {
+    // todo: Return as ByteString? Add getPayloadAs(MessageType)?
```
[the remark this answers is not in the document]
dmitry-timofeev: I think that if we encourage the use of protobuf then the payload is likely a serialized protobuf message. A message type can be created both from a byte array and a ByteString. The latter will not require two extra copies. The downside is that we use a protobuf type in a TransactionMessage interface — but as all the messages are in protobuf now, we can probably live with that.

exonum-java-binding/common/src/main/proto/consensus.proto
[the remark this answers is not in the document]
dmitry-timofeev: I added them (and a patch atop of them) as separate commits to be able to re-apply easily (hopefully, to a separate repo and not to the re-imported definitions 🙃 ).

exonum-java-binding/common/src/main/java/com/exonum/binding/common/message/ByteStringFunnel.java
```diff
@@ -0,0 +1,43 @@
...
+package com.exonum.binding.common.message;
+
+import com.exonum.binding.common.hash.Funnel;
+import com.exonum.binding.common.hash.PrimitiveSink;
+import com.google.protobuf.ByteString;
+
+/**
+ * A funnel for ByteStrings which puts their bytes to the sink without copying.
+ */
+enum ByteStringFunnel implements Funnel<ByteString> {
```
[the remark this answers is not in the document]
dmitry-timofeev: I've found only a single such usage, and decided to inline the `#getAsReadonlyByteBuffer` instead of making the funnel public.

PR #1106, opened 2019-09-06
Update Changelog for the 0.8.0 release

vitvakatu:
## Overview

---

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/CHANGELOG.md
+++ b/exonum-java-binding/CHANGELOG.md
@@ -15,6 +15,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [0.8.0] - 2019-09-09
+
+### Overview
+
+This release brings mainly internal fixes and improvements. It is based on Exonum 0.12.
+
 ### Changed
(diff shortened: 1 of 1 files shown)


exonum-java-binding/CHANGELOG.md
```diff
@@ -15,6 +15,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [0.8.0] - 2019-09-09
+
+### Overview
+
+This release brings mainly internal fixes and improvements. It is based on Exonum 0.12.
```
MakarovS: Sorry for late review, this was merged before I had a chance to look at it - but shouldn't we mention MerkleDb support here? Or does it fall under "internal fixes" from user's point of view and mention of Exonum 0.12 is sufficient?

vitvakatu: I thought about it, but it turns out that these changes are internal, there is no difference from the user's point of view (yet). We didn't move to MerkleDB, we just moved to Exonum 0.12 which (accidentally) has MerkleDB as one of the components.

So I think we should not mention that.

PR #1120, opened 2019-09-16
Reimplement View to support mutable Fork references [ECR-3525]

vitvakatu:
## Overview

New `View` implementation which works with mutable references and is more safe than the current.

## How to review

- All `*_index.rs` files contain a single change: no more need in dereferencing result of `View::get`
- `storage/db.rs` module was completely changed, review the file as whole, do not try to understand git diff.
- Take special interest in documentation (it is our internal docs, so we must understand it, but noone else must) and tests; there are a lot of corner-cases when dealing with such an unsafe API. It might not be possible (or useful) to check all of them, but if you have some ideas on additional test cases - you're welcome.
---
See: https://jira.bf.local/browse/ECR-3525

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/rust/src/storage/db.rs
+++ b/exonum-java-binding/core/rust/src/storage/db.rs
@@ -15,7 +15,6 @@
 use exonum_merkledb::{Fork, Snapshot};
 use jni::{objects::JClass, JNIEnv};
 
-use cast_handle;
 use handle::{self, Handle, NonOwnedHandle};
 
 pub(crate) type Key = Vec<u8>;
@@ -57,13 +56,14 @@ pub(crate) enum ViewOwned {
 }
 
 /// Hides the differences between owning and non-owning `View` variants
-/// and simplifies the use indexes API.
+/// and simplifies the use of the indexes API.
 #[derive(Clone)]
 pub(crate) enum ViewRef<'a> {
     Snapshot(&'a dyn Snapshot),
     Fork(&'a Fork),
 }
 
+#[allow(clippy::useless_transmute)]
 impl<'a> ViewRef<'a> {
     unsafe fn from_fork(fork: &'a Fork) -> Self {
         // Make a provided reference `'static`.
@@ -111,6 +111,8 @@ impl View {
     /// SIGINT will occur.
     ///
     /// Both indexes mutability and `&mut self` methods of `Fork` available.
+    // TODO: remove dead_code after ECR-3519
+    #[allow(dead_code)]
     pub fn from_ref_mut_fork(fork: &mut Fork) -> Self {
         View::RefFork(NonOwnedHandle::new_mut(fork))
     }
@@ -119,12 +121,8 @@ impl View {
     /// in indexes operations.
     pub fn get(&self) -> ViewRef<'_> {
         match self {
-            View::RefFork(handle) => {
-                ViewRef::Fork(handle.get())
-            }
-            View::RefSnapshot(snapshot_ref) => {
-                ViewRef::Snapshot(*snapshot_ref)
-            },
+            View::RefFork(handle) => ViewRef::Fork(handle.get()),
+            View::RefSnapshot(snapshot_ref) => ViewRef::Snapshot(*snapshot_ref),
             View::Owned(owned) => match owned {
                 ViewOwned::Fork(fork) => unsafe { ViewRef::from_fork(&fork) },
                 ViewOwned::Snapshot(snapshot) => unsafe { ViewRef::from_snapshot(&**snapshot) },
@@ -164,7 +162,7 @@ pub extern "system" fn Java_com_exonum_binding_core_storage_database_Views_nativ
 mod tests {
     use super::*;
     use exonum_merkledb::{Database, Entry, IndexAccess, TemporaryDB};
-    use to_handle;
+    use handle::{cast_handle, to_handle};
 
     const TEST_VALUE: i32 = 42;
 
--- a/exonum-java-binding/core/rust/src/storage/db.rs
+++ b/exonum-java-binding/core/rust/src/storage/db.rs
@@ -120,8 +120,7 @@ impl View {
     pub fn get(&self) -> ViewRef<'_> {
         match self {
             View::RefFork(handle) => {
-                let fork = cast_handle::<Fork>(handle.as_handle());
-                ViewRef::Fork(fork)
+                ViewRef::Fork(handle.get())
             }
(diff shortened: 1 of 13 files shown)


exonum-java-binding/core/rust/src/storage/db.rs
```diff
@@ -28,104 +28,120 @@ pub(crate) type Value = Vec<u8>;
...
-/// way in Rust to make a `View` value not movable. Furthermore, it have to be moved from the stack
-/// to the heap in order to be converted into `Handle` for the java side. So a `Fork` value
-/// should be placed in the heap to prevent its movement after creating a reference to it.
-
-pub(crate) struct View {
-    /// Unsafe reference to the owned View (owned either by Rust or by Java)
-    reference: ViewRef,
-    /// Allows to drop owned view if needed
-    owned: Option<ViewOwned>,
+/// As `View` does not have a lifetime, nothing protects us from dereferencing a freed reference,
```
dmitry-timofeev: I don't get what you mean here by 'dereferencing a freed reference' and how it is connected with View instantiation.

exonum-java-binding/core/rust/src/storage/db.rs
```diff
@@ -28,104 +28,120 @@ pub(crate) type Value = Vec<u8>;
...
+/// the extreme caution must be taken in places where `View` is constructed and send to Java.
```
dmitry-timofeev: sent to Java

exonum-java-binding/core/rust/src/storage/db.rs
```diff
@@ -28,104 +28,120 @@ pub(crate) type Value = Vec<u8>;
...
+/// Java code must never store a `View` handle beyond the scope it was initially acquired
```
dmitry-timofeev: Not sure I understand what you mean here. Usually, Java releases a handle, and it must not be used subsequently. Is it anyhow different for this thing?

exonum-java-binding/core/rust/src/storage/db.rs
```diff
@@ -28,104 +28,120 @@ pub(crate) type Value = Vec<u8>;
...
+    /// Created `View` must never outlive the reference it was created with,
+    /// or `SIGINT` will occur.
+    RefMutFork(&'static mut Fork),
+    /// Immutable Snapshot view, constructed from `&Snapshot`. There is no need in mutable
+    /// variant.
+    ///
+    /// Created `View` must never outlive the reference it was constructed with,
+    /// or `SIGINT` will occur.
+    RefSnapshot(&'static dyn Snapshot),
+    /// Covers both `Snapshot` and `Fork` cases. Rust uses move semantic and single-ownership
```
dmitry-timofeev: I'd clarify, what Owned means. If I understand correctly, the
client code owns a Snapshot/View object and then transfers the ownership of it to the View, which then owns it. I.e. 'owned' means different things in different contexts (= when used in factory method below and here in enum variant).

exonum-java-binding/core/rust/src/storage/db.rs
```diff
@@ -141,95 +157,118 @@ mod tests {
...
+    fn get_mut_fork() {
+        let db = TemporaryDB::new();
+        let mut fork = db.fork();
+        let view = View::from_ref_mut_fork(&mut fork);
+        let mock_method = |_: &mut Fork| {};
+        match view {
+            View::RefMutFork(fork_ref) => {
+                mock_method(fork_ref);
+            }
+            _ => panic!(),
```
dmitry-timofeev: What is tested here and what the panics mean (also in other methods)?

exonum-java-binding/core/rust/src/storage/db.rs
vitvakatu: The View does not have a lifetime, so we use `unsafe` to prolong the lifetime of the reference it was constructed with to `'static`. If this original reference is destroyed (leaves the scope), our prolonged reference stored inside View is no longer valid and will lead to SIGINT if we use it. So we must carefully review all the places where View is constructed from references to make sure View never outlives the original reference. Does it make sense now?

I agree, `freed` cannot be used in case of references here, I will rephrase.

exonum-java-binding/core/rust/src/storage/db.rs
vitvakatu: Yep, no difference here. Java can't store handle to `View::Ref*` variants, because noone can verify that the reference will be still valid. Do you think I should remove this sentence as `NativeHandle` docs covers this enough? I just wasn't able to find any docs in native code explaining this mechanics.

exonum-java-binding/core/rust/src/storage/db.rs
vitvakatu: You're right, but I can't see why the meaning in different here and there. 🤔 We have "owned" Snaphot/Fork -> we use `from_owned` method -> we have "owned" View, i.e. "View that owns Snapshot/Fork", which is the opposite of "View that borrows Snapshot/Fork".

exonum-java-binding/core/rust/src/storage/db.rs
vitvakatu: It's just a pattern matching that has only one valid variant (and other variants must lead to test failure). I'd add panic message explaining the purpose.

exonum-java-binding/core/rust/src/storage/db.rs
dmitry-timofeev: Yes, thank you, that's clearer

exonum-java-binding/core/rust/src/storage/db.rs
dmitry-timofeev: > Java can't store handle to View::Ref* variants, because noone can verify that the reference will be still valid.

dmitry-timofeev: What do you mean by Java not being able to store a handle to ref variants? How could it use them then? Or do you mean for longer than the method invocation?

vitvakatu: Yep, for longer than the method invocation (that was the purpose of "beyond the scope it was initially acquired") 🤔 Probably I should avoid such phrases, my fault.

exonum-java-binding/core/rust/src/storage/db.rs
dmitry-timofeev: I mean that the subject which owns the snapshot/fork is different, but that it probably isn't that important.

PR #1129, opened 2019-09-26
Implement JavaServiceRuntime in Rust [ECR-3480]

ipenkin:
## Overview
Implemented trait ```exonum::runtime::Runtime``` as proxy for java ```ServiceRantimeAdapter```. 
```NodeContext``` was modified due to new messages format, some changes may be required after #1085
Dependency to native exonum has been updated, https://github.com/exonum/exonum/commit/b542dae is used currently.
Old proxies for ```Service``` and ```Transaction``` need to be removed.

_There are a couple of questions for discussion marked as todos._

---
See: https://jira.bf.local/browse/ECR-3480
---

### TODO list:
- [ ] ~~test-set for java-independent code of ```Runtime``` impl~~ (after logic simplification)
- [ ] ~~consider getting ```ArtifactProtobufSpec``` from ```ServiceRuntimeAdapter```~~ (not required so far)
- [ ] consider ```execute``` method modification for cases where caller is ```Blockchain```

### Definition of Done

- [ ] Public API has Javadoc

### CODE REVIEW

**Comment by vitvakatu** [comment]
Please restore deleted `Cargo.lock`, as it is probably needed for correct work of `exonum-java` application

File: exonum-java-binding/core/rust/src/proxy/runtime.rs [hunk 1]
```diff
@@ -0,0 +1,596 @@
...
+                JavaType::from_str("V").unwrap(),
+                &[
+                    JValue::from(artifact_id),
+                    JValue::from(spec),
+                ],
+            )?;
+            Ok(())
+        }));
+
+        self.deployed_artifacts.insert(id);
```
vitvakatu: Why we insert id into a list of deployed artifacts even if an error returned from java?

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,596 @@
...
+                env.new_string(id.to_string())?
+            );
+            let spec = JObject::from(env.byte_array_from_slice(
+                &deploy_spec.into_bytes())?
+            );
+
+            env.call_method_unchecked(
+                self.runtime_adapter.as_obj(),
+                runtime_adapter::deploy_artifact_id(),
+                JavaType::from_str("V").unwrap(),
```
vitvakatu: Parsing type from string is unnecessary slow

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,596 @@
...
+                        JValue::from(validator_id),
+                        JValue::from(height),
+                    ],
+                ),
+            );
+            Ok(())
+        }));
+    }
+
+    fn api_endpoints(&self, context: &ApiContext) -> Vec<(String, ServiceApiBuilder)> {
```
vitvakatu: This method won't work, as there is no way to attach Java endpoints to the actix-web server of Core. This method is removed in latest Core revision.

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,596 @@
...
+pub struct JavaArtifactId {
+    /// `groupId` maven property
+    pub group: String,
+    /// `artifactId` maven property
+    pub artifact: String,
+    /// `version` maven property in format {major.}{minor}.{patch}
+    pub version: Version,
+}
+
+struct AfterCommitContext<'a> {
```
vitvakatu: Why the AfterCommitContext is needed?

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,596 @@
...
+    exec: Executor,
+    runtime_adapter: GlobalRef,
+    deployed_artifacts: HashSet<JavaArtifactId>,
+    started_services: HashMap<InstanceId, Instance>,
+    started_services_by_name: HashMap<String, InstanceId>,
+}
+
+/// Service identification properties within `JavaRuntimeProxy`
+#[derive(Debug, Clone)]
+struct Instance {
```
vitvakatu: I believe something similar (InstanceId?) exists in Core, doesn't it?

exonum-java-binding/core/rust/src/proxy/runtime.rs
vitvakatu: Sorry, my fault, the method is not removed, but can be used only by Rust runtime

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,596 @@
...
+        if artifact.runtime_id != Self::RUNTIME_ID as u32 {
+            return Err(Error::IncorrectArtifactId.into());
+        }
+        artifact
+            .name
+            .parse()
+            .map_err(|_| Error::IncorrectArtifactId.into())
+    }
+
+    fn add_started_service(&mut self, instance: Instance) {
```
vitvakatu: Runtime must handle attempt to start already started service

exonum-java-binding/core/rust/src/proxy/runtime.rs
ipenkin: There is similar structure in core - ```InstanceSpec```, which is extended by ```ArtifactId```. 
```InstanceSpec``` or simple ```InstanceId```  (integer identifier) can be used instead of custom struct ```Instance ```, the second variant is available if mapping services by name is not required for future.

File: exonum-java-binding/core/rust/src/proxy/runtime.rs (hunk 2 above)
Comment (ipenkin) [reply exonum-java-binding/core/rust/src/proxy/runtime.rs]: @vitvakatu should it be cached like ```method id```

Comment (vitvakatu) [reply exonum-java-binding/core/rust/src/proxy/runtime.rs]: JavaType is a simple enum, so here you can use `JavaType::Primitive(Primitive::Void)`

File: exonum-java-binding/core/rust/Cargo.toml [hunk 7]
```diff
@@ -15,23 +15,31 @@ resource-manager = []
 invocation = ["jni/invocation"]
 
 [dependencies]
-exonum = { version = "0.11", features = ["rocksdb_snappy", "rocksdb_lz4", "rocksdb_zlib", "rocksdb_bzip2"] }
-exonum-time = "0.11"
-exonum-testkit = "0.11"
+exonum = { version = "0.12", features = ["rocksdb_snappy", "rocksdb_lz4", "rocksdb_zlib", "rocksdb_bzip2"] }
+exonum-derive = "0.12"
+exonum-merkledb = "0.12"
```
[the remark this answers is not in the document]
ipenkin:
```
error[E0433]: failed to resolve: use of undeclared type or module `exonum_merkledb`
#[derive(Serialize, Deserialize, Clone, ProtobufConvert, PartialEq)]
                                        ^^^^^^^^^^^^^^^ use of undeclared type or module `exonum_merkledb`
```

Couldn't derive from `ProtobufConvert` without it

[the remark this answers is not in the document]
vitvakatu: That's unfortunate (due to implementation of ProtobufConvert), but can be solved by manually importing `use exonum::exonum_merkledb;` in file with `derive(ProtobufConvert)`.

exonum-java-binding/core/rust/src/proxy/runtime.rs
dmitry-timofeev: Not the runtime proxy.

vitvakatu: Okay, but then the list of started services must be stored in Java, mustn't it?

exonum-java-binding/core/rust/build.rs
```diff
@@ -8,4 +11,12 @@ fn main() {
     if cfg!(target_os = "macos") && var_os("ROCKSDB_STATIC").is_some() {
         println!("cargo:rustc-link-lib=dylib=c++");
     }
+
+    //let exonum_protos = get_exonum_protobuf_files_path();
```
dmitry-timofeev: Why is the code commented?

exonum-java-binding/core/rust/src/runtime/java_service_runtime.rs
```diff
@@ -80,26 +80,26 @@ impl JavaServiceRuntime {
...
-            )
-            .l()?;
-            let service = env.new_global_ref(service)?;
-            Ok(ServiceProxy::from_global_ref(
-                self.executor.clone(),
-                service,
-            ))
-        }))
-    }
+//    pub fn create_service(&self, artifact_id: &str) -> ServiceProxy {
```
dmitry-timofeev: Why is the code commented?

exonum-java-binding/core/rust/src/runtime/java_service_runtime.rs
```diff
@@ -80,26 +80,26 @@ impl JavaServiceRuntime {
...
+//            .l()?;
+//            let service = env.new_global_ref(service)?;
+//            Ok(ServiceProxy::from_global_ref(
+//                self.executor.clone(),
+//                service,
+//            ))
+//        }))
+//    }
 
     /// Loads an artifact from the specified location involving verification of the artifact.
```
dmitry-timofeev: Hasn't this method moved to the runtime and no longer needed? Also, the naming of this class needs to be updated because it conflicts with the runtime.

exonum-java-binding/core/rust/src/utils/jni_cache.rs
```diff
@@ -37,6 +37,15 @@ static mut OBJECT_GET_CLASS: Option<JMethodID> = None;
...
+static mut RUNTIME_ADAPTER_DEPLOY_ARTIFACT: Option<JMethodID> = None;
+static mut RUNTIME_ADAPTER_CREATE_SERVICE: Option<JMethodID> = None;
+static mut RUNTIME_ADAPTER_CONFIGURE_SERVICE: Option<JMethodID> = None;
+static mut RUNTIME_ADAPTER_STOP_SERVICE: Option<JMethodID> = None;
+static mut RUNTIME_ADAPTER_EXECUTE_TX: Option<JMethodID> = None;
+static mut RUNTIME_ADAPTER_STATE_HASHES: Option<JMethodID> = None;
+static mut RUNTIME_ADAPTER_AFTER_COMMIT: Option<JMethodID> = None;
+static mut RUNTIME_ADAPTER_MOUNT_API: Option<JMethodID> = None;
+
 // todo: Remove transaction and service adapter items when native JavaServiceRuntime is implemented
```
dmitry-timofeev: ❗️

exonum-java-binding/core/rust/src/utils/jni_cache.rs
```diff
@@ -77,6 +86,54 @@ unsafe fn cache_methods(env: &JNIEnv) {
         "getMessage",
         "()Ljava/lang/String;",
     );
+    RUNTIME_ADAPTER_DEPLOY_ARTIFACT = get_method_id(
+        &env,
+        "com/exonum/binding/core/runtime/ServiceRuntimeAdapter",
```
dmitry-timofeev: Extract the class name in a constant as it is repeated several times?

exonum-java-binding/core/rust/src/testkit/mod.rs
```diff
@@ -16,200 +16,200 @@
...
-    let public_key_byte_array: JObject = env.byte_array_from_slice(&keypair.0[..])?.into();
-    let secret_key_byte_array: JObject = env.byte_array_from_slice(&keypair.1[..])?.into();
-    env.call_static_method(
-        KEYPAIR_CLASS,
-        "createKeyPair",
-        KEYPAIR_CTOR_SIGNATURE,
-        &[secret_key_byte_array.into(), public_key_byte_array.into()],
-    )
-}
+//use self::time_provider::JavaTimeProvider;
```
dmitry-timofeev: Why is the code commented out? Is it *all* broken? It used to compile.

exonum-java-binding/core/rust/src/proxy/node.rs
```diff
@@ -69,34 +63,34 @@ impl NodeContext {
...
+        let _service_id = transaction.call_info.instance_id;
+//        if !self.blockchain.service_map().contains_key(&service_id) {
+//            return Err(format_err!(
+//                "Unable to broadcast transaction: service(ID={}) not found",
+//                service_id
+//            ));
+//        }
+
+        let key_pair = self.api_context.service_keypair();
+        let verified = Verified::from_value(
```
dmitry-timofeev: `signed`/`signedTx`/`signedMessage` (as we do not verify the transaction message — we create it (= sign)?

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,624 @@
...
+use Handle;
+use JniErrorKind;
+use JniResult;
+
+/// A proxy for `ServiceRuntimeAdapter`s.
+#[derive(Clone)]
+pub struct JavaRuntimeProxy {
+    exec: Executor,
+    runtime_adapter: GlobalRef,
+    deployed_artifacts: HashSet<JavaArtifactId>,
```
dmitry-timofeev: I don't see why the proxy needs to keep track of artifacts and services.

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,624 @@
...
+/// Service identification properties within `JavaRuntimeProxy`
+#[derive(Debug, Clone)]
+struct Instance {
+    id: InstanceId,
+    name: String,
+}
+
+/// Artifact identification properties within `JavaRuntimeProxy`
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct JavaArtifactId {
```
dmitry-timofeev: I don't see any code using these separate components. Please make it a string/a string typedef/a wrapper around it.

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,624 @@
...
+        let id = match self.parse_artifact(&artifact) {
+            Ok(id) => id,
+            Err(err) => return Box::new(Err(err).into_future()),
+        };
+
+        if let Err(err) = self.can_deploy_artifact(&id) {
+            return Box::new(Err(err).into_future());
+        }
+
+        Box::new(Self::parse_jni(self.exec.with_attached(|env| {
```
dmitry-timofeev: (here and each other method) I don't see the exception handling. If it does occur, the native is responsible to act on it (handle, according to the Runtime method specification, and clear, so that it does not remain 'thrown').

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,624 @@
...
+            self.add_deployed_artifact(id);
+            result
+        })
+        .into_future())
+    }
+
+    fn artifact_protobuf_spec(&self, id: &ArtifactId) -> Option<ArtifactProtobufSpec> {
+        let id = self.parse_artifact(id).ok()?;
+
+        if self.deployed_artifacts.contains(&id) {
```
dmitry-timofeev: Why this check is needed? I'd assume that the framework won't call the runtimes
that don't have a given artifact. The runtime implementation (in Java) might get this check if it gets implemented, but it does not seem to be needed in the proxy.

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,624 @@
...
+        } else {
+            None
+        }
+    }
+
+    fn start_service(&mut self, spec: &InstanceSpec) -> Result<(), ExecutionError> {
+        let adapter = self.runtime_adapter.as_obj().clone();
+        let service_name = spec.name.clone();
+        let id = spec.id;
+        let artifact = self.parse_artifact(&spec.artifact)?;
```
dmitry-timofeev: Here we parse the artifactId, only to convert it to string later. It seems unnecessary, as the correctness of such ids is checked before deploy in Java (and also each time it is instantiated in Java).

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,624 @@
...
+                runtime_adapter::create_service_id(),
+                JavaType::Primitive(Primitive::Void),
+                &[
+                    JValue::from(name),
+                    JValue::from(id as i32),
+                    JValue::from(artifact_id),
+                ],
+            )
+            .map(|_| ())
+        }))?;
```
dmitry-timofeev: Same here: exceptions must be properly handled.

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,624 @@
...
+    }
+
+    fn execute(
+        &self,
+        _dispatcher: &Dispatcher,
+        context: &mut ExecutionContext,
+        call_info: CallInfo,
+        arguments: &[u8],
+    ) -> Result<(), ExecutionError> {
+        let tx = if let (Some(key), Some(hash)) =
```
dmitry-timofeev: Is there an easier way to check that? If not, I'd request one from core, because this way is not intuitive.

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,624 @@
...
+            (context.caller.author(), context.caller.transaction_hash())
+        {
+            (key, hash)
+        } else {
+            // TODO: caller is Blockchain (not Transaction) is not supported  yet
+            return Err(Error::NotSupportedOperation.into());
+        };
+
+        Self::parse_jni(self.exec.with_attached(|env| {
```
dmitry-timofeev: As usual, the exception must be handled. On top of that, TransactionExecutionException must be converted into appropriate Errors, see the present TransactionProxy.

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,624 @@
...
+                        JValue::from(height),
+                    ],
+                ),
+            );
+            Ok(())
+        }));
+    }
+
+    fn api_endpoints(&self, context: &ApiContext) -> Vec<(String, ServiceApiBuilder)> {
+        let started_ids: Vec<i32> = self
```
dmitry-timofeev: Please see the docs of the runtime method:
(1) It must be invoked for the services to connect to API (not the *already* connected)
(2) It must not be empty (shall not be invoked if no new services need to be connected). If that complicates things, we might make the runtime ignore this case.
(3) As @vitvakatu mentioned, there is a method in the runtime that supplies just the right ids (at least must).

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,624 @@
...
+
+    /// Returns the current database snapshot. This snapshot is used to
+    /// retrieve schema information from the database.
+    pub fn snapshot(&self) -> &dyn Snapshot {
+        self.snapshot
+    }
+
+    /// If the current node is a validator, return its identifier, for other nodes return `None`.
+    pub fn validator_id(&self) -> Option<ValidatorId> {
+        CoreSchema::new(self.snapshot)
```
dmitry-timofeev: It is too complex. Why has it been removed? Shall we remove it as well?

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,624 @@
...
+        &self.service_keypair.1
+    }
+
+    /// Returns the current blockchain height. This height is "height of the last committed block".
+    pub fn height(&self) -> Height {
+        CoreSchema::new(self.snapshot).height()
+    }
+
+    /// Returns reference to communication channel with dispatcher.
+    pub(crate) fn dispatcher_channel(&self) -> &DispatcherSender {
```
dmitry-timofeev: This and some other methods of the context seem unused. Are they needed?

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,624 @@
...
+        to_handle(View::from_ref_snapshot(self.snapshot))
+    }
+}
+
+impl From<&ServiceStateHashes> for (InstanceId, Vec<Hash>) {
+    fn from(value: &ServiceStateHashes) -> Self {
+        let hashes: Vec<Hash> = value
+            .state_hashes
+            .iter()
+            .map(|bytes| Hash::from_bytes(bytes.into()).unwrap())
```
dmitry-timofeev: I'd consider extracting: `to_hashes(hashes: Vec<Vec<u8>>) -> Vec<Hash>`

exonum-java-binding/core/rust/src/proxy/service.rs
```diff
@@ -13,210 +13,211 @@
...
-                    "mountPublicApiHandler",
-                    "(J)V",
-                    &[JValue::from(node_handle)],
-                ),
-            );
-            Ok(())
-        }));
-    }
-}
+//
```
dmitry-timofeev: Why is this commented out and not removed?

exonum-java-binding/core/rust/src/runtime/service_factory_adapter.rs
```diff
@@ -73,9 +73,9 @@ impl JavaServiceFactoryAdapter {
             JAVA_SERVICE_RUNTIME = Some(runtime);
         }
     }
-}
+/*}
 
-impl ServiceFactory for JavaServiceFactoryAdapter {
+impl ServiceFactory for JavaServiceFactoryAdapter { */
     fn service_name(&self) -> &str {
```
dmitry-timofeev: Why is the code commented? Why do these methods remain? Are they used/needed?

exonum-java-binding/core/rust/src/testkit/time_provider.rs
```diff
@@ -14,46 +14,46 @@
...
 
-use chrono::{DateTime, Utc};
-use exonum_merkledb::BinaryValue;
-use exonum_time::time_provider::TimeProvider;
-use jni::{
-    objects::{GlobalRef, JObject},
-    Executor, JNIEnv,
-};
-use utils::unwrap_jni;
+//use chrono::{DateTime, Utc};
```
dmitry-timofeev: Same as with testkit — is it broken to be commented out? It used to compile, didn't it?

exonum-java-binding/core/rust/src/proxy/runtime.rs
dmitry-timofeev: How would you implement the service runtime without storing the services in Java?

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,596 @@
...
+
+impl Instance {
+    fn new(id: InstanceId, name: String) -> Self {
+        Self { id, name }
+    }
+}
+
+impl JavaArtifactId {
+    /// Creates new artifact description
+    pub fn new(group: &str, artifact: &str, major: u64, minor: u64, patch: u64) -> Self {
```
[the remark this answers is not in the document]
dmitry-timofeev: I don't think we shall parse the artifact id (and the version) here at all, see my other comment. The proxy does not need to handle any of these fields separately, just pass it as a string to the runtime.

exonum-java-binding/core/rust/src/proxy/node.rs
```diff
@@ -32,6 +31,7 @@ use handle::{cast_handle, drop_handle, to_handle, Handle};
 use storage::View;
 use utils::{unwrap_exc_or, unwrap_exc_or_default, unwrap_jni_verbose};
 use JniResult;
+use exonum::runtime::CallInfo;
```
skletsun: NIT: I'd nest this import line into the exonum import group above.

exonum-java-binding/core/rust/Cargo.toml
```diff
@@ -40,7 +46,9 @@ rpath = true
 
 # FIXME: using git dependency until new Exonum with MerkleDB will be released
 [patch.crates-io]
-exonum = { git = "https://github.com/exonum/exonum", rev = "2868b8b" }
-exonum-testkit = { git = "https://github.com/exonum/exonum", rev = "2868b8b" }
-exonum-configuration = { git = "https://github.com/exonum/exonum", rev = "2868b8b" }
-exonum-time = { git = "https://github.com/exonum/exonum", rev = "2868b8b" }
+exonum = { git = "https://github.com/exonum/exonum", rev = "b542dae" }
+exonum-testkit = { git = "https://github.com/exonum/exonum", rev = "b542dae" }
+#exonum-configuration = { git = "https://github.com/exonum/exonum", rev = "b542dae" }
```
vitvakatu: Exonum configuration is removed, so this line also can.

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,596 @@
...
+    }
+}
+
+impl Instance {
+    fn new(id: InstanceId, name: String) -> Self {
+        Self { id, name }
+    }
+}
+
+impl JavaArtifactId {
```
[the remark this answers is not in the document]
ipenkin: replaced by simple `String` wrapper

exonum-java-binding/core/rust/exonum-java/Cargo.toml
```diff
@@ -9,9 +9,9 @@ authors = ["Exonum team <exonum@bitfury.com>"]
 # in a correct application early, it is currently required in the application
 # because some public APIs in Java permit passing arbitrary pointers (e.g., Snapshot#newInstance).
 java_bindings = { path = "..", features = ["invocation", "resource-manager"] }
-exonum-configuration = "0.11"
+#exonum-configuration = "0.11"
```
vitvakatu: Same here, exonum-configuration no longer exists.

exonum-java-binding/core/rust/build.rs
ipenkin: It's left for future use. If not only custom protos will be required.

vitvakatu: Please remove, as there is no need in this code right now.

exonum-java-binding/core/rust/src/proto/mod.rs
```diff
@@ -0,0 +1,22 @@
+// Copyright 2018 The Exonum Team
```
vitvakatu: nit: 2019

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,434 @@
...
+            Err(Error::IncorrectArtifactId.into())
+        } else {
+            artifact
+                .name
+                .parse()
+                .map_err(|_| Error::IncorrectArtifactId.into())
+        }
+    }
+
+    fn parse_jni<T>(res: JniResult<T>) -> Result<T, ExecutionError> {
```
vitvakatu: This function hides error description. Can we avoid that?

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,434 @@
...
+        res.map_err(|err| match err.0 {
+            JniErrorKind::JavaException => Error::JavaException.into(),
+            _ => Error::UnspecifiedError.into(),
+        })
+    }
+
+    /// If the current node is a validator, returns `Some(validator_id)`, for other nodes return `None`.
+    fn validator_id(snapshot: &Snapshot, service_keypair: &(PublicKey, SecretKey)) -> Option<ValidatorId> {
+        CoreSchema::new(snapshot)
+            .actual_configuration()
```
vitvakatu: nit: It's not the scope of this PR, but `actual_configuration` is removed in latest DS.

cc @dmitry-timofeev

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,434 @@
...
+
+    fn parse_jni<T>(res: JniResult<T>) -> Result<T, ExecutionError> {
+        res.map_err(|err| match err.0 {
+            JniErrorKind::JavaException => Error::JavaException.into(),
+            _ => Error::UnspecifiedError.into(),
+        })
+    }
+
+    /// If the current node is a validator, returns `Some(validator_id)`, for other nodes return `None`.
+    fn validator_id(snapshot: &Snapshot, service_keypair: &(PublicKey, SecretKey)) -> Option<ValidatorId> {
```
vitvakatu: Why do you need to pass SecretKey to such function?

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,434 @@
...
+            .validator_keys
+            .iter()
+            .position(|validator| service_keypair.0 == validator.service_key)
+            .map(|id| ValidatorId(id as u16))
+    }
+
+    /// If the current node is a validator, return its identifier, for other nodes return `default`.
+    fn validator_id_or(
+        snapshot: &Snapshot,
+        service_keypair: &(PublicKey, SecretKey),
```
vitvakatu: See above ^, there is no need in passing `SecretKey`

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,434 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "JavaRuntimeProxy()")
+    }
+}
+
+impl From<Error> for ExecutionError {
+    fn from(value: Error) -> ExecutionError {
+        ExecutionError::new(
+            ErrorKind::runtime(value as u8),
+            format!("{:?}", value.clone()),
```
vitvakatu: This isn't a valid description, is it?

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,434 @@
...
+        }
+    }
+
+    fn parse_artifact(&self, artifact: &ArtifactId) -> Result<JavaArtifactId, ExecutionError> {
+        if artifact.runtime_id != Self::RUNTIME_ID as u32 {
+            Err(Error::IncorrectArtifactId.into())
+        } else {
+            artifact
+                .name
+                .parse()
```
vitvakatu: It's simple String, you don't need to `parse` it.

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,434 @@
...
+}
+
+impl From<ServiceRuntimeStateHashes> for StateHashAggregator {
+    fn from(value: ServiceRuntimeStateHashes) -> Self {
+        StateHashAggregator {
+            runtime: value.runtime(),
+            instances: value.instances(),
+        }
+    }
+}
```
vitvakatu: nit: no newline at the end of the file

exonum-java-binding/core/rust/src/proxy/runtime.rs
ipenkin: fixed in https://github.com/exonum/exonum-java-binding/pull/1129/commits/17a658560ef4f58bcc3359d4a8251b90002da62f

exonum-java-binding/core/rust/src/proxy/runtime.rs
ipenkin: replaced by `consensus_config()`

exonum-java-binding/core/rust/src/proxy/runtime.rs
ipenkin: fixed at https://github.com/exonum/exonum-java-binding/pull/1129/commits/17a658560ef4f58bcc3359d4a8251b90002da62f

exonum-java-binding/core/rust/src/proxy/runtime.rs
ipenkin: changed at https://github.com/exonum/exonum-java-binding/pull/1129/commits/a2fa7547c1d0fbc66cf7294f4ea1f90f7c47c617

exonum-java-binding/core/rust/src/proxy/node.rs
ipenkin: fixed at https://github.com/exonum/exonum-java-binding/pull/1129/commits/73162f5d3143280a6bdb9ecef1b278a7808a0648

`AnyTx` -> `Signed<AnyTx>` as argument from Java

exonum-java-binding/core/rust/src/proxy/runtime.rs
dmitry-timofeev: See also https://docs.oracle.com/en/java/javase/11/docs/specs/jni/design.html#java-exceptions , but remember that jni-rs detects exceptions for you.

Existing proxies do that, but they might also do something specific atop of it (not necessarily needed here).

exonum-java-binding/core/rust/src/runtime/error.rs
```diff
@@ -0,0 +1,31 @@
...
+/// List of possible Java runtime errors.
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, IntoExecutionError)]
+#[exonum(crate = "crate", kind = "runtime")]
+pub enum Error {
+    /// Unable to parse artifact identifier or specified artifact has non-empty spec.
+    IncorrectArtifactId = 0,
+    /// Checked java exception is occurred
+    JavaException = 1,
+    /// Unspecified error
+    UnspecifiedError = 2,
```
dmitry-timofeev: Too broad — I wouldn't like it to be abused. "OtherJniError" is how it is used,
    and I'd keep it such (better names are welcome).

exonum-java-binding/core/rust/src/runtime/java_runtime_factory.rs
```diff
@@ -32,15 +32,15 @@ const LOAD_ARTIFACT_SIGNATURE: &str = "(Ljava/lang/String;)Ljava/lang/String;";
 const CREATE_SERVICE_SIGNATURE: &str =
     "(Ljava/lang/String;)Lcom/exonum/binding/core/service/adapters/UserServiceAdapter;";
 
-/// Controls JVM and java service.
+/// Controls JVM and java runtime.
 #[allow(dead_code)]
 #[derive(Clone)]
-pub struct JavaServiceRuntime {
+pub struct JavaRuntimeFactory {
```
dmitry-timofeev: Better names? Maybe JavaSystemRuntime? @vitvakatu, will it survive your changes?
 Also, I don't see service_runtime to be used at all after initialization 💭

exonum-java-binding/core/rust/src/proxy/node.rs
```diff
@@ -68,35 +61,29 @@ impl NodeContext {
...
-            transaction.service_transaction(),
-            service_id,
-            self.blockchain.service_keypair.0,
-            &self.blockchain.service_keypair.1,
-        );
-        let tx_hash = msg.hash();
-
-        self.transaction_sender.broadcast_transaction(msg)?;
+    pub fn submit(&self, transaction: Verified<AnyTx>) -> Result<Hash, failure::Error> {
+        // TODO: check service is active
```
dmitry-timofeev: @alekseysidorov Why doesn't the core check that immediately when a transaction is submitted?

exonum-java-binding/core/rust/src/proxy/node.rs
```diff
@@ -124,10 +112,17 @@ pub extern "system" fn Java_com_exonum_binding_core_service_NodeProxy_nativeSubm
             &env,
             || -> JniResult<jbyteArray> {
                 let payload = env.convert_byte_array(payload)?;
-                let service_transaction =
-                    ServiceTransaction::from_raw_unchecked(transaction_id as u16, payload);
-                let raw_transaction = RawTransaction::new(service_id as u16, service_transaction);
-                match node.submit(raw_transaction) {
+                let tx = match Verified::from_bytes(payload.into()) {
```
dmitry-timofeev: This code cannot be correct because [Verified::from_bytes](https://github.com/exonum/exonum/blob/b542dae0bc10d8d8c015a347e9ce2c56201cac08/exonum/src/messages/signed.rs#L172-L177) expects a *message*,
but we pass just the payload. This method must create a transaction message and sign it with
the key of this node, ideally, using the core functionality, if it is available.

Second, please move all the logic inside Node#submit and test it. Keep here only
type conversion logic.

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,419 @@
...
+            let kind: ErrorKind = match err.kind() {
+                JniErrorKind::JavaException => Error::JavaException.into(),
+                _ => Error::UnspecifiedError.into(),
+            };
+            (kind, err).into()
+        })
+    }
+
+    /// If the current node is a validator, returns `Some(validator_id)`, for other nodes return `None`.
+    fn validator_id(snapshot: &dyn Snapshot, pub_key: &PublicKey) -> Option<ValidatorId> {
```
dmitry-timofeev: The code obtaining the ValidatorId is (still) unnecessarily complex. 

@alekseysidorov Why has the validator id been removed from afterCommit arguments? Shan't the core then provide an easy way to obtain it? Or shall we remove it as well?

exonum-java-binding/core/rust/integration_tests/tests/service_runtime.rs
```diff
@@ -124,8 +124,8 @@ fn create_service_for_unknown_artifact() {
 }
 
 // Creates a new instance of JavaServiceRuntime for same JVM.
-fn get_runtime() -> JavaServiceRuntime {
-    JavaServiceRuntime::create_with_jvm(VM.clone(), 0)
+fn get_runtime() -> JavaRuntimeFactory {
```
dmitry-timofeev: Comment is not updated. Though the first question is what JavaRuntimeFactory is (see the discussion there).

exonum-java-binding/core/rust/src/runtime/java_runtime_factory.rs
vitvakatu: I haven't decided yet, but it looks like this structure will be removed (it can be replaced with a few functions, as there is no need in saving any state). Though I think `JavaRuntimeFactory` is a good name for a some thing that creates JavaRuntimeProxy.

dmitry-timofeev:
>    Exceptions handling in all RuntimeProxy methods. parse_jni returns error to Rust, but doesn't clear JVM exception (see https://github.com/exonum/exonum-java-binding/pull/1129/files#r328678546)

Yes, including handling TransactionExecutionException in execute (see previous TransactionProxy). However, as the code is structured, `parse_jni` cannot clear them — it must happen inside closures that have a JNIEnv. Service and Transaction proxies did that, but one must copy that blindly but review thoroughly as they also did something on top, not necessarily needed in the runtime.

>    api_endpoints method does nothing and should be removed (replaced with blank implementation). Alternative method notify_api_change must be implemented in the following PRs)

Well, it could be implemented separately, but then please submit a separate task, as this functionality is part of the Runtime.

exonum-java-binding/core/rust/src/runtime/java_runtime_factory.rs
dmitry-timofeev: Ok, let's keep as is, maybe with a todo referencing your task.

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,419 @@
...
+}
+
+impl From<ServiceRuntimeStateHashes> for StateHashAggregator {
+    fn from(value: ServiceRuntimeStateHashes) -> Self {
+        StateHashAggregator {
+            runtime: value.runtime(),
+            instances: value.instances(),
+        }
+    }
+}
```
skletsun:
```suggestion
}

From<JavaRuntimeProxy> for (u32, Box<dyn Runtime>)
    fn from(r: JavaRuntimeProxy) -> Self {
        (JavaRuntimeProxy::RUNTIME_ID as u32, Box::new(r))
    }
```

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,452 @@
...
+    fn notify_api_changes(&self, context: &ApiContext, changes: &[ApiChange]) {
+        let added_instances_ids: Vec<i32> = changes.iter()
+            .filter_map(|change| {
+                if let ApiChange::InstanceAdded(instance_id) = change {
+                    Some(*instance_id as i32)
+                } else {
+                    None
+                }
+            })
+            .collect();
```
dmitry-timofeev: Please see the docs of the runtime method: added_instances_ids must not be empty (connect_apis shall not be invoked if no new services need to be connected). That is a temporary requirement till proper Node handling is implemented.

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,452 @@
...
+        let node = NodeContext::new(self.exec.clone(), context.clone());
+
+        unwrap_jni(self.exec.with_attached(|env| {
+            let node_handle = to_handle(node);
+            let ids_array = env.new_int_array(added_instances_ids.capacity() as i32)?;
```
dmitry-timofeev: It is incorrect: capacity is not a number of elements.

exonum-java-binding/core/rust/src/proxy/runtime.rs
dmitry-timofeev: No, there are still two code fragments doing 
```
            .iter()
            .map(|bytes| Hash::from_bytes(bytes.into()).unwrap())
            .collect()
```

exonum-java-binding/core/rust/src/utils/errors.rs
```diff
@@ -70,6 +70,30 @@ pub fn check_error_on_exception<T>(env: &JNIEnv, result: JniResult<T>) -> Result
...
+/// Handles and clears any Java exceptions.
+///
+/// Any JNI errors are logged with their descriptions,  for JNI errors
+/// like `JniErrorKind::JavaException` it gets (and clears) any exception
+/// that is currently being thrown.
+///
+/// Panics:
+/// - Panics if JNI error is `JniErrorKind::JavaException` but couldn't get (and clear) exception
+///   object and describe it.
+pub fn log_jni_error_or_exception<T>(env: &JNIEnv, result: JniResult<T>) -> JniResult<T> {
```
dmitry-timofeev: I'd consider renaming to communicate that it not just logs (it is a secondary responsibility, but handles/clears the exception, if any).
            handle_exception/clear_exception?

exonum-java-binding/core/rust/src/proxy/runtime.rs
dmitry-timofeev: @alekseysidorov 🏓

exonum-java-binding/core/rust/src/proxy/node.rs
dmitry-timofeev: @alekseysidorov 🏓

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,452 @@
...
+        }));
+
+        ServiceRuntimeStateHashes::from_bytes(bytes.into())
+            .unwrap()
+            .into()
+    }
+
+    fn before_commit(&self, _dispatcher: &DispatcherRef, fork: &mut Fork) {
+        unwrap_jni(self.exec.with_attached(|env| {
+            let view_handle = to_handle(View::from_ref_fork(fork));
```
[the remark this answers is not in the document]
vitvakatu: Good catch, yes. It will fail at runtime now. `View::from_ref_mut_fork` or whatever-I-called-that-method.

exonum-java-binding/core/rust/src/proxy/runtime.rs
skletsun: Sorry - somehow I lost parts of the code when copied it and didn't noticed that 😟. Now it's correct.

skletsun:
Since this branch is already used as a base for a few other branches I'd suggest changing `0.8.0-SNAPSHOT` everywhere in code/configs to `0.9.0-SNAPSHOT`.

dmitry-timofeev:
> Since this branch is already used as a base for a few other branches I'd suggest changing 0.8.0-SNAPSHOT everywhere in code/configs to 0.9.0-SNAPSHOT.

It must **not** be done here manually — the branch just has to be synced with dynamic-services.

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,561 @@
...
+    ///
+    /// Any JNI errors are converted into `ExecutionError` with their descriptions, for JNI errors
+    /// like `JniErrorKind::JavaException` it gets (and clears) any exception that is currently
+    /// being thrown, then exception is passed to `ExceptionHandlers::DEFAULT`.
+    fn handle_any_exception<R>(
+        env: &JNIEnv,
+        result: JniResult<R>,
+    ) -> Result<R, ExecutionError>
+    {
+        result.map_err(|err| {
```
vitvakatu: redundant code duplication, it's enough to call `handle_exception` with empty array

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,561 @@
...
+
+    /// Handles and clears any Java exceptions.
+    ///
+    /// Any JNI errors are converted into `ExecutionError` with their descriptions, for JNI errors
+    /// like `JniErrorKind::JavaException` it gets (and clears) any exception that is currently
+    /// being thrown, then exception is passed to corresponding `ExceptionHandler` according their
+    /// type and `exception_handlers` mapping.
+    /// `ExceptionHandlers::DEFAULT` is called in case of there is no any handlers or handlers are
+    /// not matched to exception type.
+    fn handle_exception<H, R>(
```
vitvakatu: So now we have three methods `convert_jni`, `handle_exception` and `handle_any_exception` with pretty similar semantics and purpose. Can we reduce it?

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,561 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "JavaRuntimeProxy()")
+    }
+}
+
+/// Artifact identification properties within `JavaRuntimeProxy`
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct JavaArtifactId(String);
+
+impl ToString for JavaArtifactId {
```
skletsun: According to docs from the standard library this trait should not be implemented directly:

>  This trait is automatically implemented for any type which implements the
[`Display`] trait. As such, `ToString` shouldn't be implemented directly:
[`Display`] should be implemented instead, and you get the `ToString`
implementation for free.

exonum-java-binding/core/rust/Cargo.toml
```diff
@@ -16,24 +16,40 @@ invocation = ["jni/invocation"]
...
+[build-dependencies]
+exonum-build = { version = "0.12" }
+
 [profile.dev]
 rpath = true
 
 [profile.release]
 rpath = true
+
+# FIXME: using git dependency until new Exonum with MerkleDB will be released
```
vitvakatu:
```suggestion
# FIXME: using git dependency until new Exonum with dynamic services will be released
```

exonum-java-binding/core/rust/exonum-java/Cargo.toml
```diff
@@ -9,8 +9,7 @@ authors = ["Exonum team <exonum@bitfury.com>"]
 # in a correct application early, it is currently required in the application
 # because some public APIs in Java permit passing arbitrary pointers (e.g., Snapshot#newInstance).
 java_bindings = { path = "..", features = ["invocation", "resource-manager"] }
-exonum-configuration = "0.12"
-exonum-btc-anchoring = "0.12"
+#exonum-btc-anchoring = "0.11" # commented out until anchoring switches to merkledb
```
vitvakatu: nit: can remove, I'll take care of it

exonum-java-binding/core/rust/src/proxy/node.rs
```diff
@@ -69,34 +63,28 @@ impl NodeContext {
...
+        let signed = SignedMessage::new(
+            tx.to_bytes(),
+            pub_key.to_owned(),
+            secret_key,
         );
-        let tx_hash = msg.hash();
 
-        self.transaction_sender.broadcast_transaction(msg)?;
+        let tx_hash = signed.object_hash();
+        // TODO: check Core behaviour/any errors on service inactivity
```
vitvakatu:
```suggestion
        // TODO(ECR-3679): check Core behaviour/any errors on service inactivity
```

exonum-java-binding/core/rust/src/proxy/node.rs
```diff
@@ -105,34 +93,39 @@ impl NodeContext {
...
+                };
+
+                match node.submit(tx) {
                     Ok(tx_hash) => convert_hash(&env, &tx_hash),
                     Err(err) => {
-                        // node#submit can fail for two reasons: unknown transaction id and
-                        // an error in ApiSender#send. The former is the service implementation
-                        // error; the latter is an internal, unrecoverable error.
+                        // node#submit can fail on an error in ApiSender#send.
+                        // The former is the service implementation error;
```
vitvakatu: There is no `former` and `latter`.

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,537 @@
...
+        snapshot: &dyn Snapshot,
+        pub_key: &PublicKey,
+        default: i32
+    ) -> i32 {
+        Self::validator_id(snapshot, pub_key)
+            .map_or(default, |id| i32::from(id.0))
+    }
+
+    /// Converts `JniResult<R>` into `Result<R, ExecutionError>`.
+    fn convert_jni<R>(res: JniResult<R>) -> Result<R, ExecutionError> {
```
vitvakatu: It looks like `convert_jni` is exactly a `handle_exception`, but without exception handling. So in each method we return `JniResult<Result<R, ExecutionError>>` and then pass it to `convert_jni`, which has a signature of `JniResult<Result<R, ExecutionError>> -> Result<R, ExecutionError>`. So wouldn't it make sense to remove `convert_jni`, but instead use `handle_exception(JniResult<R>)` on every result returned from `self.exec.with_attached`?

vitvakatu: See comment below, it would need to change signature of `handle_exception` to `handle_exception(&Executor, JniResult<R>, Handlers)`

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,537 @@
...
+                JavaType::Primitive(Primitive::Void),
+                &[
+                    JValue::from(id),
+                    JValue::from(view_handle),
+                    JValue::from(params),
+                ],
+            ).and_then(JValue::v);
+            Ok(Self::handle_exception::<&ExceptionHandler, ()>(env, jni_res, &[]))
+        });
+        Self::convert_jni(result)?
```
vitvakatu:
```diff
-          Ok(Self::handle_exception::<&ExceptionHandler, ()>(env, jni_res, &[]))
+          jni_res
        });
-      Self::convert_jni(result)?
+      Self::handle_exception::<&ExceptionHandler, ()>(&self.exec, result, &[])
```

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,537 @@
...
+    /// type and `exception_handlers` mapping.
+    /// `ExceptionHandlers::DEFAULT` is called in case of there is no any handlers or handlers are
+    /// not matched to exception type.
+    fn handle_exception<H, R>(
+        env: &JNIEnv,
+        result: JniResult<R>,
+        exception_handlers: &[(&GlobalRef, H)],
+    ) -> Result<R, ExecutionError>
+        where
+            H: Fn(&JNIEnv, JObject) -> ExecutionError,
```
vitvakatu:
```suggestion
            H: ExecutionHandler,
```

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,537 @@
...
+
+    const SERVICE_LOADING: &'static ExceptionHandler = &|env, exception| {
+        assert!(!exception.is_null(), "No exception thrown.");
+        let msg = unwrap_jni(get_exception_message(env, exception));
+        let desc = format!("ServiceLoadingException(message: {:?})", msg);
+        (Error::JavaException, desc).into()
+    };
+
+    fn get_tx_error_code(env: &JNIEnv, exception: JObject) -> JniResult<i8> {
+        let err_code = env.call_method(exception, "getErrorCode", "()B", &[])?;
```
vitvakatu: I think we need to cache this method, as it can be called really a lot.

exonum-java-binding/core/rust/src/proxy/runtime.rs
ipenkin: Full trait spec is used due to `error[E0404]: expected trait, found type alias 'ExceptionHandler'`.

Also this change is required to fix it: `trait ExceptionHandler = Fn(&JNIEnv, JObject) -> ExecutionError;` instead of using `type` keyword. But trait aliases are experimental https://github.com/rust-lang/rust/issues/41517

exonum-java-binding/core/rust/src/proxy/node.rs
```diff
@@ -69,34 +63,28 @@ impl NodeContext {
...
+            pub_key.to_owned(),
+            secret_key,
         );
-        let tx_hash = msg.hash();
 
-        self.transaction_sender.broadcast_transaction(msg)?;
+        let tx_hash = signed.object_hash();
+        // TODO(ECR-3679): check Core behaviour/any errors on service inactivity
+        self.api_context.sender().broadcast_transaction(
+            signed.into_verified()?
```
dmitry-timofeev: I don't think we shall *verify* a transaction message that has been just created.
            Can we use from_value instead?

exonum-java-binding/core/rust/src/proxy/runtime.rs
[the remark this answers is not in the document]
dmitry-timofeev: from_ref_mut_fork

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,542 @@
...
+
+    const TX_EXECUTION: &'static ExceptionHandler = &|env, exception| {
+        assert!(!exception.is_null(), "No exception thrown.");
+        let code = unwrap_jni(Self::get_tx_error_code(env, exception)) as u8;
+        let msg = unwrap_jni(get_exception_message(env, exception))
+            .unwrap_or(String::new());
+        ExecutionError::new(ErrorKind::service(code), msg)
+    };
+
+    const SERVICE_LOADING: &'static ExceptionHandler = &|env, exception| {
```
dmitry-timofeev: Do we need to handle service loading exception specially (it is converted to JavaException just as the default)? Is there any benefit in defining more specific runtime errors or will JavaException do in this case?

exonum-java-binding/core/rust/src/proxy/node.rs
alekseysidorov: Dispatcher finds proper runtime for transaction by their instance ID in lookup table, so if runtime has not specified service instance that means that there is a bug in dispatcher or runtime.

exonum-java-binding/core/rust/src/proxy/runtime.rs
alekseysidorov: In old implementation after commit context contains parsed blockchain configuration so seems a little bit redundant in new code. Without this configuration impossible to get validator ID, so you have to get it explicitily from information schema to get validator ID.

dmitry-timofeev: Shall the runtime provide the validator id in after_commit? Do services need to distinguish if they are validators or auditors, which the validator id enables to do?

alekseysidorov: Runtime should not provide any consensus configuration and similar things so as not to break encapsulation.
In the end, the consensus algorithm may be different.

dmitry-timofeev: @alekseysidorov @popzxc 

Does Exonum keep supporting submitting txs in services? If I understand correctly, services running on auditors must not generally do that. How are they supposed to determine if they run on auditors or not?

exonum-java-binding/core/rust/integration_tests/tests/service_runtime_bootstrap.rs
```diff
@@ -1,59 +0,0 @@
...
-extern crate java_bindings;
-
-use integration_tests::vm::{fakes_classpath, java_library_path, log4j_path};
-use java_bindings::{Config, InternalConfig, JavaServiceRuntime, JvmConfig, RuntimeConfig};
-
-#[test]
-// Fails on Java 12. Ignored until [ECR-3133] is fixed because the cause of the issue also prevents
-// the execution of system tests.
-#[ignore]
-fn bootstrap() {
```
dmitry-timofeev: Why is this one removed?

exonum-java-binding/core/rust/exonum-java/src/node_builder.rs
```diff
@@ -81,9 +78,7 @@ fn prepare_service_factories<P: AsRef<Path>>(path: P) -> Vec<Box<dyn ServiceFact
 
 fn system_service_factory_for_name(name: &str) -> Box<dyn ServiceFactory> {
     match name {
-        CONFIGURATION_SERVICE => Box::new(ConfigurationServiceFactory) as Box<dyn ServiceFactory>,
-        BTC_ANCHORING_SERVICE => Box::new(BtcAnchoringServiceFactory) as Box<dyn ServiceFactory>,
```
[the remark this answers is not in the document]
dmitry-timofeev: @ipenkin Have you submitted the issue?

exonum-java-binding/core/src/test/java/com/exonum/binding/core/blockchain/CoreSchemaProxyIntegrationTest.java
```diff
@@ -30,22 +30,27 @@
...
 import java.util.function.Consumer;
+
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 @RequiresNativeLibrary
 class CoreSchemaProxyIntegrationTest {
 
   @Test
+  @Disabled("ECR-3591")
```
dmitry-timofeev: These test failures are unrelated to ECR-3591. Please use ECR-3709.

exonum-java-binding/core/rust/exonum-java/src/node_builder.rs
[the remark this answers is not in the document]
ipenkin: ECR-3714 was submitted

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,538 @@
...
+                self.runtime_adapter.as_obj(),
+                runtime_adapter::deploy_artifact_id(),
+                JavaType::Primitive(Primitive::Void),
+                &[
+                    JValue::from(artifact_id),
+                    JValue::from(spec),
+                ],
+            ).and_then(JValue::v);
+
+            Ok(Self::handle_exception::<&ExceptionHandler, ()>(env, jni_res, &[]))
```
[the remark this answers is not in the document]
ipenkin: done

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -0,0 +1,538 @@
...
+            .map_or(default, |id| i32::from(id.0))
+    }
+
+    /// Converts `JniResult<R>` into `Result<R, ExecutionError>`. Need to be used after all
+    /// other checks. It doesn't panic unlike `utils::unwrap_jni`.
+    fn convert_jni<R>(res: JniResult<R>) -> Result<R, ExecutionError> {
+        res.map_err(|err|
+            (
+                Error::OtherJniError,
+                format!("JNI error in not managed code fragment: {:?}", err)
```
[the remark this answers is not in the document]
ipenkin: is not actual
comment was added to new impl

exonum-java-binding/core/rust/integration_tests/tests/service_runtime_bootstrap.rs
ipenkin: was reverted back

exonum-java-binding/core/rust/exonum-java/src/node_builder.rs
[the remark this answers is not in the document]
dmitry-timofeev: Good, thanks!

exonum-java-binding/core/rust/src/proxy/runtime.rs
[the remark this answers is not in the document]
dmitry-timofeev: Thanks, but that's different from what I suggested, and somewhat complicated, but let's resolve it later: ECR-3715.

PR #1136, opened 2019-09-30
Update time service to dynamic services [ECR-3597]

dmitry-timofeev:
## Overview
Update time service to dynamic services:

- Support multiple instances of time oracle. User must always specify
the name of the time oracle they intend to use.
- Use the runtime index to check if the service with the given name
is active.

---
See: https://jira.bf.local/browse/ECR-3597

### Definition of Done

- [x] Public API has Javadoc


dmitry-timofeev:
The tests depend on the testkit, otherwise, it must be ready.

exonum-java-binding/time-oracle/src/main/java/com/exonum/binding/time/TimeSchemaProxy.java
```diff
@@ -16,74 +16,74 @@
...
+import com.exonum.binding.messages.Runtime.ArtifactId;
+import com.exonum.binding.messages.Runtime.InstanceSpec;
 import java.time.ZonedDateTime;
 
 class TimeSchemaProxy implements TimeSchema {
 
-  static {
-    LibraryLoader.load();
-  }
+  private static final int RUST_RUNTIME_ID = 0;
```
vitvakatu: Shouldn't this constant be somewhere in core?

dmitry-timofeev: If it gets re-used, then yes; but currently can't see a good home for it (a new enum?)

MakarovS: Enum in core sounds good, we'll probably need it somewhere else

exonum-java-binding/time-oracle/src/main/java/com/exonum/binding/time/TimeSchemaProxy.java
```diff
@@ -16,74 +16,74 @@
...
+    this.name = name;
+    this.view = view;
     checkIfEnabled();
-    this.dbView = dbView;
+  }
+
+  private void checkIfEnabled() {
+    MapIndex<String, InstanceSpec> serviceInstances = new RuntimeSchema(view).serviceInstances();
+    checkArgument(serviceInstances.containsKey(name), "No service instance "
+        + "with the given name (%s) started.");
```
MakarovS: `    checkArgument(serviceInstances.containsKey(name), "No service instance "
        + "with the given name (%s) started.", name);`

exonum-java-binding/time-oracle/src/main/java/com/exonum/binding/time/TimeSchemaProxy.java
MakarovS: We could use that enum (and value of `JAVA_RUNTIME_ID`) in TestKit, for valid artifact ids creation (in case we don't want to do that in JNI - https://github.com/exonum/exonum-java-binding/pull/1161#discussion_r334222288 )

dmitry-timofeev: We could consider creating CoreArtifactId (with the runtime id) and ServiceArtifactId (with no runtime id, to be used in metadata) so that the native does not have to perform these conversions. Or, drop the latter and always use four-component from `<runtime_id>:<group_id>:...` (even in JAR manifest).

dmitry-timofeev: Or, drop the requirements on the format and just require `<runtime_id>:<artifact_id>`

dmitry-timofeev: Now references RuntimeId.RUST.id

dmitry-timofeev:
Merging as time oracle tests are with #1215 as a 🐔 and 🥚

PR #1174, opened 2019-10-22
Updated how transaction result is represented [ECR-3693]

dmitry-timofeev:
## Overview

Updated how transaction result is represented. This PR uses an alternative format with an enum to represent the error kind, yet to be done in the core (ECR-3717).

Dropped the internal type representing the transaction result. It has both pros and cons:
- (+) No need for an extra value type(s) and conversion code
- (?) Directly serializable in protobuf
- (-) Less-than-optimal interface and documentation of the generated code.
- (-) Less freedom to hide any complexities and/or provide better interface.
- (-) Worse JSON compatibility.

---
See: ECR-3693

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/common/src/main/proto/runtime.proto
```diff
@@ -59,6 +59,24 @@ message ExecutionError {
...
+    RUNTIME = 2;
+    SERVICE = 3;
+  }
+  ErrorKind kind = 1;
+  /// User defined error code that can have different meanings for the different
+  /// error kinds.
+  uint32 code = 2;
+  /// Optional description which doesn't affect `object_hash`.
+  string description = 3;
+}
```
MakarovS: Line

exonum-java-binding/common/src/test/java/com/exonum/binding/common/blockchain/ExecutionStatusesTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+  @ParameterizedTest
+  @ValueSource(ints = {-1, -2, Integer.MIN_VALUE})
+  void serviceErrorRejectsNegative(int code) {
+    IllegalArgumentException e = assertThrows(IllegalArgumentException.class,
+        () -> ExecutionStatuses.serviceError(code, ""));
+
+    assertThat(e).hasMessageContaining(String.valueOf(code));
+  }
+}
```
MakarovS: Line

exonum-java-binding/common/src/test/java/com/exonum/binding/common/blockchain/ExecutionStatusesTest.java
```diff
@@ -0,0 +1,71 @@
...
+    assertThat(error.getKind()).isEqualTo(ErrorKind.SERVICE);
+    assertThat(error.getCode()).isEqualTo(code);
+    assertThat(error.getDescription()).isEqualTo(description);
+  }
+
+  private static List<Arguments> errors() {
+    return ImmutableList.of(
+        arguments(0, "Min error code"),
+        arguments(1, ""),
+        arguments(Integer.MAX_VALUE, "Max error code")
```
MakarovS: Did core remove `255` as maximum error code? Can't find such a restriction in native anymore.

exonum-java-binding/common/src/main/proto/runtime.proto
```diff
@@ -65,10 +65,28 @@ message ExecutionError {
 message ExecutionStatus {
   oneof result {
     google.protobuf.Empty ok = 1;
-    ExecutionError error = 2;
+    ExecutionError2 error = 2;
   }
 }
 
+// fixme: A temporary replacement until the message is fixed in the core:
```
bullet-tooth: Probably it would be better to use `TODO`(+ ticket link, if known) instead of `fixme` because IDEA has special indexing of that.

dmitry-timofeev: Replaced with protos from the core.

exonum-java-binding/common/src/test/java/com/exonum/binding/common/blockchain/ExecutionStatusesTest.java
dmitry-timofeev: Actually, it is still there, as Service error variant accepts only u8 as a code.

PR #1177, opened 2019-10-23
Optimization with protobuf files [ECR-3472]

skletsun:
## Overview
Protobuf specification files are taken now from a separate repository that is incorporated into the main project as a git submodule. `git submodule init` и `git submodule update` should be executed in order to fetch the content of the submodule.

---
See: https://jira.bf.local/browse/ECR-3472

### Definition of Done

- [x] Public API has Javadoc


CONTRIBUTING.md
```diff
@@ -36,6 +37,8 @@ You need to install the following dependencies:
   Also on Mac you need a [`coreutils`](https://formulae.brew.sh/formula/coreutils) package installed.
 
 ### Building
+__Important__: In order to retrieve submodules you have to execute `git submodule update --init` before first build.
```
vitvakatu: You also need to execute `git submodule update` before _each_ build, don't you?

skletsun: No, the info about submodule revision is recorded under the main repo, so I believe that it makes sense executing this only after fetching updates for the main repo. And yes, it should also be stated somewhere.

dmitry-timofeev: fetching updates or switching to any new branch?

skletsun: When one creates a new branch within the current branch with submodules that are checked out, the new branch contains the same revision to submodules and the content of submodules folders remains the same.
When one gets switched to the branch that points to a different submodule revision then `git submodule update` is required I guess.
Well, the deeper we go here the more I feel like instruction turns into a guide to git for the user - wouldn't it be enough to just mention that protobuf files (or just some parts of the project) are a git submodule and user should pay attention to this?

dmitry-timofeev: Thanks, submodule mention shall be enough.

PR #1178, opened 2019-10-23
Remove non owned handlers [ECR-910]

vitvakatu:
## Overview

---
See: https://jira.bf.local/browse/ECR-910

### Definition of Done

- [ ] Public API has Javadoc


exonum-java-binding/core/rust/src/handle/resource_manager/imp.rs
```diff
@@ -129,25 +129,6 @@ pub fn remove_handle<T: 'static>(handle: Handle) {
...
 }
 
-/// Adds native-owned handle to the resource manager.
-///
-/// # Panics
-///
-/// See `add_handle_impl` for the details.
-pub fn register_handle<T: 'static>(handle: Handle) {
-    assert_ne!(handle, 0);
-    add_handle_impl::<T>(handle, HandleOwnershipType::NativeOwned);
```
dmitry-timofeev: NativeOwned — is this ownership type used anywhere? Shan't we revert all changes from 8c395c571b1a9097c217a2920b010fd466865b8b?

exonum-java-binding/core/rust/src/handle/mod.rs
```diff
@@ -68,13 +39,6 @@ pub fn to_handle<T: 'static>(val: T) -> Handle {
     handle
 }
 
-/// Returns a handle (a raw pointer) to the given native-owned object. This handle should not be
-/// freed manually.
-pub fn as_handle<T>(val: &mut T) -> NonOwnedHandle<T> {
```
dmitry-timofeev: I remember it was used in Node ITs — is that correct that it is not needed for newer, yet to be implemented tests?

exonum-java-binding/core/rust/src/handle/resource_manager/imp.rs
vitvakatu: Thanks, removed all that stuff

exonum-java-binding/core/rust/src/handle/mod.rs
vitvakatu: Not needed, as NodeContext is no more native-owned structure, but rather java-owned.

dmitry-timofeev: Yeah, Node aspires to become one (ECR-2334) 😄

exonum-java-binding/core/rust/src/handle/mod.rs
```diff
@@ -17,13 +17,9 @@
 //! Wrappers and helper functions around Java pointers. Used for memory management
 //! between native and Java.
 
-// TODO Remove `allow(dead_code)` after [https://jira.bf.local/browse/ECR-910].
-#![allow(dead_code)]
-
 use jni::sys::jlong;
 use jni::JNIEnv;
```
skletsun: NIT: I'd also nest all `use` statements since we once decided to do so when modifying "old" files.

PR #1181, opened 2019-10-23
Add serviceName and serviceId to TransactionContext [ECR-3639]

MakarovS:
## Overview
Add serviceName and serviceId to TransactionContext.

---
See: https://jira.bf.local/browse/ECR-3639

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntimeAdapter.java
```diff
@@ -157,10 +157,13 @@ void executeTransaction(int serviceId, int txId, byte[] arguments,
       Fork fork = viewFactory.createFork(forkNativeHandle, cleaner);
       HashCode hash = HashCode.fromBytes(txMessageHash);
       PublicKey authorPk = PublicKey.fromBytes(authorPublicKey);
+      String serviceName = serviceRuntime.getServiceNameById(serviceId);
```
dmitry-timofeev: I remember we discussed to push the instantiation of TransactionContext inside the runtime because it already has this knowledge, the adapter will not have to request if from the runtime (and to be appropriately tested), and the runtime — provide an accessor for that. Is there a reason for the other approach?

dmitry-timofeev:
> Q: Do we actually need to include service id? Shouldn't service name be sufficient for our purposes, as it's always known to the user (unlike service id)?

The present TransactionConverter does not include the service id (com.exonum.binding.core.service.TransactionConverter#toTransaction), so if the transaction execution logic does need the id in some operations, the converter will have to be supplied with that id, or the tx execution logic will have to access it elsewhere.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/transaction/InternalTransactionContext.java
```diff
@@ -21,6 +21,7 @@
 import com.exonum.binding.common.crypto.PublicKey;
 import com.exonum.binding.common.hash.HashCode;
 import com.exonum.binding.core.storage.database.Fork;
+import java.util.Objects;
 
 /**
  * Default implementation of the transaction context.
```
dmitry-timofeev: Would it make sense to use `@AutoValue` here?

exonum-java-binding/core/src/main/java/com/exonum/binding/core/transaction/InternalTransactionContext.java
```diff
@@ -65,4 +66,25 @@ public String getServiceName() {
   public Integer getServiceId() {
     return serviceId;
   }
+
+  @Override
+  public boolean equals(Object o) {
```
dmitry-timofeev: I'd consider testing with EqualsVerifier.verify() to ensure it remains correct)

exonum-java-binding/core/src/main/java/com/exonum/binding/core/transaction/InternalTransactionContext.java
MakarovS: Sure, added `@AutoValue`.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntimeAdapter.java
```diff
@@ -146,7 +146,8 @@ void stopService(int id) {
    * @param txMessageHash the hash of the transaction message
    * @param authorPublicKey the public key of the transaction author
    * @throws TransactionExecutionException if the transaction execution failed
-   * @see ServiceRuntime#executeTransaction(Integer, int, byte[], TransactionContext)
+   * @see ServiceRuntime#executeTransaction(Integer, int, byte[], TransactionContext, Fork,
```
bullet-tooth:
```suggestion
   * @see ServiceRuntime#executeTransaction(Integer, int, byte[], Fork,
```

exonum-java-binding/core/src/main/java/com/exonum/binding/core/transaction/TransactionContext.java
```diff
@@ -50,6 +50,18 @@
...
+   * Returns the name of the service instance.
+   */
+  String getServiceName();
+
+  /**
+   * Returns the numeric id of the service instance.
+   *
+   * @see TransactionMessage#getServiceId()
+   */
+  Integer getServiceId();
```
bullet-tooth: can it be `null`? If not why doesn't `int`?

MakarovS: I intentionally used `Integer` here, so that if `serviceId` isn't specified, NPE would be thrown by AutoValue when creating an instance. Opposed to `int`, which would just set this value to `0` by default, which is error-prone. Should I keep the `Integer` or change it to `int` and validate that it was set or change it to `int` and not bother with that?

dmitry-timofeev: But the builder checks (or shall check) that, doesn't it?

dmitry-timofeev: The builder implementation nuances (that it uses Integer to distinguish between a set and unset value) must not affect the API.

PR #1186, opened 2019-10-25
Use artifact name in native API: [ECR-3733]

dmitry-timofeev:
## Overview

Clarified the usage of artifact name in the native API;
also made ServiceArtifactId compatible with any valid
artifact id.

Verification of correct Java artifact id are pushed to the loader.

Also a RuntimeId enum is added with well-known service runtime ids.

---
See: **https://jira.bf.local/browse/ECR-3733**

### Definition of Done

- [x] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/RuntimeId.java
+++ b/exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/RuntimeId.java
@@ -17,7 +17,7 @@
 package com.exonum.binding.core.runtime;
 
 /**
- * Represents a well-known runtime ids, as assigned by the Exonum core.
+ * Represents well-known runtime ids, as assigned by the Exonum core.
  */
 public enum RuntimeId {
   RUST(0),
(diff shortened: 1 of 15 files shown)


exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/RuntimeId.java
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.exonum.binding.core.runtime;
+
+/**
+ * Represents a well-known runtime ids, as assigned by the Exonum core.
```
MakarovS: Why `well-known`?

dmitry-timofeev: Because the core hard-codes these ids in its source, they are not configurable.

PR #1189, opened 2019-10-28
Store and report type names in the resource manager

DarkEld3r:
## Overview

Type names have been added to the resource manager diagnostics.

---

### Definition of Done

- [ ] Public API has Javadoc


exonum-java-binding/core/rust/src/handle/resource_manager/imp.rs
```diff
@@ -94,9 +99,12 @@ fn check_handle_impl<T: 'static>(handle: Handle, ownership: Option<HandleOwnersh
...
             assert_eq!(
-                info.object_type, actual_object_type,
-                "Wrong type id for '{:X}' handle",
-                handle
+                info.object_type,
+                actual_object_type,
+                "Wrong type id for '{:X}' handle, expected '{}', actual '{}'",
+                handle,
+                info.type_name,
+                any::type_name::<T>(),
```
dmitry-timofeev: https://doc.rust-lang.org/std/any/fn.type_name.html — is it 1.38.0 new feature (is that grey box a 'since' field)? We are stuck on 1.36.0 at the moment.

skletsun: Yes
```
#[stable(feature = "type_name", since = "1.38.0")]
#[rustc_const_unstable(feature = "const_type_name")]
pub const fn type_name<T: ?Sized>() -> &'static str {
```

DarkEld3r: Yes, it is new feature unfortunately.

dmitry-timofeev: OK, thanks, then we will have to wait till a newer working compiler is available 🙃

dmitry-timofeev: \+ @vitvakatu 

It seems it is no longer blocked since #1429.

> A note to the one who will be syncing ~~dynamic-services~~ this PR with master: dynamic-services got rid of handles on native-owned objects (#1178), and of HandleInfo as well, hence this file will have to be merged manually.

I also wonder if it will make sense to keep string representations separately to avoid their duplication in handles of the same type.

DarkEld3r: I can easily miss something, but isn't "resource manager" still a debug tool? I guess such optimization isn't necessary in that case.

Though it can be interesting to measure overhead in the typical use-case.

dmitry-timofeev: > I can easily miss something, but isn't "resource manager" still a debug tool? I guess such optimization isn't necessary in that case.

Yes, it is primarily a debug tool (actually, a useful one — we recently had an issue with invalid handles during development 🙃). However, there are currently quite some publicly accessible APIs that allow to pass anything as handle to native from the user code. A normal user has absolutely no reason to use these API and likely never won't, but someone _might_. On the other hand, I don't know if that shall make us keep the rm in production builds 🤔

exonum-java-binding/core/rust/src/handle/resource_manager/imp.rs
```diff
@@ -65,12 +88,15 @@ pub fn check_handle<T: 'static>(handle: Handle) {
...
-        Some(type_id) => {
+        Some(info) => {
             let actual_object_type = TypeId::of::<T>();
             assert_eq!(
-                *type_id, actual_object_type,
-                "Wrong type id for '{:X}' handle",
-                handle
+                info.object_type,
+                actual_object_type,
+                "Wrong type id for '{:X}' handle, expected '{}', actual '{}'",
```
dmitry-timofeev: Are these included automatically in case of failure? I wonder if two _different_ types might have the same string representation (type_name)?

DarkEld3r: Yes, they [can](https://doc.rust-lang.org/std/any/fn.type_name.html): 
> The type name should not be considered a unique identifier of a type; multiple types may share the same type name.

Do you suggest to print a different message if the string representation is the same?

dmitry-timofeev: No, I think that would be redundant, I would just include the unique numeric ids.

dmitry-timofeev: (if they are not already automagically included)

DarkEld3r: They are included. The error message will look this way:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `TypeId { t: 16712579856250238426 }`,
 right: `TypeId { t: 7392132643490421655 }`: Wrong type id for '1' handle, expected 'u8', actual 'u16'', src/main.rs:15:5
```
Though it is a little bit messy. Would you prefer panic instead of assertion with custom formatting?

dmitry-timofeev: Looks good!
> Would you prefer panic instead of assertion with custom formatting?

Not sure I understand what you mean by that ;-) Making a panic with a fully controlled error message?

DarkEld3r: Yes, something like that:
```rust
if info.object_type != actual_object_type {
    panic!(
        "Wrong type id for '{:X}' handle, expected '{}' ({}), actual '{}' ({})",
        handle,
        info.type_name,
        info.object_type,
        any::type_name::<T>(),
        actual_object_type,
    );
}
```

dmitry-timofeev: I see, thanks! As we've already got the code, I'd go with it: it is better.

PR #1219, opened 2019-11-07
Update service runtime to the new revision 4 [ECR-3802]

dmitry-timofeev:
## Overview

Updates the service runtime to the new Exonum revision (exonum/exonum@2bbea66549764397962e61e1eeff46a68147c8dd).

Extracted the connection of service APIs into RuntimeTransport. Also fixed the documentation of Server#start clarifying that start happens asynchronously; and modified the method to return a Future to enable the clients to know when the op is complete.

There are some things to consider regarding these abstractions:
1. Shall we merge the Server into RuntimeTransport? Are there any uses of Server separate from RuntimeTransport?
2. ServiceRuntime does not currently use async operations (hence the RuntimeTransport is blocking). I decided not to downgrade the Server to a blocking interface now since it's already there and it is not known if it will be useful when service _unloading_ is added; but it may be sensible to do if we do know we don't need it.

---

⚠️ Tests won't run without the JNI cache patched.

---

See: https://jira.bf.local/browse/ECR-3802

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/checkstyle-suppressions.xml
+++ b/exonum-java-binding/core/checkstyle-suppressions.xml
@@ -23,7 +23,7 @@
     <suppress files="BlockTest\.java" checks="MethodName"/>
 
     <!-- Suppress indentation as there is a bug with lambdas -->
-    <suppress files="ServiceRuntime.java" checks="Indentation"/>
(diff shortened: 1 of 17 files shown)


dmitry-timofeev:
Merged into 8779153

PR #1222, opened 2019-11-12
Add no key hashing option to ProofMapIndexProxy [ECR-3779]

MakarovS:
## Overview
Add no key hashing option to ProofMapIndexProxy, update tests.

---
See: https://jira.bf.local/browse/ECR-3779

### Definition of Done

- [x] Public API has Javadoc


dmitry-timofeev:
Github shows the modifications weirdly, use `git diff master:exonum-java-binding/core/src/test/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxyIntegrationTest.java HEAD:exonum-java-binding/core/src/test/java/com/exonum/binding/core/storage/indices/BaseProofMapIndexProxyIntegrationTest.java` to see the actual differences between the previous tests and the new base tests.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxy.java
```diff
@@ -44,8 +44,15 @@
  * that a certain key is mapped to a particular value <em>or</em> that there are no mapping for
  * the key in the map.
  *
- * <p>This map is implemented as a Merkle-Patricia tree. It does not permit null keys and values,
- * and requires that keys are 32-byte long.
+ * <p>This map is implemented as a Merkle-Patricia tree. It does not permit null keys and values.
+ *
+ * <p>This map can be instantiated in two ways - either with keys hashing with methods
```
bullet-tooth: with..with...
how about : ...either with keys hashing *using* methods...

exonum-java-binding/core/src/test/java/com/exonum/binding/core/storage/indices/BaseProofMapIndexProxyIntegrationTest.java
```diff
@@ -0,0 +1,806 @@
...
+  void update(ProofMapIndexProxy<HashCode, String> index) {
+    index.put(key1, V1);
+  }
+
+  /**
+   * Creates `numOfEntries` map entries, sorted by key:
+   * [(00…0PK1, V1), (00…0PK2, V2), … (00…0PKi, Vi)].
+   */
+  List<MapEntry<HashCode, String>> createSortedMapEntries() {
+    // Use PROOF_KEYS which are already sorted.
```
dmitry-timofeev: The implementation and documentation are not correct. Nowhere the base class says that getProofKeys are expected to be sorted
     and how. The usages of this method must be reviewed to determine if they need sorting.
     This method/getProofKeys and documentation must be updated accordingly.

exonum-java-binding/core/src/test/java/com/exonum/binding/core/storage/indices/BaseProofMapIndexProxyIntegrationTest.java
```diff
@@ -0,0 +1,806 @@
...
+  private List<MapEntry<HashCode, String>> createMapEntries(Stream<HashCode> proofKeys) {
+    Stream<HashCode> keys = proofKeys.distinct();
+    Stream<String> roundRobinValues = IntStream.range(0, Integer.MAX_VALUE)
+        .mapToObj(i -> values.get(i % values.size()));
+    return Streams.zip(keys, roundRobinValues, MapEntry::valueOf)
+        .collect(Collectors.toList());
+  }
+
+  /**
+   * Creates 257 entries for a ProofMap that, when added to it, will make the underlying
```
dmitry-timofeev: Tests that rely on keys being equal to proof paths make no sense in the base but
non-hashing flavour of the ProofMap where this property holds. In hashing variant
there is no apparent connection between a key bits and the bits in the corresponding proof path. 

The tests that set particular key bits to achieve a certain tree structure to exercise the proof validation must be moved to non-hashing variant.

exonum-java-binding/core/src/test/java/com/exonum/binding/core/storage/indices/BaseProofMapIndexProxyIntegrationTest.java
```diff
@@ -0,0 +1,806 @@
...
+
+      assertThat(map, provesThatPresent(entries));
+    });
+  }
+
+  @Test
+  void getMultiProof_FourEntryMap_DoesNotContain() {
+    runTestWithView(database::createFork, (map) -> {
+      /*
+       This map will have the following structure:
```
dmitry-timofeev: That is incorrect in the base, see below.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxy.java
```diff
@@ -44,8 +44,15 @@
...
+ * {@link #newInstance(String, View, Serializer, Serializer)} and
+ * {@link #newInGroupUnsafe(String, byte[], View, Serializer, Serializer)} or with no key hashing
+ * with methods {@link #newInstanceNoKeyHashing(String, View, Serializer, Serializer)} and
+ * {@link #newInGroupUnsafeNoKeyHashing(String, byte[], View, Serializer, Serializer)}. In case of
+ * no key hashing keys are required to be 32-byte long. Note that the former option is considered
```
dmitry-timofeev: We discussed why it shall be the default option, and the documentation must properly reflect that.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxy.java
```diff
@@ -86,9 +93,34 @@
...
     IndexAddress address = IndexAddress.valueOf(name);
     long viewNativeHandle = view.getViewNativeHandle();
-    LongSupplier nativeMapConstructor = () -> nativeCreate(name, viewNativeHandle);
+    LongSupplier nativeMapConstructor = () -> nativeCreate(name, viewNativeHandle, true);
+
+    return getOrCreate(address, view, keySerializer, valueSerializer, nativeMapConstructor, true);
+  }
+
+  /**
+   * Creates a ProofMapIndexProxy with no key hashing. Requires that keys are 32-byte long.
```
dmitry-timofeev: Would it be possible to add a short warning with a link to the class-level section?

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxy.java
```diff
@@ -86,9 +93,34 @@
...
+   * @param view a database view. Must be valid.
+   *             If a view is read-only, "destructive" operations are not permitted.
+   * @param keySerializer a serializer of keys, must always produce 32-byte long values
+   * @param valueSerializer a serializer of values
+   * @param <K> the type of keys in the map
+   * @param <V> the type of values in the map
+   * @throws IllegalStateException if the view is not valid
+   * @throws IllegalArgumentException if the name is empty
+   * @see StandardSerializers
+   * @see ProofMapIndexProxy
```
dmitry-timofeev: Remove, that's this class already.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxy.java
```diff
@@ -151,6 +210,17 @@
...
   }
 
+  private static <K> ProofMapKeyCheckingSerializerDecorator<K> decorateKeySerializer(
+      Serializer<K> keySerializer, boolean keyHashing) {
+    if (!keyHashing) {
+      ProofMapKeySizeCheckingSerializerDecorator<K> sizeCheckingSerializerDecorator =
+          ProofMapKeySizeCheckingSerializerDecorator.from(keySerializer);
+      return ProofMapKeyCheckingSerializerDecorator.from(sizeCheckingSerializerDecorator);
+    } else {
+      return ProofMapKeyCheckingSerializerDecorator.from(keySerializer);
```
dmitry-timofeev: Why is this needed at all if it can return just keySerializer, with no extra decorator types?

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxy.java
```diff
@@ -184,10 +254,11 @@ public boolean containsKey(K key) {
   /**
    * {@inheritDoc}
    *
-   * @param key a proof map key, must be 32-byte long when serialized
+   * @param key a proof map key
    * @param value a storage value to associate with the key
    * @throws IllegalStateException if this map is not valid
-   * @throws IllegalArgumentException if the size of the key is not 32 bytes
+   * @throws IllegalArgumentException if the size of the key is not 32 bytes (in case of a
+   *     no key hashing proof map)
```
dmitry-timofeev: Here and elsewhere 'a no key hashing proof map' does not sound correct.
 * ... a non-key-hashing proof map
 * ... a proof map that does not hash keys
 * if the size of the key is not 32 bytes *and* this map uses non-hashed keys ...

dmitry-timofeev: @Maria-Nosyk Would you tell us please which works best?

exonum-java-binding/core/src/test/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxyGroupIntegrationTest.java
```diff
@@ -27,6 +27,10 @@
 
 class ProofMapIndexProxyGroupIntegrationTest extends BaseMapIndexGroupTestable<HashCode> {
 
+  private static final HashCode PK1 = HashCode.fromString(K1);
```
dmitry-timofeev: Nit: may just use `String`s (Kx are already strings).

exonum-java-binding/core/src/test/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxyNoKeyHashingIntegrationTest.java
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  @Test
+  void entriesTest() {
+    runTestWithView(database::createFork, (map) -> {
+      List<MapEntry<HashCode, String>> entries = createSortedMapEntries();
+
+      putAll(map, entries);
+
+      Iterator<MapEntry<HashCode, String>> entriesIterator = map.entries();
```
dmitry-timofeev: (keys, values, entries) have incomplete documentation of the order of the elements missing the default variant.

dmitry-timofeev: The conditional documentation (with the fact that the API users always need to know if hashing happens or not to access the map from other services or verify the proofs it produces) makes me wonder if we shall add two separate types. But that's certainly for a separate issue.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/ProofMapKeyCheckingSerializerDecorator.java
```diff
@@ -16,15 +16,12 @@
...
-import static com.exonum.binding.core.storage.indices.StoragePreconditions.checkProofKey;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.exonum.binding.common.serialization.Serializer;
 
 /**
- * A serializer decorator that checks proof map keys for correctness.
- *
- * @see StoragePreconditions#checkProofKey(byte[])
+ * A serializer decorator that checks proof map keys are not null.
```
dmitry-timofeev: Revert the class to its past form.

dmitry-timofeev:
I expect this PR to have non-hashing tests `@Disabled` (since there is no support of them in native yet); and hashing one — enabled and passing (the appropriate native method requires a not-yet used jboolean flag, probably, prefixed with underscore so that clippy does not complain).

Also, the client code in various schemas must be updated to use the appropriate constructor.

exonum-java-binding/core/src/test/java/com/exonum/binding/core/storage/indices/BaseProofMapIndexProxyIntegrationTest.java
```diff
@@ -0,0 +1,806 @@
...
+import java.util.Map;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Test;
+
+abstract class BaseProofMapIndexProxyIntegrationTest
```
dmitry-timofeev: This package uses Testable as a suffix to indicate it is not a concrete test but abstract (I don't remember exactly, but I think surefire might also fail to instantiate a non-instantiable 'test' class, as it detects them based on the naming pattern *Test.java).

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxy.java
MakarovS: To check for not null values?

dmitry-timofeev: There is one for that already.

dmitry-timofeev: com.exonum.binding.common.serialization.CheckingSerializerDecorator

exonum-java-binding/core/src/test/java/com/exonum/binding/core/storage/indices/BaseProofMapIndexProxyIntegrationTestable.java
```diff
@@ -305,21 +225,19 @@ void getProof_FourEntryMap_FirstAndLastByte_Contains() {
...
 
-      for (MapEntry<HashCode, String> e : entries) {
-        assertThat(map, provesThatPresent(e.getKey(), e.getValue()));
-      }
+      assertThat(map, provesThatPresent(entries));
     });
   }
 
   @Test
   void getProof_MultiEntryMapDoesNotContain() {
```
MakarovS: This test creates 32-byte keys, but it still seems to be appropriate for both proof map cases.

MakarovS:
> I expect this PR to have non-hashing tests @Disabled (since there is no support of them in native yet); and hashing one — enabled and passing (the appropriate native method requires a not-yet used jboolean flag, probably, prefixed with underscore so that clippy does not complain).

Ok, will disable those, I thought core changes are already merged and we'd be able to integrate this PR with native code, but I see it's still WIP.

exonum-java-binding/core/src/test/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxyNoKeyHashingIntegrationTest.java
```diff
@@ -0,0 +1,482 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Test;
+
+class ProofMapIndexProxyNoKeyHashingIntegrationTest
+    extends BaseProofMapIndexProxyIntegrationTestable {
+
+  static final List<HashCode> PROOF_KEYS = Stream.of(
```
dmitry-timofeev: TEST_KEYS/SORTED_TEST_KEYS

exonum-java-binding/core/src/test/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxyNoKeyHashingIntegrationTest.java
```diff
@@ -0,0 +1,482 @@
...
+  )
+      .map(BaseProofMapIndexProxyIntegrationTestable::createRawProofKey)
+      .sorted(UnsignedBytes.lexicographicalComparator())
+      .map(HashCode::fromBytes)
+      .collect(Collectors.toList());
+
+  private static final HashCode INVALID_PROOF_KEY = HashCode.fromString("1234");
+
+  @Override
+  List<HashCode> getProofKeys() {
```
dmitry-timofeev: Nit: getKeys or getTestKeys

exonum-java-binding/core/src/test/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxyNoKeyHashingIntegrationTest.java
```diff
@@ -0,0 +1,482 @@
...
+    }
+    return createPrefixed(keyPrefixBits.toByteArray(), PROOF_MAP_KEY_SIZE);
+  }
+
+  /**
+   * Creates `numOfEntries` map entries, sorted by key:
+   * [(00…0PK1, V1), (00…0PK2, V2), … (00…0PKi, Vi)].
+   */
+  List<MapEntry<HashCode, String>> createSortedMapEntries() {
+    // Use PROOF_KEYS which are already sorted.
```
dmitry-timofeev: Use them directly instead of interface method, with appropriate
    naming there won't be a need for comment

exonum-java-binding/core/src/test/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxyNoKeyHashingIntegrationTest.java
dmitry-timofeev: ❕

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxy.java
Maria-Nosyk: The third variant is absolutely perfect!

dmitry-timofeev: Brilliant, thanks!

dmitry-timofeev:
> Also, the client code in various schemas must be updated to use the appropriate constructor.

As most of ProofMaps in services and schemas continue using nonHashing, they shall be migrated to that. (The only hashing is probably DispatcherSchema).

Separately we might consider migrating some services to the default hashing variant, e.g., QaService with its two maps 🤔

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxy.java
MakarovS: Is `non-key-hashing proof map` a bad option though? Third variant is a bit too wordy in some cases.

Maria-Nosyk: Yup, the construction you suggest is, first, hard to understand, second, contradicts general requirements to technical documentation. The second variant suggested by @dmitry-timofeev is also acceptable.

MakarovS:
> As most of ProofMaps in services and schemas continue using nonHashing, they shall be migrated to that. (The only hashing is probably DispatcherSchema).

Ok, will do that in a separate PR.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxy.java
```diff
@@ -429,6 +443,12 @@ public void remove(K key) {
 
   private native void nativeValuesIterFree(long iterNativeHandle);
 
+  /**
+   * {@inheritDoc}
+   *
+   * The entries are ordered by keys in lexicographical order if this map is a non-key-hashing
```
dmitry-timofeev: `... is a <a href="...">non-key-hashing proof map</a>.`

dmitry-timofeev:
> Ok, will do that in a separate PR.

I don't think it is reasonable: this one changes the specification of `newInstance` (and it becomes *broken* according to its spec until the native is done), replacing that with `newInstanceNonHashing` (which *works* in accordance with the spec). Why the clients, of which there are 6 total, shall remain on the broken version and **complicate** the native implementation, which, when done, will render the clients **actually** broken?

It is 15 seconds search-replace (6 usages in prod code) + 2–5 minutes documenting the non-hashing aspect, where applicable.

MakarovS:
> It is 15 seconds search-replace (6 usages in prod code) + 2–5 minutes documenting the non-hashing aspect, where applicable.

Ok, updated those. Not sure what documentation is applicable for those, other than `// Use non-hashing proof map, as keys of this map are already 32-byte long hashes`, which seems redundant, given Javadoc of the corresponding method.

dmitry-timofeev:
The variant in use must be documented in schemas as they are supposed to provide adequate specification of the service persistent data. Whether or not a proof map hashes its keys affects the clients verifying proofs.

> as keys of this map are already 32-byte long hashes

What the API documentation must say is whether or not a hashing variant is used (it mustn't explain why). As 32-byte length is *not* enough to use a non-key-hashing map, as its documentation says, the justification might be either in the API docs or in implementation docs.

dmitry-timofeev:
> Whether or not a proof map hashes its keys affects the clients verifying proofs. 

If the proofs included the type of map from which they are created, the specification of key hashing won't be needed. I am not sure if it makes sense to include such flag in the proof.

dmitry-timofeev:
@MakarovS please resolve the build.

MakarovS:
> @MakarovS please resolve the build.

Fixed.

dmitry-timofeev:
If the ordering is the same, then it might make sense to move the tests of iterators back to the base.

dmitry-timofeev:
https://jira.bf.local/browse/ECR-3871

PR #1223, opened 2019-11-12
Fix ITs and travis job (minor changes)

ipenkin:
## Overview
ITs and travis job are fixed before ECR-3591

---
See: https://jira.bf.local/browse/ECR-3591

### Definition of Done


exonum-java-binding/core/rust/integration_tests/tests/node_context.rs
```diff
@@ -31,12 +31,11 @@ use integration_tests::vm::create_vm_for_tests_with_fake_classes;
...
         api::ApiContext,
-        blockchain::{Blockchain, Service, Transaction},
-        crypto::{gen_keypair, Hash, PublicKey, SecretKey},
-        messages::{RawTransaction, ServiceTransaction},
+        crypto::{gen_keypair, PublicKey, SecretKey},
         node::{ApiSender, ExternalMessage},
+        runtime::{AnyTx, CallInfo},
     },
-    exonum_merkledb::{Snapshot, TemporaryDB},
+    exonum_merkledb::TemporaryDB,
```
skletsun: AFAIR we once agreed that we always re-import `merkledb`  from `exonum`.

dmitry-timofeev: What exactly do you suggest? Is it a singular occurrence, or anything has to be done project-wide? I don't see the patch adding any new imports.

dmitry-timofeev:
This patch needs to be updated as #1221 is in.

exonum-java-binding/core/rust/integration_tests/tests/node_context.rs
skletsun: My suggestion was to put this line from `java_bindings::{` to `java_bindings::{exonum::{` but it seems not important in case of IT's. Yes, there is no new patch, it's like we do for older non-nested imports for example. But once again - I believe that these imports are equivalent for IT-s so please just forget what I wrote 🙌

dmitry-timofeev: ok, thanks, on a side note — why do we use the nested imports, which are not search/replace-friendly, because there is a million ways in which they can be written?

skletsun: I feel like just because it's the newer syntax that avoids some repetition. Personally I didn't run into problems with searching/replacing so far 🙂.

dmitry-timofeev: Could you replace all instances of exonum java_bindings::exonum::foo::Bar to java_bindings::foo::Bar across the code base with a single pattern (if it may be nested like java_bindings::{exonum::foo::Bar, java_bindings::exonum::{foo::Bar, ... + with any other imports and whitespace in between)?

PR #1224, opened 2019-11-13
Update QA service for DS [ECR-3586]:

dmitry-timofeev:
## Overview

1. Added an ability to set key pair in TransactionMessage.Builder, so that the builder with a _set_ key pair can be passed around. #signedWith(keys) is provided to just set the keys (then can be followed by any other setters and completed with #build); #sign — as a shortcut to builder.signedWith(keys).build().
2. Added a Testkit#port method returning the TCP port on which the server providing service transport listens
3. Added a system-time provider for cases when the tests don’t need to control the current time.
4. QA service:
    1. Removed conversion of transactions into JSON, as it is no longer supported by the core.
    2. Removed conversion of transactions into RawTransactions
    3. Removed usages of RequiresNativeLibrary as it is no longer used in filtering
    4. Removed operations submitting a self-signed transaction except the two enabling Node testing. The rest of transactions shall be submitted through the usual channel.
    5. Moved transaction decoding tests to the transaction converter.

---
See: https://jira.bf.local/browse/ECR-3586

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/qa-service/src/test/java/com/exonum/binding/qaservice/Integration.java
```diff
@@ -17,19 +17,20 @@
...
-public @interface PromoteToCore {
-  /**
-   * Why the element might be promoted to the core library.
-   */
-  String value() default "";
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Inherited
+@Tag("integration")
+public @interface Integration {
```
MakarovS: Why did you go with this annotation instead of `*IntegrationTest.java`?

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/SystemTimeProviderTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import org.junit.jupiter.api.RepeatedTest;
+
+class SystemTimeProviderTest {
+
+  @RepeatedTest(2)
```
MakarovS: Why `Repeated`?

exonum-java-binding/qa-service/src/test/java/com/exonum/binding/qaservice/Integration.java
dmitry-timofeev: The filtering by file name appeared broken with @Nested tests (<excludes>**/*IntegrationTest.java<excludes> did not exclude the @Nested tests appearing inside *IntegrationTest). Regex by class didn't work either :(

I'll try to put together a reproducing project and submit an issue to surefire project.

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/SystemTimeProviderTest.java
dmitry-timofeev: The intention was to test that any invocation reports the up-to date time; I can clarify the intention or remove it.

dmitry-timofeev: As it uses the factory method to get the provider, we don't actually check that the instance reports the up-to-date time.

I think it is not worth to test that subsequent invocations of getTime produce the current time, it must too simple to break.

exonum-java-binding/qa-service/src/test/java/com/exonum/binding/qaservice/Integration.java
dmitry-timofeev: Turns out, it reproduces only if the `@Nested ` class is names *Tests.

dmitry-timofeev: https://issues.apache.org/jira/browse/SUREFIRE-1720 

We could use a pattern instead, for higher consistency with other modules.

dmitry-timofeev: OK, I'll think will do that later: https://jira.bf.local/browse/ECR-3852

PR #1227, opened 2019-11-16
Fix parallel TestKit tests [ECR-3797]

MakarovS:
## Overview
Fix parallel TestKit tests.

---
See: https://jira.bf.local/browse/ECR-3797

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitExtensionTest.java
```diff
@@ -27,18 +36,14 @@
...
+import org.junit.jupiter.api.io.TempDir;
 import org.junit.platform.engine.TestExecutionResult;
 import org.junit.platform.testkit.engine.EngineExecutionResults;
 import org.junit.platform.testkit.engine.EngineTestKit;
 import org.junit.platform.testkit.engine.Event;
 import org.junit.platform.testkit.engine.Events;
 
-class TestKitExtensionTest extends TestKitTestWithArtifactsCreated {
-
-  private static final TestKit.Builder defaultBuilder = TestKit.builder()
```
dmitry-timofeev: As all the test classes below are invoked explicitly from within this test class, could we create the directory and an artifact here and once, so that that code isn't duplicated?

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/TestKitTestUtils.java
```diff
@@ -27,10 +27,8 @@
 import com.exonum.core.messages.Runtime.InstanceSpec;
 import java.io.IOException;
 import java.nio.file.Path;
-import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.io.TempDir;
 
-class TestKitTestWithArtifactsCreated {
```
dmitry-timofeev: Might also make non-extendable/instantiable as it is no longer suppossed to be.

PR #1228, opened 2019-11-19
Update native to Exonum revision 6d7cc3a9c6 [ECR-3846]

vitvakatu:
## Overview

Updated Core to revision 6d7cc3a9c6

---
See: https://jira.bf.local/browse/ECR-3846

### Definition of Done

- [ ] Public API has Javadoc


exonum-java-binding/core/rust/src/storage/entry.rs
```diff
@@ -44,10 +44,14 @@ pub extern "system" fn Java_com_exonum_binding_core_storage_indices_EntryIndexPr
         let name = utils::convert_to_string(&env, name)?;
         Ok(handle::to_handle(
             match handle::cast_handle::<View>(view_handle).get() {
-                ViewRef::Snapshot(snapshot) => {
-                    IndexType::SnapshotIndex(Index::new(name, &*snapshot))
-                }
-                ViewRef::Fork(fork) => IndexType::ForkIndex(Index::new(name, fork)),
+                ViewRef::Snapshot(snapshot) => IndexType::SnapshotIndex(
+                    Index::from_access(snapshot, name.into())
+                        .expect("Index type does not match specified one"),
```
dmitry-timofeev: Can we use #new that doesn't return a Result but Self? 

Is it the only possible Error ("Index type does not match specified one" — will a user see the actual in the description?)? Is this error documented in the core?

vitvakatu: #new is private, so no. There is a bunch of methods like `fn get_proof_list<I, V>(self, addr: I) -> ProofListIndex<Self::Base, V>`, but they all use the same code as here (with `unwrap` instead of `expect`).

Error message contains also index name, expected and actual type. The only reason I added `expect` - to pass existing Java tests in BaseIndexProxyIntegrationTests.

> Is it the only possible Error?

There is also second variant called `Custom`, but I can't find if it's used anywhere in Core.

vitvakatu: @slowli Are any other errors possible in this method?

dmitry-timofeev: But will that error message be included in the exception message that we throw in case of panic here, or disappear? If unwrap provides that, it'd be better. 

Or, better, can we just return Error and `unwrap_exc_or` will throw exception with Error#to_string?

While we are here, please document in the `unwrap_exc_or` documentation what nested Results mean, because each time this method appears one has to dig in to understand that.

exonum-java-binding/qa-service/src/test/java/com/exonum/binding/qaservice/transactions/ThrowingTxTest.java
```diff
@@ -77,7 +77,7 @@ but also to access the description and match it against a condition (e.g., conta
     ExecutionStatus txResult = blockchain.getTxResult(throwingTx.hash()).get();
     assertTrue(txResult.hasError());
     ExecutionError error = txResult.getError();
-    assertThat(error.getKind()).isEqualTo(ErrorKind.RUNTIME);
+    assertThat(error.getKind()).isEqualTo(ErrorKind.SERVICE);
```
dmitry-timofeev: Why SERVICE if SERVICE means service-defined errors? What will be the error code? Will it clash with service-defined error-codes?

dmitry-timofeev: I do not necessarily object to representing 'unexpected' errors with SERVICE, but it requires a predefined and documented (and forbidden to be used explicitly ) error code to represent those.

dmitry-timofeev: I am in favour of redefining SERVICE as any errors *occurring* in service transactions instead of service-*defined*.

exonum-java-binding/core/rust/src/storage/entry.rs
vitvakatu: Current exception message is: `Index type does not match specified one: AccessError { addr: IndexAddress { name: "test_index", bytes: None }, kind: WrongIndexType { expected: List, actual: ValueSet } }`

Returning error is... complicated, because `unwrap_exc_or` is not working with any kinds of errors, but rather with `jni` errors.

exonum-java-binding/qa-service/src/test/java/com/exonum/binding/qaservice/transactions/ThrowingTxTest.java
vitvakatu: Was my mistake, runtime execution error got invalid error kind.

exonum-java-binding/core/rust/src/storage/entry.rs
dmitry-timofeev: Do I understand correctly, that if we replace expect(msg) with unwrap, it will be
"AccessError { addr: IndexAddress { name: "test_index", bytes: None }, kind: WrongIndexType { expected: List, actual: ValueSet } }"?

dmitry-timofeev: If that's correct, IMO it's good enough in this rare case.

vitvakatu: Yep, it initially was. I will change it.

vitvakatu:
> the proto sources must also be updated

No changes to proto sources were made

dmitry-timofeev:
> No changes to proto sources were made

Shall we keep some file in the exonum-proto-sources (REVISION.txt with a source revision) so that we can _always_ update it, even when no changes are made? Currently (1) you had to manually verify an absence of changes; (2) anyone who sees a dependency on commit that corresponds to an older revision will wonder if this might be a compatibility problem (protos come from different revision than exonum code).

exonum-java-binding/core/rust/src/utils/errors.rs
```diff
@@ -138,6 +138,12 @@ type ExceptionResult<T> = thread::Result<result::Result<T, JniError>>;
...
 /// at the Java side. So this function should be used only for the `panic::catch_unwind` result.
+///
+/// The function accepts `Result<JniResult, Error>`, where outer `Result<..., Error>` corresponds
+/// to a return type of `panic::catch_unwind` and is `Err` in case of catched panic; inner
+/// `JniResult` represents the errors during Rust-Java interoperability. Therefore, the function
+/// can't be used for handling __any__ user-defined error type, but supports only a set of
+/// JNI-related errors and also handles unexpected panics.
 pub fn unwrap_exc_or<T>(env: &JNIEnv, res: ExceptionResult<T>, error_val: T) -> T {
     match res {
         Ok(val) => {
```
dmitry-timofeev: Shan't we have the switch annotated for clarity, e.g.,

// A panic
// No panic
// A JNI error

vitvakatu:
> Shall we keep some file in the exonum-proto-sources (REVISION.txt with a source revision) so that we can always update it, even when no changes are made? Currently (1) you had to manually verify an absence of changes; (2) anyone who sees a dependency on commit that corresponds to an older revision will wonder if this might be a compatibility problem (protos come from different revision than exonum code).

Good idea, I'll send a PR for exonum-proto-sources

PR #1234, opened 2019-11-22
Support configure interface [ECR-3590]

dmitry-timofeev:
## Overview

Add Configurable interface corresponding to `exonum.Configure`
and its support to ServiceRuntime.

---
See: https://jira.bf.local/browse/ECR-3590

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceWrapper.java
```diff
@@ -41,6 +46,26 @@
  */
 final class ServiceWrapper {
 
+  /**
+   * Default interface comprised of transactions defined in the service implementation
+   * (intrinsic to this service).
+   * todo: @slowli What's the proper term for this interface (intrinsic, implicit, default)?
```
dmitry-timofeev: @slowli Is there one?

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntimeAdapter.java
```diff
@@ -163,25 +163,30 @@ private static ServiceInstanceSpec parseInstanceSpec(byte[] instanceSpec) {
    * Executes the service transaction.
    *
    * @param serviceId the service numeric identifier
+   * @param interfaceName the name of the interface in which the transaction is defined
    * @param txId the transaction type identifier within the service
    * @param arguments the transaction arguments
    * @param forkNativeHandle a handle to a native fork object
+   * @param callerServiceId the id of the service which invoked the transaction (in case of
+   *      inner transactions; or 0 when the caller is an external message
```
vitvakatu: nit: missing closing brace

vitvakatu: Also, I can't find any mention that `0` can't be a service ID. Doesn't `-1` suit better in this case?

update: `0` is a supervisor service's ID, but supervisor service isn't strictly essential for blockchain operations (or it shouldn't be, at least).

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceWrapper.java
```diff
@@ -80,7 +105,24 @@ void initialize(Fork view, Configuration configuration) {
...
+      case DEFAULT_INTERFACE_NAME: {
+        executeIntrinsicTransaction(txId, arguments, context);
+        break;
+      }
+      case CONFIGURE_INTERFACE_NAME: {
+        executeConfigurableTransaction(txId, arguments, callerServiceId, context);
+        break;
+      }
+      default: throw new IllegalArgumentException(
+          format("Unknown interface (name=%s, txId=%d)", interfaceName, txId));
```
MakarovS: Why static import for `String.format`? That's inconsistent with the rest of our codebase

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Configurable.java
```diff
@@ -0,0 +1,64 @@
...
+import com.exonum.binding.core.storage.database.Fork;
+
+/**
+ * A configurable Exonum service. Allows services to update their configuration through
+ * the supervisor service during their operation.
+ *
+ * <p>The configuration update process includes the following steps: a proposal
+ * of a new configuration; verification of its correctness; approval of the proposal;
+ * and application of the new configuration. The protocol of the proposal and approval steps
+ * is determined by the installed supervisor service. The verification and application
```
MakarovS: `are determined`?

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/ServiceWrapperTest.java
```diff
@@ -46,12 +58,15 @@
 
   private static final ServiceArtifactId TEST_ARTIFACT_ID =
       ServiceArtifactId.newJavaId("com.acme:foo:1.2.3");
+  private static final String TEST_SERVICE_NAME = "test-service";
+  private static final int TEST_SERVICE_ID = 1;
+  private static final String DEFAULT_INTERFACE = "";
```
MakarovS: Why not `ServiceWrapper.DEFAULT_INTERFACE_NAME`?

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntimeAdapter.java
dmitry-timofeev: 0 here doesn't matter as it isn't supposed to be used in the case, same as hash code and public key in the other case (which are all zeroes). Whether a transaction is external is determined by the interface name currently. It could be set to something different to have something different from the supervisor id, but I hope this signature won't last long.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Configurable.java
dmitry-timofeev: The protocol ... is determined

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceWrapper.java
dmitry-timofeev: To have it more consice and fit in a single expression. I wouldn't consider the consistency of static imports of a particular method as something to be given higher priority over readability and local consistency.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Configurable.java
MakarovS: You are right, my mistake

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceWrapper.java
dmitry-timofeev: Pink @slowli

slowli: Sorry, didn't notice the ping. The core [uses "default"](https://github.com/exonum/exonum/blob/38f5d3d73fa372b6803a54fb16d640eb75e59f18/exonum/src/runtime/mod.rs#L361), however, I'd not consider this a 100% settled term. Do you think another term from the list is better? IMO, there's a tie between "default" and "intrinsic".

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntimeAdapter.java
slowli: > Whether a transaction is external is determined by the interface name currently.

The keyword is "currently." I'm not 100% sure that we'll keep the inability of external clients to call non-default service interface in 1.0, but this restriction will be lifted _eventually_. In this case the current implementation seemingly becomes _dangerous_: any transaction from the external client will be considered to be authorized by the supervisor! (I'm sure it'll be fixed by that time.)

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceWrapper.java
dmitry-timofeev: TIL default is not an adjective. 

I agree that both make sense: [intrinsic](https://www.oxfordlearnersdictionaries.com/definition/english/intrinsic?q=intrinsic) says that this interface belongs to, or is a part of a service artifact (with no extra definition elsewhere); [default](https://www.lexico.com/en/definition/default) says that that interface is used when no interface is specified.

dmitry-timofeev: ["implicit"](https://www.oxfordlearnersdictionaries.com/definition/english/implicit?q=implicit) will probably highlight that there is no explicit definition, but I don't think that is what must be communicated.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntimeAdapter.java
dmitry-timofeev: The thing is that what is currently available is very limited and I don't expect this implementation to last (particularly, this monstrous parameter list). It just doesn't support (and explicitly forbids) anything but 'exonum.Configure'. Supporting anything else will require changes.

PR #1245, opened 2019-11-27
Adapt ProofMapIndexProxy to the configurable key hashing [ECR-3765].

skletsun:
## Overview

ProofMapIndexProxy now supports two key hashing modes. Important changes:
- The proofs are returned serialized to protobuf.
- The format of incoming keys for mutiproof has been changed - now the Java array of byte arrays is expected as input.
---
See: https://jira.bf.local/browse/ECR-3765

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/core/rust/src/storage/proof_map_index.rs
```diff
@@ -12,49 +12,74 @@
...
     PairIter,
 };
 use utils;
 use JniResult;
 
-type Key = [u8; PROOF_MAP_KEY_SIZE];
-type Index<T> = ProofMapIndex<T, Key, Value>;
+use std::{panic, ptr};
+
+type RawKey = [u8; 32];
```
vitvakatu:
```suggestion
type RawKey = [u8; PROOF_MAP_KEY_SIZE];
```

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
```diff
@@ -12,49 +12,74 @@
...
+
+enum Index<T>
```
vitvakatu: I'd add a comment explaining differences between RawProofMapIndex and ProofMapIndex

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
```diff
@@ -63,18 +88,34 @@ pub extern "system" fn Java_com_exonum_binding_core_storage_indices_ProofMapInde
     _: JClass,
     name: JString,
     view_handle: Handle,
-    // TODO: to be used in ECR-3765
-    _key_hashing: jboolean,
+    key_hashing: jboolean,
 ) -> Handle {
     let res = panic::catch_unwind(|| {
         let name = utils::convert_to_string(&env, name)?;
+        let key_is_hashed = key_hashing != JNI_FALSE;
```
vitvakatu: Will `as bool` work?

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
```diff
@@ -63,18 +88,34 @@ pub extern "system" fn Java_com_exonum_binding_core_storage_indices_ProofMapInde
...
         Ok(handle::to_handle(
             match handle::cast_handle::<View>(view_handle).get() {
                 ViewRef::Snapshot(snapshot) => {
-                    IndexType::SnapshotIndex(Index::from_access(snapshot, name.into()).unwrap())
+                    if key_is_hashed {
+                        IndexType::SnapshotIndex(Index::Hashed(
```
vitvakatu: Looks like IndexType::SnapshotIndex can be moved one level above, to remove some code duplication

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
```diff
@@ -12,49 +12,74 @@
...
+    Raw(ProofMapIndexKeys<'a, RawKey>),
+    Hashed(ProofMapIndexKeys<'a, Key>),
+}
+
+// For easy conversion to RawKey.
+trait ToRawKey {
+    fn to_raw(&self) -> RawKey;
+}
+
+impl ToRawKey for Key {
```
vitvakatu: Really confusing implementation, because `Key == Vec<u8>` and we use it like
```rust
let key: Vec<u8> = env.convert_byte_array(key);
let raw_key = key.to_raw();
```

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
```diff
@@ -12,49 +12,74 @@
...
+
+enum Index<T>
+where
+    T: RawAccess,
+{
+    Raw(RawProofMapIndex<T, RawKey, Value>),
+    Hashed(ProofMapIndex<T, Key, Value>),
+}
 
 const JAVA_ENTRY_FQN: &str = "com/exonum/binding/core/storage/indices/MapEntryInternal";
```
vitvakatu: Must be MAP_ENTRY_INTERNAL_FQN actually.

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
```diff
@@ -406,13 +393,34 @@ pub extern "system" fn Java_com_exonum_binding_core_storage_indices_ProofMapInde
...
+            IndexType::SnapshotIndex(ref index) => match index {
+                Index::Raw(map) => handle::to_handle(Iter::Raw(PairIter::new(
+                    &env,
+                    map.iter_from(&key.to_raw()),
+                    JAVA_ENTRY_FQN,
+                )?)),
+                Index::Hashed(map) => handle::to_handle(Iter::Hashed(PairIter::new(
+                    &env,
+                    map.iter_from(&key),
+                    JAVA_ENTRY_FQN,
```
vitvakatu: Looks like last argument of PairIter::new is always the same now, so it can be removed

vitvakatu:
A lot of code can be deduplicated by using additional variables and moving some functions to the other indentation level. Example:
```diff
Index: exonum-java-binding/core/rust/src/storage/proof_map_index.rs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- exonum-java-binding/core/rust/src/storage/proof_map_index.rs	(date 1574842840000)
+++ exonum-java-binding/core/rust/src/storage/proof_map_index.rs	(date 1574849178323)
@@ -348,16 +348,17 @@
     map_handle: Handle,
 ) -> Handle {
     let res = panic::catch_unwind(|| {
-        Ok(match *handle::cast_handle::<IndexType>(map_handle) {
-            IndexType::SnapshotIndex(ref index) => match index {
-                Index::Raw(map) => handle::to_handle(KeysIter::Raw(map.keys())),
-                Index::Hashed(map) => handle::to_handle(KeysIter::Hashed(map.keys())),
+        let keys_iter = match handle::cast_handle::<IndexType>(map_handle) {
+            IndexType::SnapshotIndex(index) => match index {
+                Index::Raw(map) => KeysIter::Raw(map.keys()),
+                Index::Hashed(map) => KeysIter::Hashed(map.keys()),
             },
-            IndexType::ForkIndex(ref index) => match index {
-                Index::Raw(map) => handle::to_handle(KeysIter::Raw(map.keys())),
-                Index::Hashed(map) => handle::to_handle(KeysIter::Hashed(map.keys())),
+            IndexType::ForkIndex(index) => match index {
+                Index::Raw(map) => KeysIter::Raw(map.keys()),
+                Index::Hashed(map) => KeysIter::Hashed(map.keys()),
             },
-        })
+        };
+        Ok(handle::to_handle(keys_iter))
     });
     utils::unwrap_exc_or_default(&env, res)
 }
```

vitvakatu:
Another example:
```diff
Index: exonum-java-binding/core/rust/src/storage/proof_map_index.rs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- exonum-java-binding/core/rust/src/storage/proof_map_index.rs	(date 1574842840000)
+++ exonum-java-binding/core/rust/src/storage/proof_map_index.rs	(date 1574849690278)
@@ -96,15 +96,14 @@
         Ok(handle::to_handle(
             match handle::cast_handle::<View>(view_handle).get() {
                 ViewRef::Snapshot(snapshot) => {
-                    if key_is_hashed {
-                        IndexType::SnapshotIndex(Index::Hashed(
-                            ProofMapIndex::from_access(snapshot, name.into()).unwrap(),
-                        ))
+                    let index = if key_is_hashed {
+                        Index::Hashed(ProofMapIndex::from_access(snapshot, name.into()).unwrap())
                     } else {
-                        IndexType::SnapshotIndex(Index::Raw(
+                        Index::Raw(
                             RawProofMapIndex::from_access(snapshot, name.into()).unwrap(),
-                        ))
-                    }
+                        )
+                    };
+                    IndexType::SnapshotIndex(index)
                 }
                 ViewRef::Fork(fork) => {
                     if key_is_hashed {
```

vitvakatu:
A bit of additional code to convert types (and an ugly hack because of stupid compiler):
```diff
Index: exonum-java-binding/core/rust/src/storage/proof_map_index.rs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- exonum-java-binding/core/rust/src/storage/proof_map_index.rs	(date 1574842840000)
+++ exonum-java-binding/core/rust/src/storage/proof_map_index.rs	(date 1574850462416)
@@ -15,7 +15,7 @@
 use exonum_merkledb::{
     access::{FromAccess, RawAccess},
     proof_map_index::{
-        ProofMapIndexIter, ProofMapIndexKeys, ProofMapIndexValues, PROOF_MAP_KEY_SIZE,
+        ProofMapIndexIter, ProofMapIndexKeys, ProofMapIndexValues, PROOF_MAP_KEY_SIZE, Hashed
     },
     Fork, IndexAddress, ObjectHash, ProofMapIndex, RawProofMapIndex, Snapshot,
 };
@@ -68,6 +68,18 @@
     fn to_raw(&self) -> RawKey;
 }
 
+impl<T: RawAccess> From<ProofMapIndex<T, Key, Value>> for Index<T> {
+    fn from(map: ProofMapIndex<T, Key, Value>) -> Self {
+        Index::Hashed(map)
+    }
+}
+
+impl<T: RawAccess> From<RawProofMapIndex<T, RawKey, Value>> for Index<T> {
+    fn from(map: RawProofMapIndex<T, RawKey, Value>) -> Self {
+        Index::Raw(map)
+    }
+}
+
 impl ToRawKey for Key {
     fn to_raw(&self) -> RawKey {
         assert_eq!(
@@ -96,26 +108,20 @@
         Ok(handle::to_handle(
             match handle::cast_handle::<View>(view_handle).get() {
                 ViewRef::Snapshot(snapshot) => {
-                    if key_is_hashed {
-                        IndexType::SnapshotIndex(Index::Hashed(
-                            ProofMapIndex::from_access(snapshot, name.into()).unwrap(),
-                        ))
+                    let index = if key_is_hashed {
+                        ProofMapIndex::<_, _, _, Hashed>::from_access(snapshot, name.into()).unwrap().into()
                     } else {
-                        IndexType::SnapshotIndex(Index::Raw(
-                            RawProofMapIndex::from_access(snapshot, name.into()).unwrap(),
-                        ))
-                    }
+                        RawProofMapIndex::from_access(snapshot, name.into()).unwrap().into()
+                    };
+                    IndexType::SnapshotIndex(index)
                 }
                 ViewRef::Fork(fork) => {
-                    if key_is_hashed {
-                        IndexType::ForkIndex(Index::Hashed(
-                            ProofMapIndex::from_access(fork, name.into()).unwrap(),
-                        ))
+                    let index = if key_is_hashed {
+                        ProofMapIndex::<_, _, _, Hashed>::from_access(fork, name.into()).unwrap().into()
                     } else {
-                        IndexType::ForkIndex(Index::Raw(
-                            RawProofMapIndex::from_access(fork, name.into()).unwrap(),
-                        ))
-                    }
+                        RawProofMapIndex::from_access(fork, name.into()).unwrap().into()
+                    };
+                    IndexType::ForkIndex(index)
                 }
             },
         ))
```

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
skletsun: Within a single method, Java passes us a byte array that can be one of two native types. We always have to convert it to the `Vec<u8>`, but in case of raw PMI we also `convert` it to its key format. This trait is just a substitution of explicit `convert_to_raw(key)`.

vitvakatu: Will implementing `From<Vec<u8>> for RawKey` work?

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
skletsun: `PairIter` is used not only for `ProofMapIndexProxy`.

vitvakatu: I see, the usage in value_set_index has a different argument. All other are the same though, so we can still think about helper function

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
skletsun: Perhaps, but usually it is recommended comparing to `JNI_TRUE`. I believe that `jboolean != JNI_FALSE` is even safer 🙃.

vitvakatu: Why is it safer? I'd better prefer `should_hash_key = should_hash_key == JNI_TRUE`, just for better readability. Parsing boolean expressions isn't my best side.

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
skletsun: `From` (1) consumes original value (probably, not a big deal) and (2) must never fail conversion (but OTOH we're sure conversion will never happen for other than 256 bit keys).

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
skletsun: Ok, switched to comparing against `JNI_TRUE`.

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
```diff
@@ -387,12 +376,16 @@ pub extern "system" fn Java_com_exonum_binding_core_storage_indices_ProofMapInde
...
 ) -> Handle {
     let res = panic::catch_unwind(|| {
-        Ok(handle::to_handle(
-            match *handle::cast_handle::<IndexType>(map_handle) {
-                IndexType::SnapshotIndex(ref map) => map.values(),
-                IndexType::ForkIndex(ref map) => map.values(),
+        Ok(match *handle::cast_handle::<IndexType>(map_handle) {
+            IndexType::SnapshotIndex(ref index) => match index {
+                Index::Raw(map) => handle::to_handle(map.values()),
+                Index::Hashed(map) => handle::to_handle(map.values()),
```
dmitry-timofeev: Can't handle::to_handle remain outside?

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
```diff
@@ -446,13 +449,17 @@ pub extern "system" fn Java_com_exonum_binding_core_storage_indices_ProofMapInde
     key: jbyteArray,
```
dmitry-timofeev: xFrom have never been supported in Java. We could add support for them, but shall we (in any foreseeable future) — @slowli ? If I understand correctly, they require the user to be aware of how keys are serialized and how that results in a particular ordering.

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
```diff
@@ -611,26 +638,39 @@ pub extern "system" fn Java_com_exonum_binding_core_storage_indices_ProofMapInde
     handle::drop_handle::<ProofMapIndexValues<Value>>(&env, iter_handle);
 }
 
-fn convert_to_key(env: &JNIEnv, array: jbyteArray) -> JniResult<Key> {
-    let bytes = env.convert_byte_array(array)?;
-    assert_eq!(PROOF_MAP_KEY_SIZE, bytes.len());
+// Converts array of Java bytes arrays to the vector of keys.
+fn convert_to_keys(env: &JNIEnv, array: jobjectArray) -> JniResult<Vec<Key>> {
+    let num_elements = env.get_array_length(array)?;
+    let mut keys = vec![];
```
dmitry-timofeev:
```suggestion
    let mut keys = Vec::with_capacity(num_elements as usize);
```

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
```diff
@@ -611,26 +638,39 @@ pub extern "system" fn Java_com_exonum_binding_core_storage_indices_ProofMapInde
...
+    for i in 0..num_elements {
+        let byte_array: jbyteArray = env.get_object_array_element(array, i)?.into_inner();
```
dmitry-timofeev: [Reference management in loops](https://docs.rs/jni/0.14.0/jni/struct.JavaVM.html#local-reference-management)

dmitry-timofeev: Is that correct that AutoLocals don't work well with any type wrappers (= they erase the type destroying any extra type-safety that these wrappers originally provide)?

dmitry-timofeev: Here no wrappers are used at all, but sys types.

dmitry-timofeev:
1. Tests for this type must be enabled (non-hashing maps)
2. Proof tests must be disabled (as this PR breaks them).

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
dmitry-timofeev: May also patch in testkit.

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
skletsun: Yeap, thanks, missed that.

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
skletsun: > May also patch in testkit.

Added as a separate changeset.

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
skletsun: Actually yes, if one works with that wrapper only via `auto_local.as_obj()` (which one is supposed to do since the moment the original object was "wrapped" with `AutoLocal`) then there is no way to see what is exactly behind the returned `JObject`.

exonum-java-binding/core/rust/src/storage/proof_map_index.rs
dmitry-timofeev: https://jira.bf.local/browse/ECR-3942

PR #1239, opened 2019-11-26
Update scripts for launching services [ECR-3841]

vitvakatu:
## Overview

Based on #1236 

- Removed old scripts in exonum-java
- Added a simple exonum-java/start-node.sh script for starting a single node
- Added exonum-launcher configuration for cryptocurrency-demo and qa service

---
See: https://jira.bf.local/browse/ECR-4841

### Definition of Done

- [ ] There are no TODOs left in the code
- [ ] Change is covered by automated [tests](https://github.com/exonum/exonum-java-binding/blob/master/CONTRIBUTING.md#tests)
- [ ] The [coding guidelines](https://github.com/exonum/exonum-java-binding/blob/master/CONTRIBUTING.md#the-code-style) are followed
- [ ] Public API has Javadoc
- [ ] Method preconditions are checked and documented in the Javadoc of the method
- [ ] Changelog is updated if needed (in case of notable or breaking changes)
- [ ] The continuous integration build passes

PR #1249, opened 2019-11-29
Implement supervisor_mode parameter support [ECR-3874]

vitvakatu:
## Overview

Add new parameter `supervisor_mode` for `generate-template` step of node configuration. 
The parameter gets passed through every following configuration step and then is used on `run` step to enable the specified supervisor mode on started node.

`run-dev` uses `simple` mode.

User documentation will be updated in a separate task (TBD: task number)
Done simple refactoring of `cmd` module (it is moved from `runtime/cmd` to own `cmd ` module). Additional refactoring of the module might be done in a separate task (TBD: task number) 
Scripts for launching nodes will be updated after #1239 gets merged

Q: Why so much code? Can't it be done simpler?
A: adding a custom parameter to the very configuration step is pretty simple in terms of code complexity, but involves a lot of code boilerplate because of need in EJB-specific structures and functions. However, adding additional parameters will be much simpler because the most of the boilerplate is already there.

---
See: https://jira.bf.local/browse/ECR-3874

### Definition of Done

- [ ] Public API has Javadoc


dmitry-timofeev:
> Q: Why so much code? Can't it be done simpler?
A: adding a custom parameter to the very configuration step is pretty simple in terms of code complexity, but involves a lot of code boilerplate because of need in EJB-specific structures and functions. However, adding additional parameters will be much simpler because the most of the boilerplate is already there.

Do I understand correctly that what is conceptually done here is a new parameter taking two possible values and the code conditionally creating one or another service based on that parameter? Why would _that_ make a seven hundred line diff? Doesn't it say of a deficiency in the exonum configuration implementation if such a simple thing requires so much code?

vitvakatu:
> Why would that make a seven hundred line diff?

The diff is actually about +390 lines, about 150 of them are tests for new functionality. Tests are quite large because they require a lot of preparations with a lot of (boilerplate) parameters. This can be improved by deriving `Default` trait for Core structures in the future.
The code isn't only taking a single parameter from user, but transfers this parameter through every configuration step, validates its equality in every public config and does it in a way that allows further extend of ejb-specific parameters list.

However, it's not a lot of code, and we should find the way to reduce its amount. Unfortunately I couldn't come up with an API that both flexible and statically typed yet.

dmitry-timofeev:
To reviewers, the git does not detect the rename, so to see the changes you have to use:
```
dmitry-mac:exonum-java-binding user$ pwd
/Users/user/Documents/exonum-java-binding/exonum-java-binding
dmitry-mac:exonum-java-binding user$ git diff main/master:./core/rust/src/runtime/cmd.rs HEAD:./core/rust/src/cmd/mod.rs
```

exonum-java-binding/core/rust/src/cmd/mod.rs
```diff
@@ -0,0 +1,627 @@
...
+        for path in &self.standard.public_configs {
+            public_configs.push(load_config_file(path)?);
+        }
+        let supervisor_mode = public_configs[0].supervisor_mode;
+
+        if public_configs
+            .iter()
+            .any(|config| config.supervisor_mode != supervisor_mode)
+        {
+            return Err(format_err!("Different supervisor modes in public configs"));
```
dmitry-timofeev: I'd include expected/actual or just a set of the modes, as we can't tell what's expected.

exonum-java-binding/core/rust/src/cmd/mod.rs
```diff
@@ -0,0 +1,627 @@
...
+    supervisor_mode: SupervisorMode,
+}
+
+impl EjbCommand for GenerateConfig {
+    fn execute(self) -> Result<EjbCommandResult, failure::Error> {
+        let ejb_template_config: EjbTemplateConfig =
+            load_config_file(&self.standard.common_config)?;
+        let supervisor_mode = ejb_template_config.supervisor_mode;
+
+        if let StandardResult::GenerateConfig {
```
dmitry-timofeev: Does anyone consider the reverse-ordered pattern matching readable (first
pattern-match, then perform the operation producing the result)?
I'd consider rewriting to a more readable syntax: execute and then do pattern matching.

exonum-java-binding/core/rust/src/cmd/mod.rs
```diff
@@ -0,0 +1,627 @@
...
+        run_node_with_supervisor(SupervisorMode::Simple).unwrap();
+    }
+
+    #[test]
+    fn run_node_with_decentralized_supervisor() {
+        run_node_with_supervisor(SupervisorMode::Decentralized).unwrap();
+    }
+
+    #[test]
+    fn different_suprevisor_modes_in_public_configs() -> Result<(), failure::Error> {
```
dmitry-timofeev: suprevisor

exonum-java-binding/core/rust/src/cmd/mod.rs
```diff
@@ -0,0 +1,627 @@
...
+
+        Ok(())
+    }
+
+    fn ejb_shared_config(supervisor_mode: SupervisorMode) -> EjbSharedConfig {
+        EjbSharedConfig {
+            standard: SharedConfig {
+                common: Default::default(),
+                node: NodePublicConfig {
+                    address: "".to_string(),
```
dmitry-timofeev: Is it a valid value?

exonum-java-binding/core/rust/src/cmd/supervisor_mode.rs
```diff
@@ -0,0 +1,53 @@
...
+
+impl FromStr for SupervisorMode {
+    type Err = failure::Error;
+
+    fn from_str(input: &str) -> Result<Self, Self::Err> {
+        match input.to_ascii_lowercase().as_str() {
+            "simple" => Ok(SupervisorMode::Simple),
+            "decentralized" => Ok(SupervisorMode::Decentralized),
+            _ => Err(format_err!(
+                "Invalid supervisor mode. Can be \"simple\" or \"decentralized\""
```
dmitry-timofeev: `Invalid supervisor mode: <actual>. ...`

Can also use single quotation marks ' to not have to escape them.

dmitry-timofeev:
It must be WIP according to your recent message in mm.

vitvakatu:
Closing now, the functionality is moved to the core

PR #1274, opened 2019-12-07
Add transaction methods extraction class [ECR-3923]

MakarovS:
## Overview
Add transaction methods extraction class. It will be used by `ServiceWrapper` which would store a map of transaction methods and invoke them by an id.

---
See: https://jira.bf.local/browse/ECR-3923

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/TransactionMethodExtractorTest.java
```diff
@@ -0,0 +1,137 @@
...
+  void validServiceMethodExtraction() throws Exception {
+    Map<Integer, MethodHandle> transactions =
+        TransactionMethodExtractor.extractTransactionMethods(ValidService.class);
+    Method transactionMethod =
+        ValidService.class.getMethod("transactionMethod", byte[].class, TransactionContext.class);
+    MethodHandles.Lookup lookup = MethodHandles.lookup();
+    MethodHandle methodHandle = lookup.unreflect(transactionMethod);
+    assertThat(transactions).hasSize(1);
+    MethodHandle actualMethodHandle = transactions.get(ValidService.TRANSACTION_ID);
+    assertThat(actualMethodHandle.toString()).isEqualTo(methodHandle.toString());
```
MakarovS: I had `assertThat(transactions).containsExactly(entry(ValidService.TRANSACTION_ID, methodHandle));` initially, but `methodHandles` are not equal to each other. Is there a better way to check this?

exonum-java-binding/core/src/main/java/com/exonum/binding/core/transaction/TransactionMethod.java
```diff
@@ -0,0 +1,40 @@
...
+
+package com.exonum.binding.core.transaction;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Indicates that a method is a transaction method. The annotated method should:
```
dmitry-timofeev: The documentation is inadequate — see the Transaction and Transaction#execute
(since this annotation applies to methods, it must be documented appropriately).

exonum-java-binding/core/src/main/java/com/exonum/binding/core/transaction/TransactionMethod.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+// TODO: rename to Transaction after migration?
+public @interface TransactionMethod {
+
+  /**
+   * Returns the transaction type identifier which is unique within the service.
+   */
+  int id();
```
dmitry-timofeev: As there are no other properties than 'id', it is sensible to use 'value' instead.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/TransactionMethodExtractor.java
```diff
@@ -0,0 +1,93 @@
...
+ * Finds and validates transaction methods in a service.
+ */
+final class TransactionMethodExtractor {
+
+  /**
+   * Returns a map of transaction ids to transaction methods found in a service class.
+   *
+   * @see TransactionMethod
+   */
+  static Map<Integer, MethodHandle> extractTransactionMethods(Class<?> serviceClass) {
```
dmitry-timofeev: What's the point of producing method handles of unknown signature? Is it
  (byte[], TransactionContext) or (TransactionContext, byte[])? The caller won't be able
   to determine that. It also won't work with arbitrary protobuf messages as arguments even if the order becomes fixed.

   I think some object that can be invoked with context and byte[] and invoke the corresponding transaction method is needed.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/TransactionMethodExtractor.java
```diff
@@ -0,0 +1,93 @@
...
+    Map<Integer, MethodHandle> transactions = new HashMap<>();
+    MethodHandles.Lookup lookup = MethodHandles.lookup();
+    while (serviceClass != Object.class) {
```
dmitry-timofeev: This method does several things making testing more complicated:

   * Finds all methods annotated with a particular annotation
   * Validates these annotations
   * Converts to something callable

At least the first step, contributing to this outer loop, must be extracted and tested
      separately.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/TransactionMethodExtractor.java
```diff
@@ -0,0 +1,93 @@
...
+    return transactions;
+  }
+
+  /**
+   * Checks that the given transaction method signature is correct.
+   */
+  private static void validateTransactionMethod(Method transaction, Class<?> serviceClass,
+      Map<Integer, MethodHandle> transactions, int transactionId) {
+    checkArgument(!transactions.containsKey(transactionId),
+        "Service %s had more than one transaction with id: %s", serviceClass.getName(),
```
dmitry-timofeev: `Service %s has more than one transaction with the same id (%s): first: %s; second: %s`

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/TransactionMethodExtractor.java
```diff
@@ -0,0 +1,93 @@
...
+        transactionId);
+
+    String errorMessage = String.format("Method %s in a service class %s annotated with"
+        + " @TransactionMethod should have precisely two parameters of the following types:"
+        + " \"byte[]\" and \"com.exonum.binding.core.transaction.TransactionContext\"",
```
dmitry-timofeev: Consider using ' instead of escaped ".

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/TransactionMethodExtractor.java
```diff
@@ -0,0 +1,93 @@
...
+        transaction.getName(), serviceClass.getName());
+    checkArgument(transaction.getParameterCount() == 2, errorMessage);
+    for (Class<?> parameterType: transaction.getParameterTypes()) {
+      checkArgument(isParameterTypeValid(parameterType), errorMessage);
```
dmitry-timofeev: Include the invalid `parameterType`. The present message might be enough now, but including the 'bad' parameter will become more important when we allow more than `byte[]`

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/TransactionMethodExtractor.java
```diff
@@ -0,0 +1,93 @@
...
+    checkArgument(!transactions.containsKey(transactionId),
+        "Service %s had more than one transaction with id: %s", serviceClass.getName(),
+        transactionId);
+
+    String errorMessage = String.format("Method %s in a service class %s annotated with"
+        + " @TransactionMethod should have precisely two parameters of the following types:"
+        + " \"byte[]\" and \"com.exonum.binding.core.transaction.TransactionContext\"",
+        transaction.getName(), serviceClass.getName());
+    checkArgument(transaction.getParameterCount() == 2, errorMessage);
+    for (Class<?> parameterType: transaction.getParameterTypes()) {
```
dmitry-timofeev: This code accepts (byte[], byte[]) but shouldn't. Please test that properly.

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/TransactionMethodExtractorTest.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+class TransactionMethodExtractorTest {
+
+  @Test
+  void validServiceMethodExtraction() throws Exception {
+    Map<Integer, MethodHandle> transactions =
+        TransactionMethodExtractor.extractTransactionMethods(ValidService.class);
```
dmitry-timofeev: I think local classes will be quite appropriate.

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/TransactionMethodExtractorTest.java
dmitry-timofeev: If _invocation_ of the transaction method were implemented as the task specifies,
it could be tested by invoking the operation and verifying that it was indeed invoked:

```java
Object service = spy(new Service());
TransactionInvoker invoker = new TransactionInvoker(service);
invoker.invoke(id, context, bytes);

verify(service).operation(context, bytes);
```

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/TransactionMethodExtractorTest.java
```diff
@@ -0,0 +1,137 @@
...
+            DuplicateTransactionIdsService.class.getName(),
+            DuplicateTransactionIdsService.TRANSACTION_ID));
+  }
+
+  @Test
+  void missingTransactionMethodArgumentsServiceMethodExtraction() throws Exception {
+    Exception e = assertThrows(IllegalArgumentException.class,
+        () -> TransactionMethodExtractor
+            .extractTransactionMethods(MissingTransactionMethodArgumentsService.class));
+    Method transactionMethod =
```
dmitry-timofeev: (here and elsewhere) I don't think it is appropriate to use reflection to only specify the expected method name.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/transaction/TransactionMethod.java
```diff
@@ -0,0 +1,57 @@
...
+ * transaction cannot be executed normally and has to be rolled back. The transaction will be
+ * committed as failed (error kind {@linkplain ErrorKind#SERVICE SERVICE}), the
+ * {@linkplain ExecutionError#getCode() error code} with the optional description will be saved
+ * into the storage. The client can request the error code to know the reason of the failure.
+ *
+ * <p>The annotated method might also throw {@linkplain RuntimeException} if an unexpected error
+ * occurs. A correct transaction implementation must not throw such exceptions. The transaction
+ * will be committed as failed (status "panic").
+ *
+ * @see <a href="https://exonum.com/doc/version/0.12/architecture/transactions">Exonum Transactions</a>
```
dmitry-timofeev: `0.13-rc.2`, see https://wiki.bf.local/display/EJB/Java+Binding+0.9.0-rc1+Release+Checklist

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/TransactionInvoker.java
```diff
@@ -0,0 +1,76 @@
...
+  private final Map<Integer, MethodHandle> transactionMethods;
+
+  TransactionInvoker(Service service) {
+    this.service = service;
+    this.transactionMethods =
+        TransactionMethodExtractor.extractTransactionMethods(service.getClass());
+  }
+
+  /**
+   * Execute the transaction, possibly modifying the blockchain state.
```
dmitry-timofeev: I feel that the copy-pasted _user_-facing documentation is not appropriate here.
This one may be more concise and bring a different message (what it *does* for which Service
implementations, not what it *shall* do as an interface implementation). The spec is different, and the target audience is completely different.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/TransactionInvoker.java
```diff
@@ -0,0 +1,76 @@
...
+   *     failure
+   * @throws IllegalArgumentException if there is no transaction method with given id in a
+   *     corresponding service
+   * @throws RuntimeException if an unexpected error occurs. A correct transaction implementation
+   *     must not throw such exceptions. The transaction will be committed as failed
+   *     (status "panic")
+   */
+  void invokeTransaction(int transactionId, byte[] arguments, TransactionContext context)
+      throws TransactionExecutionException {
+    checkArgument(transactionMethods.containsKey(transactionId),
```
dmitry-timofeev: (Meta) We must extend the task on errors to cover routing errors — this is an instance of such error.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/TransactionInvoker.java
```diff
@@ -0,0 +1,76 @@
...
+        "No method with transaction id (%s)", transactionId);
+    try {
+      MethodHandle methodHandle = transactionMethods.get(transactionId);
```
dmitry-timofeev: That's fine for this PR, but I'd remind that MHs won't work for _any_ message (next task):

> The caller won't be able
to determine that. It also won't work with arbitrary messages as arguments even if
the order is fixed.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/TransactionInvoker.java
```diff
@@ -0,0 +1,76 @@
...
+      methodHandle.invoke(service, arguments, context);
+    } catch (Throwable throwable) {
+      if (throwable instanceof TransactionExecutionException) {
+        throw (TransactionExecutionException) throwable;
+      } else {
+        throw new RuntimeException(throwable);
```
dmitry-timofeev: (Possibly for future improvement) The MH#invoke spec says
> Throws:
Throwable – anything thrown by the underlying method propagates unchanged through the method handle call

So why don't we _forbid_ checked exceptions in transaction methods (except TransactionExecutionException)
during _discovery_ and throw RuntimeExceptions *as is* on invocation?

Please create an issue for that.

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/TransactionInvokerTest.java
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  @Test
+  void invokeThrowingTransaction() {
+    TransactionInvoker invoker = new TransactionInvoker(new ThrowingService());
+    TransactionExecutionException e = assertThrows(TransactionExecutionException.class,
+        () -> invoker.invokeTransaction(ThrowingService.TRANSACTION_ID, ARGUMENTS, context));
+    assertThat(e.getErrorCode()).isEqualTo(ThrowingService.ERROR_CODE);
+  }
+
```
dmitry-timofeev: Please test handling of 'other' _service originated_ exceptions, since it is important to handle them properly. 

May add a second method to 'ThrowingService'.

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/TransactionInvokerTest.java
```diff
@@ -0,0 +1,110 @@
...
+  class BasicService implements Service {
```
dmitry-timofeev: The classes below are unnecessarily nested classes of the test class — shall be made `static`.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/transaction/TransactionMethod.java
```diff
@@ -0,0 +1,57 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Indicates that a method is a transaction method. The annotated method should execute the
+ * transaction, possibly modifying the blockchain state. The method should:
+ * <ul>
+ *   <li>be public
+ *   <li>have exactly two parameters of types 'byte[]' and '{@link TransactionContext}' in this
```
dmitry-timofeev: Parameter documentation (most importantly, what is `byte[]`, also what is `context` for)?
The `com.exonum.binding.core.service.TransactionConverter#toTransaction(int, byte[])` and `Transaction#execute` have it correspondingly.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/TransactionMethodExtractor.java
```diff
@@ -0,0 +1,110 @@
...
+final class TransactionMethodExtractor {
+
+  /**
+   * Returns a map of transaction ids to transaction methods found in a service class.
+   *
+   * @see TransactionMethod
+   */
+  static Map<Integer, MethodHandle> extractTransactionMethods(Class<?> serviceClass) {
+    Map<Integer, Method> transactionMethods = findTransactionMethods(serviceClass);
+    Lookup lookup = MethodHandles.lookup();
```
dmitry-timofeev: this code uses relative lookup (relative to this class) but does not need it (it
doesn't access any private or package-private members, for instance, which such call-site sensitive
lookup provides). Why don't we use
publicLookup (likely, 'relative' to the defining class to satisfy class-loading constraints
we bumped into earlier with protobuf reflective serializer)?

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/TransactionMethodExtractor.java
```diff
@@ -0,0 +1,110 @@
...
+      }
+      serviceClass = serviceClass.getSuperclass();
+    }
+    return transactionMethods;
+  }
+
+  private static void checkDuplicates(Map<Integer, Method> transactionMethods, int transactionId,
+      Class<?> serviceClass, Method method) {
+    if (transactionMethods.containsKey(transactionId)) {
+      String errorMessage = String.format("Service %s has more than one transaction with the same"
```
dmitry-timofeev: (Nit): this one is actually first (given the order of traversal). Also, variable.

```suggestion
      String firstMethodName = transactionMethods.get(transactionId).getName();
      String errorMessage = String.format("Service %s has more than one transaction with the same"
              + " id (%s): first: %s; second: %s",
          serviceClass.getName(), transactionId, firstMethodName, method.getName());
```

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/TransactionMethodExtractorTest.java
```diff
@@ -0,0 +1,185 @@
...
+    @TransactionMethod(TRANSACTION_ID)
+    void transactionMethod(byte[] arguments, TransactionContext context);
+  }
+
+  class ValidServiceInterfaceImplementation implements ServiceInterface {
+
+    static final int TRANSACTION_ID_2 = 2;
+
+    @TransactionMethod(TRANSACTION_ID)
+    public void transactionMethod(byte[] arguments, TransactionContext context) {}
```
dmitry-timofeev: I considered this case (the present *duplicates* the annotation which isn't something that shall happen, I think):
```java
  interface ServiceInterface {

    @TransactionMethod(1) // <-- Annotation is present on the interface
    void transactionMethod(byte[] arguments, TransactionContext context);
  }

  class ServiceInterfaceImpl implements ServiceInterface {

    @Override // <-- NO duplicate TransactionMethod annotation!
    public void transactionMethod(byte[] arguments, TransactionContext context) {}

  }
```

I think it is better to offload this to a sep. issue because there are more to that: e.g., what if an interface method and the implementation method have *different* @TransactionMethods?

Please document the considerations in a new issue.

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/TransactionMethodExtractorTest.java
```diff
@@ -0,0 +1,185 @@
...
+
+    @TransactionMethod(TRANSACTION_ID_2)
+    @SuppressWarnings("WeakerAccess") // Should be accessible
+    public void transactionMethod2(byte[] arguments, TransactionContext context) {}
```
dmitry-timofeev: A bike-shed target (@slowli , @bullet-tooth , @MakarovS, @vitvakatu )! Shall we require the arguments to be in order:
1. (MessageT arguments, TransactionContext context)
2. (TransactionContext context, MessageT arguments)
3. Allow any (mind extra impl. complexity!)

vitvakatu: I think there is no need in arbitrary order. The either one works, but Rust runtime uses `context, arguments` one.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/TransactionInvoker.java
MakarovS: Created ECR-3988

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/TransactionMethodExtractorTest.java
MakarovS: Created ECR-3861

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/TransactionMethodExtractorTest.java
MakarovS: I agree with allowing any order, there'd be a couple more lines of code in implementation, but less unnecessary restrictions for users.
Created ECR-3990

dmitry-timofeev: I agree that it is not badly needed, and any could work: in some cases,  the `arguments` is the first argument used (e.g., to check preconditions); in some cases, the `context` (e.g., to instantiate the schema and check the preconditions _depending on the state_, or to just update it).

slowli: No particular preference. From the unification perspective, `context, arguments` ordering makes sense, but AFAIK, this is not the Java way™. Since reflection is used to handle the processing, restricting the ordering of args is not necessary. (Further, it could make sense to allow to extract commonly used parts of the context into args, e.g., the service schema and the caller.)

PR #1231, opened 2019-11-21
ECR-3591

ipenkin:
## Overview
Integration test for native `Runtime` proxy. Tests scope: 
- [ ] Exception handling mechanism
- [ ] Runtime instantiation and shutdown
- [ ] Deploy valid/invalid service artifacts
- [ ] Instantiation of valid/invalid service artifacts
- [ ] Runtime state-hashes 
- [ ] Runtime before/after commit method
- [ ] Transaction execution (positive cases)
- [ ] Transaction execution (negative cases)

---
See: https://jira.bf.local/browse/ECR-3591

### Definition of Done

- [ ] There are no TODOs left in the code
- [ ] Change is covered by automated [tests](https://github.com/exonum/exonum-java-binding/blob/master/CONTRIBUTING.md#tests)
- [ ] The [coding guidelines](https://github.com/exonum/exonum-java-binding/blob/master/CONTRIBUTING.md#the-code-style) are followed
- [ ] The continuous integration build passes

PR #1287, opened 2019-12-11
Make `assert_panics()` a library function [ECR-3353].

skletsun:
## Overview

`assert_panics()` should be used instead of `#[should_panic(expected = "Expected message")]` in tests in case there is more than one way of panicking with the expected message.

---
See: https://jira.bf.local/browse/ECR-3353

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/rust/integration_tests/tests/jni_cache_not_initialized.rs
+++ b/exonum-java-binding/core/rust/integration_tests/tests/jni_cache_not_initialized.rs
@@ -14,10 +14,11 @@
 
 extern crate java_bindings;
 
-use java_bindings::utils::jni_cache;
+use java_bindings::utils::{assert_panics, jni_cache};
 
 #[test]
-#[should_panic(expected = "JNI cache is not initialized")]
 fn cache_not_initialized() {
-    jni_cache::runtime_adapter::execute_tx_id();
+    assert_panics("JNI cache is not initialized", || {
+        jni_cache::runtime_adapter::execute_tx_id()
+    });
 }
--- a/exonum-java-binding/core/rust/src/handle/resource_manager/imp.rs
+++ b/exonum-java-binding/core/rust/src/handle/resource_manager/imp.rs
@@ -87,6 +87,7 @@ pub fn known_handles() -> usize {
 mod tests {
     use super::*;
     use std::i64;
+    use utils::assert_panics;
 
     enum T {}
     const INVALID_HANDLE: Handle = i64::MAX;
@@ -128,10 +129,11 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "Trying to add the same handle for the second time")]
     fn add_duplicated_handle() {
         add_handle::<T>(DUPLICATED_HANDLE);
-        add_handle::<T>(DUPLICATED_HANDLE);
+        assert_panics("Trying to add the same handle for the second time", || {
+            add_handle::<T>(DUPLICATED_HANDLE)
+        });
     }
 
     #[test]
@@ -159,10 +161,11 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "Wrong type id for")]
     fn check_wrong_type_handle() {
         add_handle::<T>(WRONG_TYPE_HANDLE);
         enum OtherT {}
-        check_handle::<OtherT>(WRONG_TYPE_HANDLE);
+        assert_panics("Wrong type id for", || {
+            check_handle::<OtherT>(WRONG_TYPE_HANDLE)
+        });
     }
 }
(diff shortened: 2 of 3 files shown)


exonum-java-binding/core/rust/integration_tests/tests/jni_cache_not_initialized.rs
```diff
@@ -14,10 +14,11 @@
...
 extern crate java_bindings;
 
-use java_bindings::utils::jni_cache;
+use java_bindings::utils::{assert_panics, jni_cache};
 
 #[test]
-#[should_panic(expected = "JNI cache is not initialized")]
 fn cache_not_initialized() {
-    jni_cache::runtime_adapter::execute_tx_id();
+    assert_panics("JNI cache is not initialized", || {
```
skletsun: There is no ambiguity in the test - I've just put it here in order to just make sure the function is accessible from the integration tests as well.

exonum-java-binding/core/rust/src/handle/resource_manager/imp.rs
```diff
@@ -128,10 +129,11 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "Trying to add the same handle for the second time")]
     fn add_duplicated_handle() {
         add_handle::<T>(DUPLICATED_HANDLE);
-        add_handle::<T>(DUPLICATED_HANDLE);
+        assert_panics("Trying to add the same handle for the second time", || {
```
dmitry-timofeev: that's the perfect replacement :+1:

exonum-java-binding/core/rust/src/utils/mod.rs
```diff
@@ -27,3 +27,20 @@ pub use self::errors::{
     panic_on_exception, unwrap_exc_or, unwrap_exc_or_default, unwrap_jni, unwrap_jni_verbose,
 };
 pub use self::jni::{get_class_name, get_exception_message};
+
+/// Asserts that given closure panics while executed and the resulting error message contains given
+/// substring.
+#[cfg(test)]
+pub fn assert_panics<F, R>(err_substring: &str, f: F)
```
dmitry-timofeev: I remember we had this method somewhere — was it lost with some obsolete native ITs?

dmitry-timofeev: Shall it be a part of the main code?

skletsun: Yes, it's gone within ECR-3480 with obsolete service/transactions proxy tests afair.

skletsun: I was unable to come up with scenarios of using this outside the testing scope for the moment 🙂

vitvakatu: Lol, deja vu: https://github.com/exonum/exonum-java-binding/pull/1231/files#diff-19687bf9dc0b70f9c7df7e63d6a5d2dbR646

exonum-java-binding/core/rust/src/utils/mod.rs
```diff
@@ -27,3 +27,20 @@ pub use self::errors::{
...
+where
+    F: FnOnce() -> R,
+{
+    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(f));
+    match result {
+        Ok(_) => panic!("Panic expected"),
+        Err(err) => {
+            let err_msg = any_to_string(&err);
+            assert!(err_msg.contains(err_substring));
```
dmitry-timofeev: Is that correct that it won't show neither a substring nor the actual string? 

I'd suggest including both in the form 
```
Expecting [panic] message ("%s") to contain "%s"
```

Or, as in Java:
```
Expected a panic message containing "bar", but was "foo"
```

exonum-java-binding/core/rust/src/utils/mod.rs
skletsun: Clippy seems unhappy with integration tests becase of this, thus put it to the main code.

dmitry-timofeev: Just to clarify: I asked if this function had to be a part of main code because I thought it _already_ was based on its location (exonum-java-binding/core/rust/src/utils/mod.rs, I missed initially `#[cfg(test)]`), and considered it as not needed in the main code (as it is only for testing) just as you did. So, can it be test-only? Shall we suppress clippy or is there a real problem?

skletsun: It turned out that code under the `#[cfg(test)]` annotation is not accessible for the integration tests thus this should be a part of the main code. I looked for other similar helper functions (that should be accessible for both kinds of tests only) but this one seems to be the only one. Is it worth creating a separate module (says, `test_utils`) for it or better to leave it here?

exonum-java-binding/core/rust/src/utils/mod.rs
skletsun: After some testing, I tend to the second variant because it fits better for multiline actual strings. Also, for better readability I print the actual message from the new line.

PR #1289, opened 2019-12-12
Add Supervisor service to the list of service factories [ECR-3841]

vitvakatu:
## Overview

The App was broken (probably when updating to the latest Exonum revision). 

---
See: https://jira.bf.local/browse/ECR-3841

### Definition of Done

- [ ] Public API has Javadoc


vitvakatu:
> Can node instantiation be reasonably tested here?

No, it must be tested in system tests, because running the node requires infinite event loop (it is pretty hard to stop running node without sending SIGINT), and just instantiating node (without running it) will not catch such errors.

PR #1296, opened 2019-12-16
Update README.md and setup.py for Python modules [ECR-3992]

vitvakatu:
## Overview

Python modules have been published with the following changes.

---
See: https://jira.bf.local/browse/ECR-3992

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/exonum_launcher_java_plugins/README.md
+++ b/exonum-java-binding/exonum_launcher_java_plugins/README.md
@@ -6,9 +6,17 @@
 
 ## Installation
 
-#### Requirements
+### Requirements
 
 - [Python3](https://www.python.org/downloads/)
+
+### Using PyPI
+
+```bash
+pip3 install exonum-launcher-java-plugins
+```
+
+### From Source
 
 ```bash
 # Generate sources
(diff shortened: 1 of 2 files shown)


dmitry-timofeev:
@popzxc Would you please review the extension updates?

dmitry-timofeev:
Why does it use a different version from the rest of the project?

vitvakatu:
> Why does it use a different version from the rest of the project (on Pypi)?

A small oversight by me. PyPI doesn't like certain version formats, like `0.9.0-rc1`, or `0.9.0-rc1.2` which are totally possible in semantic versioning. So I changed it to `dev1` for testing purpose and forgot to change it back.

I published another release with version `0.9.0rc1`, it is now available at PyPI.

dmitry-timofeev:
> PyPI doesn't like certain version formats

Does it forbid some version strings, allowed by semver?

popzxc:
>Does it forbid some version strings, allowed by semver?

Even if it doesn't, it's not consistent with convention. Even the `cpython` project itself uses the following format:
image

PR #1299, opened 2019-12-17
Separate JNI implementation for ProofMap [ECR-3983]

skletsun:
## Overview

Hashing and non-hashing ProofMap have different JNI implementations since now.

---
See: https://jira.bf.local/browse/ECR-3983

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/rust/src/storage/raw_proof_map_index.rs
+++ b/exonum-java-binding/core/rust/src/storage/raw_proof_map_index.rs
@@ -474,8 +474,9 @@ fn convert_to_key(env: &JNIEnv, array: jbyteArray) -> JniResult<Key> {
     assert_eq!(
         key.len(),
         PROOF_MAP_KEY_SIZE,
-        "Key size expected to be {} bits",
-        PROOF_MAP_KEY_SIZE * 8
+        "Key size expected to be {} bytes, found {} bytes",
+        PROOF_MAP_KEY_SIZE,
+        key.len()
     );
     let mut result: Key = [0; PROOF_MAP_KEY_SIZE];
     result.copy_from_slice(key.as_slice());
--- a/exonum-java-binding/core/rust/src/storage/raw_proof_map_index.rs
+++ b/exonum-java-binding/core/rust/src/storage/raw_proof_map_index.rs
@@ -47,7 +47,7 @@ enum IndexType {
 
 type Iter<'a> = PairIter<ProofMapIndexIter<'a, Key, Value>>;
 
-/// Returns a pointer to the created `ProofMapIndex` object.
+/// Returns a pointer to the created `RawProofMapIndex` object.
 #[no_mangle]
 pub extern "system" fn Java_com_exonum_binding_core_storage_indices_RawProofMapIndexProxy_nativeCreate(
     env: JNIEnv,
@@ -71,7 +71,7 @@ pub extern "system" fn Java_com_exonum_binding_core_storage_indices_RawProofMapI
     utils::unwrap_exc_or_default(&env, res)
 }
 
-/// Returns a pointer to the created `ProofMapIndex` instance in an index family (= group).
+/// Returns a pointer to the created `RawProofMapIndex` instance in an index family (= group).
 #[no_mangle]
 pub extern "system" fn Java_com_exonum_binding_core_storage_indices_RawProofMapIndexProxy_nativeCreateInGroup(
     env: JNIEnv,
@@ -95,7 +95,7 @@ pub extern "system" fn Java_com_exonum_binding_core_storage_indices_RawProofMapI
     utils::unwrap_exc_or_default(&env, res)
 }
 
-/// Destroys the underlying `ProofMapIndex` object and frees memory.
+/// Destroys the underlying `RawProofMapIndex` object and frees memory.
 #[no_mangle]
 pub extern "system" fn Java_com_exonum_binding_core_storage_indices_RawProofMapIndexProxy_nativeFree(
     env: JNIEnv,
@@ -404,7 +404,7 @@ pub extern "system" fn Java_com_exonum_binding_core_storage_indices_RawProofMapI
     utils::unwrap_exc_or(&env, res, ptr::null_mut())
 }
 
-/// Destroys the underlying `ProofMapIndex` iterator object and frees memory.
+/// Destroys the underlying `RawProofMapIndex` iterator object and frees memory.
 #[no_mangle]
 pub extern "system" fn Java_com_exonum_binding_core_storage_indices_RawProofMapIndexProxy_nativeEntriesIterFree(
     env: JNIEnv,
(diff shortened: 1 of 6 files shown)


exonum-java-binding/core/rust/src/storage/raw_proof_map_index.rs
```diff
@@ -0,0 +1,483 @@
...
+const JAVA_ENTRY_FQN: &str = "com/exonum/binding/core/storage/indices/MapEntryInternal";
+
+enum IndexType {
+    SnapshotIndex(Index<&'static dyn Snapshot>),
+    ForkIndex(Index<&'static Fork>),
+}
+
+type Iter<'a> = PairIter<ProofMapIndexIter<'a, Key, Value>>;
+
+/// Returns a pointer to the created `ProofMapIndex` object.
```
vitvakatu: Here and more: `RawProofMapIndex`?

exonum-java-binding/core/rust/src/storage/raw_proof_map_index.rs
```diff
@@ -0,0 +1,483 @@
...
+    handle::drop_handle::<ProofMapIndexValues<Value>>(&env, iter_handle);
+}
+
+// Converts Java byte array to key.
+fn convert_to_key(env: &JNIEnv, array: jbyteArray) -> JniResult<Key> {
+    let key = env.convert_byte_array(array)?;
+    assert_eq!(
+        key.len(),
+        PROOF_MAP_KEY_SIZE,
+        "Key size should be {} bits",
```
vitvakatu:
```suggestion
        "Key size expected to be {} bits, found {} bits",
```

vitvakatu: Btw, why bits, not bytes?

skletsun: I'm totally OK with bytes but bits seem to be more traditional.

exonum-java-binding/core/rust/src/utils/conversion.rs
```diff
@@ -30,10 +31,38 @@ pub fn convert_hash(env: &JNIEnv, hash: &Hash) -> JniResult<jbyteArray> {
...
+/// Converts anything convertible to a protobuf message into Java byte array.
+pub fn proto_to_java_bytes(
+    env: &JNIEnv,
+    proto: impl ProtobufConvert<ProtoStruct = impl Message>,
+) -> JniResult<jbyteArray> {
+    let bytes = proto.to_pb().write_to_bytes().unwrap();
+    env.byte_array_from_slice(&bytes)
+}
+
+/// Converts array of Java bytes arrays to the vector of Rust array representation.
```
dmitry-timofeev: I'd also add a type declaration (`byte[][]`) to ease understanding.

exonum-java-binding/core/rust/src/utils/conversion.rs
```diff
@@ -30,10 +31,38 @@ pub fn convert_hash(env: &JNIEnv, hash: &Hash) -> JniResult<jbyteArray> {
...
+    env: &JNIEnv,
+    array: jobjectArray,
+    to_rust_array: F,
+) -> JniResult<Vec<T>>
+where
+    F: Fn(&JNIEnv, jbyteArray) -> JniResult<T>,
+{
+    let num_elements = env.get_array_length(array)?;
+    let mut result = Vec::with_capacity(num_elements as usize);
+    for i in 0..num_elements {
```
dmitry-timofeev: With this basically `.map().collect(toList())`, have we considered implementing an Iterator of auto_locals for `jobjectArray`s? I am not sure we shall (put that into the jni library) because it will become a source of leaks (i.e., any operation producing new locals — filter, map, etc. — shall manage them properly) unless the user is well aware of the responsibility.

dmitry-timofeev:
6810fc0dc762bcae55de14e83467cf26649514d8 seems to be the last revision before hashing mods. To review the diff against the base code:
```
git fetch skletsun ecr-3983-proofmap-separation-jni

git diff 6810fc0dc762bcae55de14e83467cf26649514d8:exonum-java-binding/core/rust/src/storage/proof_map_index.rs skletsun/ecr-3983-proofmap-separation-jni:exonum-java-binding/core/rust/src/storage/proof_map_index_next.rs

git diff 6810fc0dc762bcae55de14e83467cf26649514d8:exonum-java-binding/core/rust/src/storage/proof_map_index.rs skletsun/ecr-3983-proofmap-separation-jni:exonum-java-binding/core/rust/src/storage/raw_proof_map_index.rs
```

exonum-java-binding/core/rust/src/storage/raw_proof_map_index.rs
dmitry-timofeev: That's an internal error, so it does not matter much (bits/bytes, it is checked in Java by

https://github.com/exonum/exonum-java-binding/blob/0b344d5f530666a3a8aa39faa7cb1b9a12aaab90/exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/ProofMapKeyCheckingSerializerDecorator.java#L49-L53

, but both actual and expected shall be included.

dmitry-timofeev:
@vitvakatu has said once he wasn't happy with our native storage code of almost identical functions — double that :-)

exonum-java-binding/core/rust/src/storage/raw_proof_map_index.rs
skletsun: Ok, switched it to bytes - with both values printed this seems more suitable.

exonum-java-binding/core/rust/src/utils/conversion.rs
dmitry-timofeev: @skletsun , @vitvakatu ?

vitvakatu: I don't see the way how `filter` or `map` usage will lead to autolocals leaks 🤔

skletsun: If I get things right `filter`/`map`/`whatever` should not produce new local but operate with existing ones created by an iterator. And at first look, the user should only be aware that object disappears after iteration unless global reference was created.
The problem that I see right now is that implementing, for example, `IntoIterator` for `jobjectArray` means implementing it for any object because `jobjectArray` is not a separate type but just a type alias for `jarray` which in turn an alias for `jobject`.

dmitry-timofeev: I thought of something like
```
var yellow_bananas = env.stream(array_of_bananas)
   .filter(banana_auto_local -> {
    // Stupid code *creating a new local for color*
    Color brownColor = env.get_field(Color.Brown);
    return !env.call_method(banana_auto_local, "hasColor", brownColor)
  })
  .collect(toVec());
```

dmitry-timofeev: >
```

  .collect(toVec());
```

Also, array of `AutoLocal`s kind of contradits their purpose

dmitry-timofeev: I think we shall keep this in mind, in case we could find a good (safe, ergonomic) way to implement that.

PR #1302, opened 2019-12-17
Update to Exonum revision 7223ce6ebcf [ECR-3972]

vitvakatu:
## Overview

Updating to almost recent master. 

---
See: https://jira.bf.local/browse/ECR-3972

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/core/rust/exonum-java/Cargo.toml
```diff
@@ -5,7 +5,6 @@ authors = ["Exonum team <contact@exonum.com>"]
 
 [dependencies]
 env_logger = "0.7"
-exonum-btc-anchoring = "0.13.0-rc.2"
```
dmitry-timofeev: It never seems to stay here for long 🙃

vitvakatu: It doesn't even feel right to add a task to enable it later :D

dmitry-timofeev:
I'd update the proto-sources to keep them in sync.

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -116,7 +113,8 @@ impl JavaRuntimeProxy {
 
                 ExceptionHandlers::DEFAULT(env, exception)
             }
-            _ => (Error::OtherJniError, err).into(),
+            // TODO(ECR-4015)
+            _ => ExecutionError::new(ErrorKind::Runtime { code: 3 }, err.to_string()),
```
vitvakatu: I feel that I'm missing something here... @slowli am I?

exonum-java-binding/core/rust/src/storage/entry.rs
```diff
@@ -25,7 +25,8 @@ use handle::{self, Handle};
 use storage::db::{Value, View, ViewRef};
 use utils;
 
-type Index<T> = Entry<T, Value>;
+// TODO(ECR-3979): replace with Entry
+type Index<T> = ProofEntry<T, Value>;
```
vitvakatu: I decided to change it to ProofEntry temporarily, because index hash will be unavailable otherwise.
Side note: do we ever need to implement support for Entry?

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
```diff
@@ -75,7 +69,7 @@
   /**
    * The active services indexed by their name. It is stored in a sorted map that offers
    * the same iteration order on all nodes with the same services, which is required
-   * for correct operation of beforeCommit and {@link #getStateHashes(Snapshot)}.
+   * for correct operation of state hash aggregation.
```
vitvakatu: Should I place a link to Schema here?

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Schema.java
```diff
@@ -16,21 +16,22 @@
...
 
-import com.exonum.binding.common.hash.HashCode;
-import java.util.List;
+import com.exonum.binding.core.storage.indices.ProofListIndexProxy;
+import com.exonum.binding.core.storage.indices.ProofMapIndexProxy;
 
 /**
  * A schema of the collections (a.k.a. indices) of a service.
+ *
+ * <p>To verify the integrity of the database state on each node in the network,
```
vitvakatu: I don't know where is the proper place for this doc, but in Service it looks weird.

exonum-java-binding/core/rust/src/proto/mod.rs
```diff
@@ -17,6 +17,4 @@
 #![allow(bare_trait_objects)]
 #![allow(renamed_and_removed_lints)]
 
-pub use self::service_runtime::*;
```
vitvakatu: The whole module is unusable now, it seems. I don't want to delete it, because it might be useful in the future.

exonum-java-binding/core/rust/src/proxy/runtime.rs
slowli: I've answered in the ECR-4015 ticket.

vitvakatu: Thanks, I will correct this code

exonum-java-binding/core/rust/src/storage/entry.rs
dmitry-timofeev: The thing is, Java Entry does not have object_hash yet:

https://github.com/exonum/exonum-java-binding/blob/e096edce28f8aedc66487d9e4ed0fa1ec6c6f2aa/exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/EntryIndexProxy.java#L173-L174

vitvakatu: 🤔 Should I remove native getObjectHash then? It won't work with Entry

dmitry-timofeev: I see, @slowli do you think we shall provide both Entry & ProofEntry? 

If yes, I'd remove the method, and make all the needed Java modifications separately to avoid confusion (Java needs the op, tests, documentation).

dmitry-timofeev: The method impl seems to match the template for such methods and will be easy to copy-paste from an existing Proof index.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/AbstractService.java
```diff
@@ -58,11 +55,6 @@ protected final ServiceInstanceSpec getInstanceSpec() {
     return instanceSpec;
   }
 
-  @Override
-  public List<HashCode> getStateHashes(Snapshot snapshot) {
-    return createDataSchema(snapshot).getStateHashes();
-  }
-
   /**
    * Creates a data schema of this service.
```
dmitry-timofeev: (May be offloaded for discussion to a sep. issue — please add)
Shall this method remain? It's not just enough to override it, the return type must
also be adjusted (e.g., `@Override QaSchema createDataSchema(View) { }` in QaService)

The (small) value it still provides is that it explicitly connects a service to a schema.

exonum-java-binding/service-archetype/src/main/resources/archetype-resources/src/main/java/MySchema.java
```diff
@@ -39,12 +39,4 @@ public MySchema(View view, String serviceName) {
     this.view = checkNotNull(view);
     this.namespace = serviceName + ".";
   }
-
-  @Override
```
dmitry-timofeev: We might add here instead a TODO: Add index factories here

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -467,75 +450,22 @@ impl fmt::Display for JavaArtifactId {
...
-
 type ExceptionHandler = Fn(&JNIEnv, JObject) -> ExecutionError;
 struct ExceptionHandlers;
 
 impl ExceptionHandlers {
     const DEFAULT: &'static ExceptionHandler = &|env, exception| {
         assert!(!exception.is_null(), "No exception thrown.");
         let message = describe_java_exception(env, exception);
-        (Error::JavaException, message).into()
+        // TODO(ECR-4015):
```
dmitry-timofeev: It is closed as 'Won't do', it's not clear what needs to be done.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Schema.java
```diff
@@ -16,21 +16,22 @@
...
+ * the core automatically tracks every Merkelized collection used by the user
```
dmitry-timofeev: AFAIK, not every (e.g. grouped proofs are not tracked).

--- 

Is there much sense in groups of Proof indexes then or shall they be removed completely (from the core also):
they do not affect the state; you cannot create proofs from them. @slowli , WDYT?

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Schema.java
```diff
@@ -16,21 +16,22 @@
...
+ * services. It aggregates state hashes of these collections into a single
+ * Merkelized meta-map. The hash of this meta-map is considered the hash of the
+ * entire blockchain state and is recorded as such in blocks and Precommit
+ * messages.
+ *
```
dmitry-timofeev: The original doc had a warning on using plain indexes as a sep. paragraph:
 * > Please note that if this service does not provide any state hashes,
 *    * the framework will not be able to verify that its transactions cause the same results
 *    * on different nodes.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
dmitry-timofeev: Actually, it is no longer required to be a sorted map because no order-sensitive iteration occurs in the runtime (@slowli — will there be in a near future? Shall we replace with a HashMap?) — afterCommit is not such.

--- 

Even if there are none order-sensitive operations *but* we log all services (currently I don't see such logs) it might be reasonable to keep to ease log navigation from different nodes.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/AbstractService.java
vitvakatu: ECR-4029

exonum-java-binding/core/rust/src/proxy/runtime.rs
vitvakatu: Accidental artifact

exonum-java-binding/core/rust/src/storage/entry.rs
vitvakatu: Really funny, I wasn't able to change it back to `Entry` because than `getConsensusConfiguration` doesn't work - Core uses `ProofEntry` there. So I think it's better to have ProofEntry there before we implement it properly.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Schema.java
vitvakatu: I left it as it is now. If we decide to implement support for grouped proofs, or to remove them completely, the statement will be true. If we decide to leave it as it is now, the statement must be updated, but I'm not sure how to phrase it (how do we call non-grouped indexes?).

exonum-java-binding/core/rust/src/storage/entry.rs
dmitry-timofeev: OK, but let's rename it then — EntryIndexProxy -> ProofEntryIndexProxy (same with its ITs).

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Schema.java
slowli: > you cannot create proofs from them

You can if these proofs are tied to a more "high-level" proof. This is the case in `cryptocurrency-advanced`, where the history of each wallet is `ProofListIndex`, which is tied to the `state_hash` via the `ProofMapIndex` of wallets. The values of the PMI index (i.e., wallets) contain the `object_hash` of the PLI corresponding to the wallet history. This is clunky and requires effort from service developers, but it works.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
slowli: > will there be in a near future?

Not that I'm aware of. The general idea is that operations pertaining to services (e.g., `before_transactions` / `after_transactions` hooks) are ordered *by the core* and called for each service separately.

exonum-java-binding/core/rust/src/storage/entry.rs
slowli: > do you think we shall provide both Entry & ProofEntry?

These types both make sense, IMO. `ProofEntry` imposes certain overhead to compute the hash code of the contents each time it is set, while `Entry` does not. Also, since `Entry` is not aggregated into the state, it could be used for private data.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
dmitry-timofeev: That's good, thanks.

exonum-java-binding/core/rust/src/storage/entry.rs
dmitry-timofeev: OK, thanks, I think then we shall replace the present Entry with a ProofEntry (required for CoreSchema), and might add Entry as an improvement later (it won't be a breaky).

PR #1306, opened 2019-12-18
Document that a separate server with its own TCP port is used

dmitry-timofeev:


dmitry-timofeev:
> I thought the exonum.com/doc would be the primary target for such clarification.

I am not sure I get that: that's exactly what I linked?

vitvakatu:
I mean, [External service API section](https://exonum.com/doc/version/0.13-rc.2/get-started/java-binding/#external-service-api) does not have any note about using a different port rather than the one used by the core. The only real mention of separated server is in the section you linked.

dmitry-timofeev:
I see, that makes sense and has its own task: https://jira.bf.local/browse/ECR-3856

PR #1312, opened 2019-12-23
Use HTTPS for Maven POM Schema

dmitry-timofeev:
## Overview
The other things are namespace declarations
and are not retrievable resources, hence _must_ remain intact.

Also, remove tidy-maven-plugin: it forbids HTTPS in schema location, required
by the Maven POM Convention.

See also:
 - ECR-4051
 - ECR-4052

---

### Definition of Done


exonum-java-binding/pom.xml
```diff
@@ -202,7 +202,7 @@
       <!-- An empty artifact, required while JUnit 4 is on the classpath to override its
            dependency on hamcrest.
 
-           See http://hamcrest.org/JavaHamcrest/distributables#upgrading-from-hamcrest-1x
+           See https://hamcrest.org/JavaHamcrest/distributables#upgrading-from-hamcrest-1x
```
bullet-tooth: actually `http` works well but `https`:
image

dmitry-timofeev: Thanks, that was committed accidentally — reverted 🙃 

I actually noticed that issue yesterday and submitted it: https://github.com/hamcrest/JavaHamcrest/issues/280

PR #1318, opened 2019-12-24
Disable parallel Maven builds on Mac

dmitry-timofeev:
## Overview

Disables parallel Maven builds (running Maven goals for independent modules in parallel) as it seems to fix the failing builds on Mac.

---

### Definition of Done


dmitry-timofeev:
It appears the parallel _Maven_ builds make the Mac builds hang on Travis (they do work locally — however, with newer JDK and OS). 

Or just increase the likelihood of that.

Will disable them on Travis.

PR #1321, opened 2019-12-26
Implement correct error handling in native runtime code [ECR-4054]

vitvakatu:
## Overview

- runtime::error::Error enum is slightly changed. `OtherJniError` renamed to `JniError`, `IncorrectArtifactId` replaced with `IllegalArgument`.
- New helper function `get_exception_cause` to call `Throwable#getCause`
- New exception handlers in runtime module: `ILLEGAL_ARGUMENT` for `IllegalArgumentException` and `TX_UNEXPECTED` for `UnexpectedTransactionExecutionException`.
- New method `jni_call_transaction` to be used in `execute`, `before_transactions` and `after_transactions` methods.
- `jni_call_default` handles `IllegalArgumentException` additionally.

---
See: https://jira.bf.local/browse/ECR-4054

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/core/rust/src/runtime/error.rs
```diff
@@ -16,12 +16,14 @@
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, ExecutionFail)]
 #[execution_fail(kind = "runtime")]
 pub enum Error {
-    /// Unable to parse artifact identifier or specified artifact has non-empty spec.
-    IncorrectArtifactId = 0,
-    /// Checked java exception is occurred
+    /// Illegal argument exception, as a result of malformed method arguments.
```
dmitry-timofeev: (nit) _e.g._, as a result — because IAE may occur between Java invocations.

exonum-java-binding/core/rust/src/utils/errors.rs
```diff
@@ -176,6 +177,20 @@ pub fn unwrap_exc_or_default<T: Default>(env: &JNIEnv, res: ExceptionResult<T>)
     unwrap_exc_or(env, res, T::default())
 }
 
+/// Returns a cause of the passed exception by using `Throwable#getCause` method.
+///
+/// Panics if `exception` is null.
```
dmitry-timofeev: It is perfectly acceptable for an exception to have `null` cause; only UTEE cannot
have `null` cause, therefore I'd reconsider such behaviour in a seemingly general purpose method.

exonum-java-binding/core/rust/src/utils/jni_cache.rs
```diff
@@ -309,15 +334,21 @@ pub mod classes_refs {
         unsafe { JAVA_LANG_RUNTIME_EXCEPTION.clone().unwrap() }
     }
 
+    /// Returns cached `JClass` for `java/lang/RuntimeException` as a `GlobalRef`.
```
dmitry-timofeev: /wrong exc. name

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -114,30 +118,82 @@ impl JavaRuntimeProxy {
...
     ///
-    /// Any JNI errors are converted into `ExecutionError` with their descriptions, for JNI errors
-    /// like `JniErrorKind::JavaException` it gets (and clears) any exception that is currently
-    /// being thrown, then exception is passed to corresponding `ExceptionHandler` according their
-    /// type and `exception_handlers` mapping.
-    /// `ExceptionHandlers::DEFAULT` is called in case of there is no any handlers or handlers are
-    /// not matched to exception type.
+    /// Used for runtime methods results of which depend on
+    /// implementation of the user service, such as `execute`,
+    /// `before_transactions` and `after_transactions`.
```
dmitry-timofeev: I noticed 'CallSite' has 'constructor' variant:
// Service constructor.
CONSTRUCTOR = 1;

Shall we handle the errors from there in the same way?

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -455,25 +503,36 @@ struct ExceptionHandlers;
...
+
+    const ILLEGAL_ARGUMENT: &'static ExceptionHandler = &|env, exception| {
+        let message = unwrap_jni(get_exception_message(env, exception)).unwrap_or_default();
+        Error::IllegalArgument.with_description(message)
+    };
+}
+
+/// Returns the error code of the `TransactionExecutionException`.
+fn get_tx_error_code(env: &JNIEnv, exception: JObject) -> JniResult<i8> {
+    assert!(!exception.is_null(), "No exception thrown.");
```
dmitry-timofeev: "Exception is null" or sth.

exonum-java-binding/core/rust/src/utils/errors.rs
vitvakatu: Good catch, but it panics when passed exception is null, not its cause is null. So I'd add a check for null cause in the appropriate places, but this particular method seems okay.
Another option might be returning `Option` to represent empty cause, I'd consider that.

exonum-java-binding/core/rust/src/proxy/runtime.rs
vitvakatu: It's interesting. In what methods of the Runtime will the constructor be called? I assume, `initateAddingService`? We can cover this method too, if `TransactionExecutionException`s and/or `UnexpectedTransactionExecutionException`s are possible there.

dmitry-timofeev: Currently, no, I submitted https://jira.bf.local/browse/ECR-4074 , but there are many questions.

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -114,30 +118,82 @@ impl JavaRuntimeProxy {
...
+    /// - Any JNI errors are converted into `ExecutionError::Runtime(JniError)`.
+    /// - `IllegalArgumentException`s are converted into
+    ///   `ExecutionError::Runtime(IllegalArgument)`.
+    /// - `TransactionExecutionException`s are converted into
+    ///   `ExecutionError::Service`.
+    /// - `UnexpectedTransactionExecutionException`s are converted into
+    ///   `ExecutionError::Unexpected`.
+    /// - Any other exceptions are converted into
+    ///   `ExecutionError::Runtime(JavaException)`.
+    fn jni_call_transaction<F, R>(&self, f: F) -> Result<R, ExecutionError>
```
dmitry-timofeev: I'd consider unit-testing these cases in native — they are easier to trigger than in Java ITs.

vitvakatu: Yes, sounds reasonable. Given that native ITs are disabled now, it doesn't make much sense implementing them in this PR. I'll create an issue with testing scenarios.

exonum-java-binding/core/rust/src/utils/errors.rs
dmitry-timofeev: I missed that, sorry. 

> for null cause 

I don't think it is needed much because UTEE _enforces_ that in its ctor.

exonum-java-binding/core/rust/src/proxy/runtime.rs
dmitry-timofeev: Do we have an issue to fix existing ITs? I'd abstain from just piling issues and make the ITs pass to be able to add new tests promptly.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/UnexpectedTransactionExecutionException.java
```diff
@@ -41,7 +41,9 @@ public UnexpectedTransactionExecutionException(Throwable cause) {
 
   private static Throwable checkValidCause(Throwable cause) {
     checkNotNull(cause, "null cause is not allowed");
-    checkArgument(!(cause instanceof TransactionExecutionException));
+    // FIXME: revert this change
+    //checkArgument(!(cause instanceof TransactionExecutionException));
+    checkArgument(!(cause instanceof TransactionExecutionException), cause.getClass().toString());
```
vitvakatu: UTEE test fails with IllegalArgumentException thrown at this line. I added class name to debug it, and it prints `"class com.exonum.binding.core.transaction.TransactionExecutionException"`.

@dmitry-timofeev @bullet-tooth any ideas on how is it possible?

vitvakatu: 🤔 `!` looks redundant here, isn't it?

vitvakatu: Nevermind, now I understand the problem 😄

vitvakatu:
To reviewers: methods from proxy/runtime module are moved to runtime/error module without any semantic changes.

exonum-java-binding/core/rust/integration_tests/src/vm.rs
```diff
@@ -85,7 +81,7 @@ fn create_vm(debug: bool, with_fakes: bool) -> JavaVM {
     let vm = JavaVM::new(jvm_args).unwrap_or_else(|e| panic!("{:#?}", e));
 
     // Initialize JNI cache for testing with fakes
```
dmitry-timofeev: Not fakes

exonum-java-binding/core/rust/integration_tests/src/vm.rs
```diff
@@ -31,41 +31,37 @@ const CONVERSION_FAILED_MESSAGE: &str = "Failed to convert FS path into utf-8";
...
 }
 
 /// Creates a configured `JavaVM`.
 /// _JavaVM should be created only *once*._
-fn create_vm(debug: bool, with_fakes: bool) -> JavaVM {
+///
+/// If `debug` is true, additional checks of correctness of JNI operations are
+/// enabled.
+///
+/// If `with_classes` is true, Java classes, dependencies of the EJB App are
```
dmitry-timofeev: I think it's fine now, but ideally these ITs must use core classpath so that they can be launched right after building core, and not after app. They do not depend on Rust app.

exonum-java-binding/core/rust/integration_tests/tests/runtime_error_handling.rs
```diff
@@ -0,0 +1,194 @@
...
+        env.throw(JThrowable::from(exception))?;
+        wait_for_thrown_exception(env)?;
+        Ok(())
+    })
+    .unwrap_err();
+    assert_eq!(err.kind(), ErrorKind::Unexpected);
+    assert_eq!(err.description(), EXPECTED_DESCRIPTION);
+}
+
+fn create_transaction_execution_exception<'a>(env: &'a JNIEnv) -> JniResult<JObject<'a>> {
```
dmitry-timofeev: There is no longer a transaction execution exception (abe31e590bf43df41a536609381f8eed126a3d12).

The method under test and tests might need to be renamed. However, we may still
use 'transaction' to distinguish it from 'default'.

exonum-java-binding/core/rust/integration_tests/tests/runtime_error_handling.rs
```diff
@@ -0,0 +1,194 @@
...
+where
+    R: std::fmt::Debug,
+{
+    assert!(result.is_err());
+    let err = result.unwrap_err();
+    assert_eq!(err, expected_error);
+}
+
+/// Performs some trivial action using JNI to wait for Java exception being thrown.
+fn wait_for_thrown_exception<'a>(env: &'a JNIEnv) -> JniResult<JObject<'a>> {
```
dmitry-timofeev: It does not wait for exception (it is already thrown after successful #throw), it invokes an exception-unsafe operation to trigger a JNI error of type 'Exception' which is caused by a previously thrown exception. 

I'd clarify the docs:
> Performs an exception-unsafe JNI operation to trigger a JNI error caused by an already thrown Java exception.

Also, I'd check the precondition that there is an exception thrown (assert(env.exception_check)).

exonum-java-binding/core/rust/integration_tests/src/vm.rs
vitvakatu: ECR-4094 created

exonum-java-binding/integration-tests/src/test/java/com/exonum/binding/test/ServiceRuntimeIntegrationTest.java
```diff
@@ -46,7 +46,7 @@
 
   private static final byte AFTER_TX_ERROR_CODE = 17;
 
-  @Disabled("ECR-4054")
+  @Disabled("ECR-4093")
```
vitvakatu: The test is not working still, see ECR-4093

dmitry-timofeev:
It is still unclear what happens with errors in 'initialize' (we have an incorrect documentation stating it is saved in call errors, but there is no corresponding CallInBlock variant): https://jira.bf.local/browse/ECR-4098

PR #1355, opened 2020-01-10
Add block and index proofs [ECR-4011]

dmitry-timofeev:
## Overview

Added block and index proofs support. They shall be created through Blockchain.

Documented the various proof types: what they prove, what comprises them, how to create them. Added an example for a commonly used proof type.

Also:
- Made SignedMessage public to be able to use it in tests; got rid of redundant serialization in #parseFrom.
- Added Block#parseFrom and Block#fromMessage.

---
See: ECR-4011

---
⚠️ **There are some questions marked as TODO (@slowli , @bullet-tooth , @vitvakatu ).**

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/core/src/main/java/com/exonum/binding/core/blockchain/Blockchain.java
```diff
@@ -43,14 +47,108 @@
  * blockchain::Schema</a> features in the Core API: blocks, transaction messages, execution
  * results.
  *
+ * <!-- This section is suppossed to be the main Javadoc on proofs, documenting how
+ *      to create various blockchain proofs from their components.
+ *
+ *      Link here with <a href="<relative path>/Blockchain.html#proofs">Blockchain Proofs</a>.
+ *      See also: https://stackoverflow.com/a/27522316/ -->
+ * <h2 id="proofs">Proofs</h2>
```
dmitry-timofeev: ⚠️ ECR-4106

slowli: The documentation here looks good 👍

exonum-java-binding/core/src/main/java/com/exonum/binding/core/blockchain/Blockchain.java
```diff
@@ -59,7 +157,46 @@
...
+   * @throws IllegalStateException if the view is not a snapshot, because a state of a service index
+   *     can be proved only for the latest committed block, not for any intermediate state during
+   *     transaction processing
+   */
+  public IndexProof createIndexProof(String fullIndexName) {
+    /*
+     todo: Shall we allow creating proofs for invalid (e.g., impossible) index names or throw
+      an exception?
+
+     todo: If index proofs for "uninitialized" indexes are forbidden, document that.
```
vitvakatu: The native will throw an exception, so I think we should document it.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/blockchain/BlockchainProofs.java
```diff
@@ -0,0 +1,71 @@
...
+    LibraryLoader.load();
+  }
+
+  /**
+   * Creates a block proof for the block at the given height.
+   * @param view a database view
+   * @param height the height of the block
+   */
+  static BlockProof createBlockProof(
+      /* todo: here snapshot is not strictly required — but shall we allow Forks (see the ticket) */
```
vitvakatu: Actually I couldn't find any use case for block proofs in execution context (where fork is available), so I propose to keep such restriction for now.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/blockchain/Blockchain.java
dmitry-timofeev: [But](https://github.com/exonum/exonum-java-binding/commit/fab544b8358bac6b1d201948e9f4782897cfd1e7#r36744088)

> Why doesn't it return a proof of absence (no such index)? Especially as an index is not "initialized" unless (?) something is recorded in it? Doesn't that place extra burden on the service developers to "initialize" an index? Shall they now write an extra "initialize" method in a schema of each service (and we — document that)?

vitvakatu: Actually, I don't know. I read sources of `proof_for_index` method - it looks like explicit check for existence of the index can be removed, and the method will return proof of absence. @slowli am I right?

exonum-java-binding/core/src/main/java/com/exonum/binding/core/blockchain/proofs/IndexProof.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Proof of authenticity for a single index in the database.
+ *
+ * <p>It is comprised of a {@link BlockProof} and a {@link MapProof} from the collection
+ * aggregating the index hashes of proof indexes for an index with a certain full name.
+ *
+ * <p>If an index does not exist in the database, then the MapProof will prove its absence.
+ *
+ * @see <a href="../Blockchain.html#service-data-proof">Service Data Proofs</a>
```
bullet-tooth: IDEA's quick documentation correctly opens the referenced javadoc, but doesn't recognize anchors. I think it's not the issue for changing links, just fyi

dmitry-timofeev: Thanks! That's weird, it works in mine:

зображення

dmitry-timofeev: Don't anchors work anywhere, even in Blockchain#callerrors?

bullet-tooth: yeap, Blockchain#callerrors doesn't work as well 🤷‍♀️

image

exonum-java-binding/core/src/main/java/com/exonum/binding/core/blockchain/Blockchain.java
slowli: Yes, if the check of existence is removed, `proof_for_index` may return proof of absence for a non-existing index. The reason the method is implemented as is currently is that returning proof for a non-existing index doesn't make sense in a typical use case (returning the index proof + proof for data *within* the index, such as proof for a wallet in the cryptocurrency service).

dmitry-timofeev: Thanks! Would you clarify please, if one first requests a proof from an empty index (= no wallets), will it become "initialized"? Will one be able to provide then a proof that the index remains empty?

dmitry-timofeev:
We briefly discussed Optional vs Exception in case of uninitialized indexes in `createIndexProof` with @vitvakatu , and decided to go with an Exception, because eventually we would like to have them all initialized automatically and not be a burden to service developers: ECR-4121.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Schema.java
```diff
@@ -27,8 +28,11 @@
  * the core automatically tracks every Merkelized collection used by the user
```
bullet-tooth:
```suggestion
 * Exonum automatically tracks every Merkelized collection used by the user
```
not sure that we use "core" term in public docs.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Schema.java
```diff
@@ -27,8 +28,11 @@
...
  * services. It aggregates state hashes of these collections into a single
  * Merkelized meta-map. The hash of this meta-map is considered the hash of the
- * entire blockchain state and is recorded as such in blocks and Precommit
- * messages.
+ * entire blockchain state and is recorded as such in {@linkplain Block#getStateHash() blocks}
+ * and Precommit messages.
+ *
+ * <p>Exonum starts aggregating a service collection state hash once it is <em>initialized</em>:
```
bullet-tooth:
```suggestion
 * <p>Exonum starts aggregating a service collections state hash once it is <em>initialized</em>:
```

dmitry-timofeev: The sentence says when Exonum starts aggregating one (any, a single) collection.

PR #1361, opened 2020-01-13
Support different supervisor modes in exonum-java

vitvakatu:
## Overview

---

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/rust/exonum-java/src/node.rs
+++ b/exonum-java-binding/core/rust/exonum-java/src/node.rs
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-use exonum_supervisor::Supervisor;
+use exonum_supervisor::{mode::Mode as SupervisorMode, Supervisor};
 use exonum_time::TimeServiceFactory;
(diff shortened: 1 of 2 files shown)


exonum-java-binding/core/rust/exonum-java/src/node.rs
```diff
@@ -70,7 +70,13 @@ fn create_blockchain(
 
     let blockchain = Blockchain::new(database, keypair, api_sender);
 
-    let supervisor_service = supervisor_service();
+    let supervisor_mode = &config
+        .run_config
+        .node_config
```
dmitry-timofeev: Why don't you take node_config from above? And, possibly, move the remaining extraction code inside supervisor_service?

vitvakatu: Because `node_config` from above isn't `config.run_config.node_config`. It's exonum core's internal node configuration which does not have supervisor mode option. ECR-4116 created to address API inconvenience that forced this.

PR #1358, opened 2020-01-13
[ECR-4104] Allow services stopping

bullet-tooth:
## Overview

---
See: https://jira.bf.local/browse/ECR-4104

### Definition of Done

- [ ] There are no TODOs left in the code
- [ ] Change is covered by automated [tests](https://github.com/exonum/exonum-java-binding/blob/master/CONTRIBUTING.md#tests)
- [ ] The [coding guidelines](https://github.com/exonum/exonum-java-binding/blob/master/CONTRIBUTING.md#the-code-style) are followed
- [ ] Public API has Javadoc
- [ ] Method preconditions are checked and documented in the Javadoc of the method
- [ ] Changelog is updated if needed (in case of notable or breaking changes)
- [ ] The continuous integration build passes

PR #1365, opened 2020-01-14
[ECR-2334]: Using different Node instances in services

bullet-tooth:
## Overview

---
See: https://jira.bf.local/browse/ECR-2334

- Now each service receives a decorated copy of the node proxy
- Native node resources sweep happens at ServiceRuntime#shutdown

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/RestrictingNodeDecorator.java
+++ b/exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/RestrictingNodeDecorator.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2020 The Exonum Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.exonum.binding.core.runtime;
+
+import static com.google.common.base.Preconditions.checkState;
+
+import com.exonum.binding.common.crypto.PublicKey;
+import com.exonum.binding.common.hash.HashCode;
+import com.exonum.binding.core.service.Node;
+import com.exonum.binding.core.storage.database.Snapshot;
+import com.exonum.binding.core.transaction.RawTransaction;
+import java.util.function.Function;
+
+/**
+ * Node decorator which can restrict access to the node by calling {@link #restrictAccess()} method.
+ */
+class RestrictingNodeDecorator implements Node {
+
+  private final Node node;
+  private boolean accessAllowed;
+
+  RestrictingNodeDecorator(Node node) {
+    this.node = node;
+    this.accessAllowed = true;
+  }
+
+  @Override
+  public HashCode submitTransaction(RawTransaction rawTransaction) {
+    checkAccess();
+    return node.submitTransaction(rawTransaction);
+  }
+
+  @Override
+  public <ResultT> ResultT withSnapshot(Function<Snapshot, ResultT> snapshotFunction) {
+    checkAccess();
+    return node.withSnapshot(snapshotFunction);
+  }
+
+  @Override
+  public PublicKey getPublicKey() {
+    return node.getPublicKey();
+  }
+
+  /**
+   * Restricts access to the node. After calling this method subsequent calling
+   * {@link #submitTransaction(RawTransaction)} or {@link #withSnapshot(Function)} methods
+   * will cause {@link IllegalStateException}.
+   */
+  void restrictAccess() {
+    this.accessAllowed = false;
+  }
+
+  private void checkAccess() {
+    checkState(accessAllowed, "Node access is not allowed");
+  }
+
+}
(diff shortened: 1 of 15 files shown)


bullet-tooth:
This PR is based on #1358, commit to reviewing: `8dd8dc9`

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/RestrictingNodeDecorator.java
```diff
@@ -0,0 +1,71 @@
...
+import com.exonum.binding.common.hash.HashCode;
+import com.exonum.binding.core.service.Node;
+import com.exonum.binding.core.storage.database.Snapshot;
+import com.exonum.binding.core.transaction.RawTransaction;
+import java.util.function.Function;
+
+/**
+ * Node decorator which can restrict access to the node by calling {@link #restrictAccess()} method.
+ */
+class RestrictingNodeDecorator implements Node {
```
dmitry-timofeev: This class is a Node decorator, hence must adhere to its specification. Currently,
a Node is still not AutoCloseable/CloseableNativeProxy, that:
 * makes this decorator break the Node specification, by throwing unspecified exceptions
 * requires all the places that previously used Node-as-interface to use particular types:
   NodeProxy or RestrictingNodeDecorator (some even perform type casts) — ServiceRuntime, ServiceWrapper, their tests.

I suggest (as we have already briefly discussed) to redefine the Node *specification* to allow it to be _#closed_ (probably, it must be a more generic AutoCloseable, because CloseableNativeProxy is more specific), and use #close to:
 * Destroy the NodeProxy (in ServiceRuntime)
 * Prevent further access to service-specific RestrictingNodeDecorator (in ServiceWrapper)
but all — on Node.

dmitry-timofeev: I'd ponder on a name that communicates that this decorator does not close the Node
it decorates:
 - NonOwningNodeDecorator (it does not _own_ the decorated Node, hence, does not destroy it)?
 - NonDestroyingNodeDecorator?
 - MultiplexingNodeDecorator?

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/RestrictingNodeDecoratorTest.java
```diff
@@ -0,0 +1,85 @@
...
+    decorator.restrictAccess();
+
+    assertThrows(IllegalStateException.class, () -> decorator.withSnapshot(SNAPSHOT_FUNCTION));
+  }
+
+  @Test
+  void getPublicKey() {
+    when(node.getPublicKey()).thenReturn(PublicKey.fromHexString("ab"));
+
+    decorator.getPublicKey();
```
dmitry-timofeev: assertEquals instead of verify — currently it does not check the actual return value.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
```diff
@@ -258,20 +258,27 @@ private void activateService(ServiceInstanceSpec instanceSpec) {
...
+      unRegisterService(service);
       logger.info("Stopped a service: {}", instanceSpec);
     } else {
       logger.warn("There is no active service with the given name {}. "
           + "Possibly restoring services state after reboot?", name);
     }
   }
 
+  private void forbidAccessToBlockchain(ServiceWrapper service) {
+    Node serviceNode = service.getNode();
```
dmitry-timofeev: ServiceWrapper exists to provide actions on services as a whole. Requesting a service to stop its operation (currently — to close its Node) is such operation. I'd move this to ServiceWrapper as #close or, better, #requestToStop.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceWrapper.java
```diff
@@ -209,4 +209,9 @@ String getPublicApiRelativePath() {
     return UrlEscapers.urlPathSegmentEscaper()
         .escape(getName());
   }
+
+  Node getNode() {
```
dmitry-timofeev: I'd suggest #close/requestToStop instead of exposing the Node — see comment in SR.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/RestrictingNodeDecorator.java
dmitry-timofeev: The node currently gets passed to `createPublicApiHandlers`, we could mention there briefly that it gets closed when the service is stopped.

dmitry-timofeev:
Pushed: 6786c582...0d4e0137

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Node.java
```diff
@@ -26,8 +26,13 @@
 /**
  * An Exonum node context. Allows to add transactions to Exonum network
  * and get a snapshot of the database state.
+ *
+ * <p>Once the node is <em>closed</em>, it can no longer be used. Methods of a closed Node
+ * will throw an {@link IllegalStateException} if invoked.
```
bullet-tooth: do we need to throw ISE at `getPublicKey`?

dmitry-timofeev: We can add it, but I think we can get away with no Exception)

dmitry-timofeev: To do that in a concise way, we could use node() method to access the node (not a field load) and put the check there.

PR #1366, opened 2020-01-16
Extract #getIndexHash in HashableIndex interface [ECR-4064].

dmitry-timofeev:
## Overview

HashableIndex is similar to ObjectHash trait in MerkleDB.

---
See: https://jira.bf.local/browse/ECR-4064

### Definition of Done

- [ ] Public API has Javadoc


exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Schema.java
```diff
@@ -38,9 +35,7 @@
  * the framework will not be able to verify that its transactions cause the same
  * results on different nodes.
  *
- * @see ProofListIndexProxy#getIndexHash()
- * @see ProofMapIndexProxy#getIndexHash()
- * @see ProofEntryIndexProxy#getIndexHash()
+ * @see com.exonum.binding.core.storage.indices.HashableIndex
```
dmitry-timofeev: The other links are index-specific.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/ProofMapIndexProxy.java
```diff
@@ -85,7 +85,8 @@
  * @param <V> the type of values in this map
  * @see View
  */
-public final class ProofMapIndexProxy<K, V> extends AbstractIndexProxy implements MapIndex<K, V> {
+public final class ProofMapIndexProxy<K, V> extends AbstractIndexProxy implements MapIndex<K, V>,
```
dmitry-timofeev: When we make the index factories and add the interfaces for each index type, those interfaces could extend HashableIndex, instead of implementation classes.

PR #1367, opened 2020-01-17
Pass instance status as bytes array

bullet-tooth:
## Overview

---
See: https://jira.bf.local/browse/ECR-4119

### Definition of Done

- [ ] Public API has Javadoc


exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntimeAdapter.java
```diff
@@ -156,14 +156,13 @@ void initiateAddingService(long forkHandle, byte[] instanceSpec, byte[] configur
    *
    * @param instanceSpec the service instance specification as a serialized {@link InstanceSpec}
    *     protobuf message
-   * @param numericInstanceStatus new status of the service instance as a numeric
-   *     representation of the {@link InstanceState.Status} enum.
+   * @param instanceState new state of the service instance as a serialized
+   *     representation of the {@link InstanceState} protobuf message
```
vitvakatu: It's invalid, `InstanceState.Status` must be used.

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/ServiceRuntimeConfigurationIntegrationTest.java
```diff
@@ -81,9 +82,10 @@ void runtimeConfigurationTest(@TempDir Path tmpArtifactDir) throws Exception {
       // Initialize and register a service instance
       String name = "s1";
       ServiceInstanceSpec instanceSpec = ServiceInstanceSpec.newInstance(name, 1, ARTIFACT_ID);
+      InstanceStatus instanceStatus = InstanceStatus.newBuilder().setSimple(Simple.NONE).build();
       Fork fork = database.createFork(cleaner);
       runtime.initiateAddingService(fork, instanceSpec, new byte[0]);
-      runtime.updateInstanceStatus(instanceSpec, InstanceState.Status.NONE);
+      runtime.updateInstanceStatus(instanceSpec, instanceStatus);
```
dmitry-timofeev: Why NONE? That must be an artifact between the Exonum native update and stopping support — it must be active, I believe.

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/ServiceRuntimeIntegrationTest.java
```diff
@@ -67,6 +68,8 @@
   static final int TEST_ID = 17;
   static final HashCode TEST_HASH = HashCode.fromBytes(bytes(1, 2, 3));
   static final PublicKey TEST_PUBLIC_KEY = PublicKey.fromBytes(bytes(4, 5, 6));
+  static final InstanceStatus TEST_STATUS = InstanceStatus.newBuilder()
+      .setSimple(Simple.NONE).build();
```
dmitry-timofeev: There is no NONE status.

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/ServiceRuntimeIntegrationTest.java
```diff
@@ -258,7 +261,7 @@ void commitService() {
         .thenReturn(serviceWrapper);
 
     // Create the service from the artifact
-    serviceRuntime.updateInstanceStatus(instanceSpec, InstanceState.Status.NONE);
+    serviceRuntime.updateInstanceStatus(instanceSpec, TEST_STATUS);
```
dmitry-timofeev: Again, same artifact; must not be NONE, but appropriate status.

dmitry-timofeev:
We can keep the updated code, if the native is also updated: https://jira.bf.local/browse/ECR-4130 (I see 3773 is integrated).

vitvakatu:
I'll update native part. Seems like it is necessary due to java protobuf capabilities (it can't deserialize internal enum)

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/ServiceRuntimeConfigurationIntegrationTest.java
bullet-tooth: yeap, I'll change during merge with stopping services

bullet-tooth:
> @bullet-tooth Proto files base revision must match the revision we use. This PR does not seem to need the updated core version, hence, the new proto files.

I see no reason to change the type without updating proto files

dmitry-timofeev:
> I see no reason to change the type without updating proto files 

There is — it is documented in Jira (integer serialization of the enum in native is not reliable). However, as we've established, byte[] -> inner proto enum does not work.

exonum-java-binding/core/rust/exonum-java/src/node.rs
```diff
@@ -45,7 +45,8 @@ pub fn run_node(command: Command) -> Result<(), failure::Error> {
 
 fn create_node(config: Config) -> Result<Node, failure::Error> {
     let node_config = config.run_config.node_config.clone();
-    let events_pool_capacity = &node_config.private_config.mempool.events_pool_capacity;
+    // TODO: rewrite with NodeBuilder API when it is available
+    let events_pool_capacity = &Default::default();
```
vitvakatu: Unfortunately, we can't use events_pool_capacity from user's config in current Core version, but it is addressed in the latest Core revision.

exonum-java-binding/core/rust/src/proxy/node.rs
```diff
@@ -66,7 +67,10 @@ impl Node {
         let verified = Verified::from_value(tx, pub_key.to_owned(), secret_key);
         let tx_hash = verified.object_hash();
         // TODO(ECR-3679): check Core behaviour/any errors on service inactivity
-        self.blockchain.sender().broadcast_transaction(verified)?;
+        self.blockchain
+            .sender()
+            .broadcast_transaction(verified)
+            .wait()?;
```
vitvakatu: `broadcast_transaction` now returns a Future! However, the actual implementation is unchanged - it used to call `wait` in the Core code, now we call it ourselves.

PR #1372, opened 2020-01-21
[ECR-4133] Add service resume

bullet-tooth:
## Overview

---
See: https://jira.bf.local/browse/ECR-4133

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/CHANGELOG.md
+++ b/exonum-java-binding/CHANGELOG.md
@@ -37,8 +37,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   allows to configure the mode of the Supervisor service. Possible values are
   "simple" and "decentralized". (#1361)
 - Support of service instances lifecycle: they can be activated, stopped and resumed now.
-  Also, service instance artifacts can be upgraded before resuming (but with no data changes)
-  which allows services API update, add new service transactions etc. (#1358, #1372)
+  Also, service instance artifacts can be upgraded before resuming which allows services
+  API update, add new service transactions, synchronous data migration etc. (#1358, #1372)
 
 [blockchain-proofs]: https://exonum.com/doc/api/java-binding/0.10.0-SNAPSHOT/com/exonum/binding/core/blockchain/Blockchain.html#proofs
 
--- a/exonum-java-binding/CHANGELOG.md
+++ b/exonum-java-binding/CHANGELOG.md
@@ -36,7 +36,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `supervisor-mode` CLI parameter added for `generate-template` command. It
   allows to configure the mode of the Supervisor service. Possible values are
   "simple" and "decentralized". (#1361)
-- Service instances can be stopped and resumed now. (#1358, #1372)
+- Support of service instances lifecycle: they can be activated, stopped and resumed now.
+  Also, service instance artifacts can be upgraded before resuming (but with no data changes)
+  which allows services API update, add new service transactions etc. (#1358, #1372)
 
 [blockchain-proofs]: https://exonum.com/doc/api/java-binding/0.10.0-SNAPSHOT/com/exonum/binding/core/blockchain/Blockchain.html#proofs
 
--- a/exonum-java-binding/CHANGELOG.md
+++ b/exonum-java-binding/CHANGELOG.md
@@ -36,7 +36,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `supervisor-mode` CLI parameter added for `generate-template` command. It
   allows to configure the mode of the Supervisor service. Possible values are
   "simple" and "decentralized". (#1361)
-- Service instances can be stopped now. (#1358)
+- Service instances can be stopped and resumed now. (#1358, #1372)
 
 [blockchain-proofs]: https://exonum.com/doc/api/java-binding/0.10.0-SNAPSHOT/com/exonum/binding/core/blockchain/Blockchain.html#proofs
 
--- a/exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
+++ b/exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
@@ -228,7 +228,7 @@ public final class ServiceRuntime implements AutoCloseable {
       synchronized (lock) {
         checkStoppedService(instanceSpec.getId());
         ServiceWrapper service = createServiceInstance(instanceSpec);
-        service.resume(fork, new ServiceConfiguration(arguments));
+        service.resume(fork, arguments);
       }
       logger.info("Resumed service: {}", instanceSpec);
     } catch (Exception e) {
--- a/exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
+++ b/exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
@@ -513,7 +513,7 @@ public final class ServiceRuntime implements AutoCloseable {
         "No service with id=%s in the Java runtime", serviceId);
   }
 
-  /** Checks that the service with the given id is started in this runtime. */
+  /** Checks that the service with the given id is not active in this runtime. */
   private void checkStoppedService(Integer serviceId) {
     checkArgument(!servicesById.containsKey(serviceId),
         "Service with id=%s should be stopped, but actually active", serviceId);
--- a/exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
+++ b/exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
@@ -210,10 +210,9 @@ public final class ServiceRuntime implements AutoCloseable {
    * Initiates resuming of previously stopped service instance. Service instance artifact could
    * be upgraded in advance to bring some new functionality.
    *
-   * @param fork a database view to apply configuration
+   * @param fork a database view to apply arguments
    * @param instanceSpec a service instance specification; must reference a deployed artifact
-   * @param configuration a service instance configuration parameters as a serialized protobuf
-   *     message
+   * @param arguments a service instance arguments as a serialized protobuf message
    * @throws IllegalArgumentException if the given service instance is active; or its artifact
    *     is not deployed
    * @throws ExecutionException if such exception occurred in the service method;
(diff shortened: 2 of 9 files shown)


exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntimeAdapter.java
```diff
@@ -151,6 +151,20 @@ void initiateAddingService(long forkHandle, byte[] instanceSpec, byte[] configur
...
+  /**
+   * Starts resuming a service with the given specification.
+   *
+   * @param instanceSpec the service instance specification as a serialized {@link InstanceSpec}
+   *     protobuf message
+   * @param configuration the service initial configuration parameters as a serialized protobuf
+   *     message
+   * @see ServiceRuntime#initializeResumingService(ServiceInstanceSpec, byte[])
+   */
+  void initializeResumingService(byte[] instanceSpec, byte[] configuration) {
```
bullet-tooth: @vitvakatu method signature for ECR-4134

exonum-java-binding/CHANGELOG.md
```diff
@@ -36,7 +36,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `supervisor-mode` CLI parameter added for `generate-template` command. It
   allows to configure the mode of the Supervisor service. Possible values are
   "simple" and "decentralized". (#1361)
-- Service instances can be stopped now. (#1358)
+- Service instances can be stopped and resumed now. (#1358, #1372)
```
dmitry-timofeev: An instance artifact may be upgraded before resuming, but with no data changes. 

(may also insert the uses cases from the epic/ECR-3773).

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Service.java
```diff
@@ -53,6 +53,24 @@ default void initialize(Fork fork, Configuration configuration) {
     // No configuration
   }
 
+  /**
+   * Performs resuming of previously stopped service instance. This method is called <em>once</em>
```
dmitry-timofeev:
```
   * Resumes the previously stopped service instance. This method is called when
   * a stopped service instance is restarted.
```

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Service.java
```diff
@@ -53,6 +53,24 @@ default void initialize(Fork fork, Configuration configuration) {
...
+   * due to violation migration procedure.</em>
+   *
+   * @param configuration the service configuration parameters
+   * @throws ExecutionException if the configuration parameters are not valid (e.g.,
+   *     malformed, or do not meet the preconditions). Exonum will save the error into
+   *     {@linkplain com.exonum.binding.core.blockchain.Blockchain#getCallErrors(long)
+   *     the registry of call errors} if the resuming fails
+   * @see Configurable
+   */
+  default void resume(Configuration configuration) {
```
dmitry-timofeev: @slowli, @bullet-tooth  In which cases the service needs to be notified of that, if it can't do anything with this configuration, just look at them? I remember that the service instance _objects_ are not supposed to have much state, therefore,

> "As an example, arguments can be used to update the service configuration."

does not seem to make much sense — that configuration must be normally kept in the persistent storage. On top of that, we do *not* provide a similar operation for _node restarts_, when a fresh service instance object is also created.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Service.java
```diff
@@ -53,6 +53,24 @@ default void initialize(Fork fork, Configuration configuration) {
...
+  /**
+   * Performs resuming of previously stopped service instance. This method is called <em>once</em>
+   * after restarting a service instance.
+   *
+   * <p><em>Note that changing the data during resume operations is not allowed
+   * due to violation migration procedure.</em>
+   *
+   * @param configuration the service configuration parameters
+   * @throws ExecutionException if the configuration parameters are not valid (e.g.,
+   *     malformed, or do not meet the preconditions). Exonum will save the error into
```
dmitry-timofeev: It will not save them.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Service.java
```diff
@@ -53,6 +53,24 @@ default void initialize(Fork fork, Configuration configuration) {
...
+   *     {@linkplain com.exonum.binding.core.blockchain.Blockchain#getCallErrors(long)
+   *     the registry of call errors} if the resuming fails
+   * @see Configurable
```
dmitry-timofeev: Configurable is not related to this, but initialize (if we decide to keep the method at all)

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
```diff
@@ -206,6 +206,23 @@ public void initiateAddingService(Fork fork, ServiceInstanceSpec instanceSpec,
...
+   *     message
+   * @throws IllegalArgumentException if the given service instance is active
+   */
+  public void initializeResumingService(ServiceInstanceSpec instanceSpec,
+      byte[] configuration) {
+    synchronized (lock) {
+      checkStoppedService(instanceSpec.getId());
+      ServiceWrapper service = createServiceInstance(instanceSpec);
+      service.resume(new ServiceConfiguration(configuration));
+    }
```
dmitry-timofeev: Currently this class is responsible to log errors, though there is ECR-3813 to shift that.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
```diff
@@ -486,16 +503,22 @@ public void close() throws InterruptedException {
...
-  private void checkService(Integer serviceId) {
+  private void checkActiveService(Integer serviceId) {
     checkArgument(servicesById.containsKey(serviceId),
         "No service with id=%s in the Java runtime", serviceId);
   }
 
+  /** Checks that the service with the given id is not active in this runtime. */
+  private void checkStoppedService(Integer serviceId) {
+    checkArgument(!servicesById.containsKey(serviceId),
+        "Service with id=%s should be stopped, but actually active", serviceId);
```
dmitry-timofeev: Add the offending instance?

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/ServiceRuntimeIntegrationTest.java
```diff
@@ -96,8 +96,6 @@ void startsServerOnInitialization() {
 
   @Test
   void deployCorrectArtifact() throws Exception {
-    serviceRuntime.initialize(mock(Node.class));
```
dmitry-timofeev: (Here and elsewhere) Why is this removed? No operation shall occur before SR is initialized.

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/ServiceRuntimeIntegrationTest.java
```diff
@@ -191,9 +183,61 @@ void startAddingService() {
...
+    serviceRuntime.initializeResumingService(instanceSpec, configuration);
+
+    // Check it was instantiated as expected
+    verify(servicesFactory).createService(eq(serviceDefinition), eq(instanceSpec),
+        any(MultiplexingNodeDecorator.class));
+
+    // and the service was configured
+    Configuration expectedConfig = new ServiceConfiguration(configuration);
+    verify(serviceWrapper).resume(expectedConfig);
+  }
```
dmitry-timofeev: // but not registered in the runtime yet:
assertThat(serviceRuntime.findService(TEST_NAME)).isEmpty();

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/ServiceRuntimeIntegrationTest.java
bullet-tooth: ok, I will add a null check to the decorator then

dmitry-timofeev: Not sure I understand — the SR expects that it is first initialized with a non-null Node before any other operations are performed.

exonum-java-binding/CHANGELOG.md
```diff
@@ -36,7 +36,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `supervisor-mode` CLI parameter added for `generate-template` command. It
   allows to configure the mode of the Supervisor service. Possible values are
   "simple" and "decentralized". (#1361)
-- Service instances can be stopped now. (#1358)
+- Support of service instances lifecycle: they can be activated, stopped and resumed now.
+  Also, service instance artifacts can be upgraded before resuming (but with no data changes)
```
dmitry-timofeev: No data changes, according to the recent discussions, is no longer correct.

I'd add to the list below 'synchronous data migration'.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Service.java
```diff
@@ -53,6 +53,26 @@ default void initialize(Fork fork, Configuration configuration) {
...
 
+  /**
+   * Resumes the previously stopped service instance. This method is called when
+   * a stopped service instance is restarted.
+   *
+   * <p>This method may perform any changes to the database. For example, update some service
+   * parameters, deprecate old entries etc.
+   *
+   * <p>Also, note that performing any bulk operations or data migration
+   * <em>is not recommended</em> here.
```
dmitry-timofeev: I'd add short motivation, as Ostrovski clarified in the discussion.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Service.java
```diff
@@ -53,6 +53,26 @@ default void initialize(Fork fork, Configuration configuration) {
...
+   * <!--TODO: Add a link to the migration procedure -->
+   *
+   * @param fork a database fork to apply changes to. Not valid after this method returns
+   * @param configuration the service configuration parameters
```
dmitry-timofeev: As we've discussed, those are hardly _configuration parameters_, just generic 'arguments'.

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/ServiceRuntimeIntegrationTest.java
```diff
@@ -369,6 +367,68 @@ void updateServiceStatusBadStatus(Status badStatus) {
...
+    // Create the service from the artifact
+    Fork fork = mock(Fork.class);
+    byte[] configuration = anyConfiguration();
+    serviceRuntime.initializeResumingService(fork, instanceSpec, configuration);
+
+    // Check it was instantiated as expected
+    verify(servicesFactory).createService(eq(serviceDefinition), eq(instanceSpec),
+        any(MultiplexingNodeDecorator.class));
+
+    // and the service was configured
```
dmitry-timofeev: resumed

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Service.java
dmitry-timofeev: `byte[]` since is it is not configuration. We shall not use Configuration for this operation — it is for configuration parameters.

It could be made one, it is made in this way probably to give more flexibility in terms of implementation (e.g., if we went with also properties).

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
```diff
@@ -206,6 +206,38 @@ public void initiateAddingService(Fork fork, ServiceInstanceSpec instanceSpec,
     }
   }
 
+  /**
+   * Initiates resuming of previously stopped service instance. Service instance artifact could
+   * be upgraded in advance to bring some new functionality.
+   *
+   * @param fork a database view to apply arguments
```
dmitry-timofeev: _Arguments_ are not applied to a database.

PR #1373, opened 2020-01-21
Aggregate updates

dmitry-timofeev:
## Overview

Also, perform some adjustments — see individual commits.

### DIFF
--- a/exonum-java-binding/core/pom.xml
+++ b/exonum-java-binding/core/pom.xml
@@ -100,7 +100,13 @@
 
     <dependency>
       <groupId>com.ryanharter.auto.value</groupId>
-      <artifactId>auto-value-gson</artifactId>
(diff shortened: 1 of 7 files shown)


exonum-java-binding/core/pom.xml
```diff
@@ -100,7 +100,13 @@
 
     <dependency>
       <groupId>com.ryanharter.auto.value</groupId>
-      <artifactId>auto-value-gson</artifactId>
+      <artifactId>auto-value-gson-extension</artifactId>
+      <scope>provided</scope>
```
bullet-tooth: as far as I remember `scope` inherits from the parent dependency management. i.e. is not required to be defined explicitly

dmitry-timofeev: 👍 
I'd probably keep it for higher clarity.

PR #1379, opened 2020-01-24
Implement experimental support for Migration, Scratchpad and Prefixed [ECR-4120]

vitvakatu:
## Overview

Implemented experimental support for Migration, Scratchpad and Prefixed types:
1. New type `EjbAccess` is introduced. It is supposed to be used where `View` was used before, in particular this type is send to Java and back.
2. Migration, Scratchpad and Prefixed are created in-place right before usage. It allows to get rid of their lifetimes during ffi calls.
  The typical workflow looks like this:
  - Java calls nativeCreateMigration(String namespace).
  - Rust creates EjbAccess::Migration and sends it back
  - Java calls nativeCreateTombstone(long nativeMigrationHandler, String address)
  - Rust calls EjbAccess::get_migration().create_tombstone(address), which creates `exonum_merkledb::Migration` in-place. The same workflow applies to the Scratchpad and Prefixed.
3. To create indexes, one must get `ViewRef` using `EjbAccess::get` method. `ViewRef` is expanded with 3 (actually 4, because of Prefixed) new variants, which significantly increases amount of code (see changes in `check_value`)
4. As `Prefixed` can be either mutable or immutable, new enum `EjbPrefixed` was introduced.

#### Q&A (Please read before asking uncomfortable questions):
Q: Why a lot of modules are disabled?
A: I didn't want to copy-paste similar code in a dozen of places. Ones more suitable approach to replace (or improve) `ViewRef` will be developed, I will update indexes as well.

Q: That's **a lot** of code! Supporting this would be a nightmare, isn't it?
A: Most of the code are simple pattern-matches, which are quite duplicated. It wouldn't be hard to write and test this code ones, but it's getting complicated ones we want to extend it, yes. However, I expect some improvements, e.g. we can try to implement `Access` trait for our `EjbAccess` and use it instead of `ViewRef`.

Q: What about ReadonlyFork?
A: No research was conducted yet. Perhaps it will be another enum variant.

Q: What about MigrationHelper?
A: Research is in-progress. Looks like we can implement native proxy with a bit of unsafe code, but the tests aren't ready yet. Further research will be done in a separate PR.

Q: Isn't current implementation heavily depends on internal implementation of MerkleDB? Creating Migration & co in-place works only because they are stateless wrappers around `Access`.
A: Probably yes.

Q: What an awful names!
A: Of course, don't look at names, we can change them at any time.

#### To sum it up
- The code exploded in a good-old combinatorial explosion.
- The solution (implementing Access for EjbAccess) looks promising
- All in all the implementation will work, even at the cost of far more native code to support.

---
See: https://jira.bf.local/browse/ECR-4120

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/rust/src/storage/db.rs
+++ b/exonum-java-binding/core/rust/src/storage/db.rs
@@ -12,7 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use exonum_merkledb::{Fork, Snapshot, migration::{Migration, Scratchpad}, access::Prefixed};
+use exonum_merkledb::{
+    access::Prefixed,
+    migration::{Migration, MigrationHelper, Scratchpad},
+    Fork, Snapshot,
+};
 use jni::{objects::JClass, JNIEnv};
 
 use handle::{self, Handle};
@@ -86,7 +90,7 @@ impl EjbAccess {
             EjbAccess::Migration(raw_view, ref namespace) => match raw_view {
                 View::RefFork(fork) => Migration::new(namespace, fork),
                 View::Owned(ViewOwned::Fork(fork)) => Migration::new(namespace, fork),
-                _ => unimplemented!()
+                _ => unimplemented!(),
             },
             _ => unimplemented!(),
         }
@@ -97,9 +101,9 @@ impl EjbAccess {
             EjbAccess::Scratchpad(raw_view, ref namespace) => match raw_view {
                 View::RefFork(fork) => Scratchpad::new(namespace, fork),
                 View::Owned(ViewOwned::Fork(fork)) => Scratchpad::new(namespace, fork),
-                _ => unimplemented!()
+                _ => unimplemented!(),
             },
-            _ => unimplemented!()
+            _ => unimplemented!(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


vitvakatu:
### How to improve Exonum MerkleDB API:

1. Get rid of lifetimes in `Migration`, `Scratchpad` and `Prefixed`. Copy `String`s instead. Using lifetimes in these structures is not only restricting for FFI code, but also illogical: the lifetime of `Migration` does not depend on `namespace`, but on `access`.
2. *Possibly* pass `MigrationHelper` by-value, not by mut-reference. I didn't yet finished the PoC, but using `&mut MigrationHelper` (which is used in MigrationScript) is really complicated and definitely requires unsafe code.
3. Get rid of lifetime in `BlockchainData` (I mean, in `InstanceDescriptor`). Again, same reason - `BlockchainData` must not depend on some fancy string literal, it's completely wrong.

cc @alekseysidorov @aleksuss @slowli do you have some opinions on the propositions above?

vitvakatu:
> If Migration/Prefixed/Scratchpad are created in place, will indexes work (I assume, that they are also destroyed in place)?

I believe they will, as indexes work now, with `&Fork` and `&Snapshot`. The trick is, we use unsafe to prolong lifetime of the reference to `&'static` (see `View::from_ref_fork` and `View::from_ref_snapshot`). It is considered safe only because we manually manage lifetimes of our Views (e.g. we destroy Fork's native handle when it can no longer be accessed). It looks like `ViewRef` doesn't need lifetime actually - it stores `&'static` references; references that are valid for the whole program execution. So as these Migration/Prefixed/Scratchpad are created from our `&'static Fork` or `&'static Snapshot`, they will be valid as long as we need.

I need to update tests a bit to prove "I believe" part though.

> How exactly Migration, Prefixed and Scratchpad are created (e.g., with which namespace) is likely to be the implementation detail of MigrationHelper. Hence, if we plan to re-use it, we must take the Accesses it creates, not create our own — because it is not specified how; and, as you mentioned, if these Accesses become stateful, we won’t be able to create their replacements in place at all. Is there a solution for this?

Yes, it's getting complicated with `MigrationHelper`. Getting rid of lifetimes in merkledb types, as I proposed above, will allow us to actually create M/S/P ones and send in through FFI border. Unfortunately it won't quite fix the situation with `MigrationHelper`, as it returns `Migration<&'lifetime_of_helper Fork>`. Though we should be able to use a bit of unsafe to fix this annoyance. I'm going to prepare a PoC shortly.

vitvakatu:
I've implemented a small PoC of MigrationHelper support. The idea is simple: we use proxy wrapper to store `&mut MigrationHelper` and prolong the lifetime using unsafe. As long as MigrationHelperProxy is used inside `MigrationScript` (or inside any scope it was created), everything should work fine (including the creation of indexes).

PR #1381, opened 2020-01-24
Improve documentation after Access:

dmitry-timofeev:
Improve documentation after recent Access introduction:
- Warn that IndexT#newInstance are not for services.
- Specify address resolution rules in Accesses.

## Overview

---

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/ValueSetIndexProxy.java
+++ b/exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/ValueSetIndexProxy.java
@@ -80,6 +80,9 @@ public final class ValueSetIndexProxy<E> extends AbstractIndexProxy
   /**
    * Creates a new value set.
    *
+   * <p><strong>Warning:</strong> do not invoke this method from service code, use
+   * {@link Access#getValueSet(IndexAddress, Serializer)}.
+   *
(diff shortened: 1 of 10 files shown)


exonum-java-binding/core/src/main/java/com/exonum/binding/core/storage/indices/ValueSetIndexProxy.java
```diff
@@ -80,6 +80,9 @@
   /**
    * Creates a new value set.
    *
+   * <p><strong>Warning:</strong> do not invoke this method from service code, use
```
vitvakatu: It's interesting, Rust has `pub(crate)` visibility, which restricts the usage of the method to the internal code of the library. Doesn't Java have similar feature?

dmitry-timofeev: No, there are package-private and protected visibility (package-private + inheritors from any package). 

At module level (Java 9+), there are exported _packages_ (everything else is non-exported = *reliably* hidden), but no methods.

Therefore, there are such ways to hide it:
 - Put indices and AbstractAccess in the same package, make this op package-private
 - Make the op package-private, *invoke via reflection*.

PR #1389, opened 2020-01-29
Update to revision c46b533 of Core [ECR-4130]

vitvakatu:
## Overview

This update brings resuming services and Access types without lifetime restrictions. Based on custom Exonum Core branch.

---
See: https://jira.bf.local/browse/ECR-4130

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/rust/src/proxy/runtime.rs
+++ b/exonum-java-binding/core/rust/src/proxy/runtime.rs
@@ -163,6 +163,34 @@ impl Runtime for JavaRuntimeProxy {
         })
     }
 
+    fn initiate_resuming_service(
+        &self,
+        context: ExecutionContext,
+        spec: &InstanceSpec,
+        parameters: Vec<u8>,
+    ) -> Result<(), ExecutionError> {
+        let serialized_instance_spec: Vec<u8> = spec.to_bytes();
+
+        jni_call_default(&self.exec, |env| {
+            let fork_handle = to_handle(View::from_ref_mut_fork(context.fork));
+            let instance_spec =
+                JObject::from(env.byte_array_from_slice(&serialized_instance_spec)?);
+            let parameters = JObject::from(env.byte_array_from_slice(&parameters)?);
+
+            env.call_method_unchecked(
+                self.runtime_adapter.as_obj(),
+                runtime_adapter::initiate_resuming_service_id(),
+                JavaType::Primitive(Primitive::Void),
+                &[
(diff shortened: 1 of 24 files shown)


exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -163,6 +163,34 @@ impl Runtime for JavaRuntimeProxy {
...
 
+    fn initiate_resuming_service(
+        &self,
+        context: ExecutionContext,
+        spec: &InstanceSpec,
+        parameters: Vec<u8>,
+    ) -> Result<(), ExecutionError> {
+        let serialized_instance_spec: Vec<u8> = spec.to_bytes();
+
+        jni_call_default(&self.exec, |env| {
```
dmitry-timofeev: Shan't it be `jni_call_transaction`? Same with `initiate_starting` — @ivan-ochc — does it work as expected?

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -163,6 +163,34 @@ impl Runtime for JavaRuntimeProxy {
...
+            let fork_handle = to_handle(View::from_ref_mut_fork(context.fork));
+            let instance_spec =
```
dmitry-timofeev: We convert InstanceSpec to bytes three times — why dont' we extract this op in a method (possibly, there are other opportunities)?

exonum-java-binding/core/rust/src/proxy/runtime.rs
vitvakatu: According to documentation of `CallType` we should handle it with `jni_call_transaction`, yes.
```rust
/// Type of a call to a service.
#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
#[serde(tag = "call_type", rename_all = "snake_case")]
pub enum CallType {
    /// Service initialization or resuming.
    Constructor,
    /// Service method.
    Method {
.....
```

exonum-java-binding/core/rust/src/proxy/runtime.rs
vitvakatu: I reused `proto_to_java` function from our utils module - seems better

dmitry-timofeev: Definitely, thanks!

PR #1403, opened 2020-02-03
[ECR-3469] Standardized service configuration

bullet-tooth:
## Overview

---
See: https://jira.bf.local/browse/ECR-3469

### Definition of Done

- [ ] Public API has Javadoc


exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Configuration.java
```diff
@@ -28,6 +32,9 @@
  * {@linkplain Service#initialize(Fork, Configuration) passes the configuration parameters}
  * to the newly created service instance.
  *
+ * <p>The configuration can be any protobuf message, but it is recommended to use
+ * {@linkplain ServiceConfiguration standard service configuration message} in most cases.
```
dmitry-timofeev:
```java
 * <p>Services that have few arguments are encouraged to use the standard protobuf 
 * message {@link ServiceConfiguration}. It supports common text-based configuration formats.
```

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Configuration.java
```diff
@@ -48,4 +58,44 @@
    */
   <MessageT extends MessageLite> MessageT getAsMessage(Class<MessageT> parametersType);
 
+  /**
+   * Returns the configuration as a plain text string.
+   *
+   * @throws IllegalArgumentException if the actual type of the configuration is not an instance of
+   *     {@link ServiceConfiguration};
+   *     or the configuration is not in the {@linkplain Format#TEXT plain text} format
```
dmitry-timofeev: I'd allow to get any format as text (e.g., a service needs a fancier Gson configuration).

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Configuration.java
```diff
@@ -48,4 +58,44 @@
...
+  String getAsString();
+
+  /**
+   * Returns the configuration as a raw JSON string.
+   *
+   * @throws IllegalArgumentException if the actual type of the configuration is not an instance of
+   *     {@link ServiceConfiguration};
+   *     or the configuration is not in the {@linkplain Format#JSON JSON} format
+   */
+  String getAsJson();
```
dmitry-timofeev: I'd drop it in favour of a more flexible getAsString.

exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/Configuration.java
```diff
@@ -48,4 +58,44 @@
...
+  /**
+   * Returns the configuration as an object  of the given type decoded from the underlying JSON.
+   *
+   * @param <T> the type of the configuration object
+   * @param configType the class of T
+   * @throws IllegalArgumentException if the actual type of the configuration is not an instance of
+   *     {@link ServiceConfiguration};
+   *     or the configuration is not in the {@linkplain Format#JSON JSON} format
+   * @throws JsonSyntaxException if json is not a valid representation for an object of the given
+   *     type
```
dmitry-timofeev: @throws JsonParseException in case of JSON parse error

exonum-java-binding/core/src/main/proto/service.proto
```diff
@@ -0,0 +1,42 @@
...
+
+syntax = "proto3";
+
+package exonum.binding;
+
+option java_package = "com.exonum.binding.common.messages";
+
+/**
+TODO: move this file to exonum-proto-sources-repo
+currently this action is blocked till native update to the latest revision
```
dmitry-timofeev: I think it must be enough to place it in common and give an appropriate name.

exonum-java-binding/core/src/main/proto/service.proto
```diff
@@ -0,0 +1,42 @@
...
+*/
+
+// An object for the service configuration which passed to `service#initialize` method.
+// It brings a possibility to configure a service using any of provided formats.
+// The format provides the type of the configuration encoded as string value.
```
dmitry-timofeev: Service configuration arguments in a text format.
The string must be in one of the pre-defined formats.

exonum-java-binding/core/src/main/proto/service.proto
```diff
@@ -0,0 +1,42 @@
...
+message ServiceConfiguration {
+    enum Format {
+        NONE = 0;
```
dmitry-timofeev: What is the purpose of NONE? If it is neither TEXT nor any from the other formats,    then the messages shall not be used.

exonum-java-binding/core/src/main/proto/service.proto
```diff
@@ -0,0 +1,42 @@
...
+        // Identifies that the given configuration represented as `plain text`.
+        TEXT = 1;
+        // Identifies that the given configuration represented as Json.
+        JSON = 2;
+        // Identifies that the given configuration represented as Properties,
+        // i.e. `key=value` pairs.
```
dmitry-timofeev: There is actually way more to this, I'd link

        // Identifies that the given configuration represented as Java Properties,
        // i.e., key value pairs.
        // See the format specification in j.u.Properties: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Properties.html#load(java.io.Reader)

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceConfiguration.java
```diff
@@ -36,6 +44,34 @@
...
+
+  @Override
+  public String getAsJson() {
+    return validateAndGet(Format.JSON);
+  }
+
+  @Override
+  public <T> T getAsJson(Class<T> configType) {
+    String configuration = getAsJson();
+    return new Gson().fromJson(configuration, configType);
```
dmitry-timofeev: We have a preconfigured Gson with support for our types in common.

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/StandardServiceConfigurationTestKitTest.java
```diff
@@ -0,0 +1,106 @@
...
+import com.google.common.collect.Maps;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+import java.util.Map;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+public class StandardServiceConfigurationTestKitTest {
```
dmitry-timofeev: Why is this test added to testkit — it checks some runtime-specific functionality,
that does not affect either testkit or the core? May be in integration-tests. 

However, I'd reconsider if it is needed — as this functionality is runtime-specific and based on the same mechanism that is already tested (= passing config parameters from test thru testkit, core, to runtime); but testing it at this level is rather complex.

exonum-java-binding/core/src/main/proto/service.proto
bullet-tooth: the purpose of None that protobuf may instantiate a message from bytes with default values in fields.
Actually It happens in test:
```
  @Test
  void getAsPlainStringIsNotServiceConfiguration() {
    byte[] serializedConfig = anyId().toByteArray();
    ServiceConfiguration serviceConfiguration = new ServiceConfiguration(serializedConfig);

    assertThrows(IllegalArgumentException.class, serviceConfiguration::getAsString);
  }
```
the underlying object has format=None and value = ""

dmitry-timofeev: If the format is unset, it will [have the 0 enum value](https://developers.google.com/protocol-buffers/docs/proto3#enum), which we can define to be plain text, which is acceptable. 
None/plain_text as a default for unset doesn't have much difference: you can always get it as string, but cannot decode in any other format.

> the underlying object has format=None and value = ""

This particular test does not test what it says: it takes some array that _may_ be decoded, and verifies there is _some_ error. It must pass something that protobuf cannot decode correctly, and verify it rejects it. Here, it seems, it just accepts the byte array as filled with defaults (zeroes).

exonum-java-binding/testkit/src/test/java/com/exonum/binding/testkit/StandardServiceConfigurationTestKitTest.java
dmitry-timofeev: > However, I'd reconsider if it is needed — as this functionality is runtime-specific and based on the same mechanism that is already tested (= passing config parameters from test thru testkit, core, to runtime); but testing it at this level is rather complex.

I am not convinced its value justifies the cost.

exonum-java-binding/core/src/main/proto/service.proto
```diff
@@ -0,0 +1,41 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+package exonum.binding;
+
+option java_package = "com.exonum.binding.common.messages";
+
+/**
+TODO: ECR-4199 move this file to exonum-proto-sources-repo
```
dmitry-timofeev: to common (the task says it correct) :-)

PR #1408, opened 2020-02-05
Update to recent Exonum revision [ECR-4201]

vitvakatu:
## Overview

This PR updates native to the most recent (almost) Exonum revision.
Attention: The native code updates are aimed at reduction of diff size, not on proper implementation. The most important consequence is that imports of iterator types in `storage` module are not optimal and must be rewritten (ECR-TBD). Also see individual review comments.

TestKit tests are __not__ working.
Integration tests module is __not__ tested.

---
See: https://jira.bf.local/browse/ECR-4201

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/rust/integration_tests/tests/node_context.rs
+++ b/exonum-java-binding/core/rust/integration_tests/tests/node_context.rs
@@ -78,10 +78,9 @@ fn submit_transaction_to_missing_service() {
 }
 
 fn create_transaction(instance_id: u32) -> AnyTx {
-    AnyTx::new(
-        CallInfo::new(instance_id, TEST_TRANSACTION_ID),
-        TEST_TRANSACTION_PAYLOAD.to_vec(),
-    )
+    let call_info = CallInfo::new(instance_id, TEST_TRANSACTION_ID);
+    let tx_args = TEST_TRANSACTION_PAYLOAD.to_vec();
+    AnyTx::new(call_info, tx_args)
 }
 
 fn create_node(keypair: (PublicKey, SecretKey)) -> (Node, Receiver<Verified<AnyTx>>) {
--- a/exonum-java-binding/core/rust/integration_tests/tests/node_context.rs
+++ b/exonum-java-binding/core/rust/integration_tests/tests/node_context.rs
@@ -78,13 +78,10 @@ fn submit_transaction_to_missing_service() {
 }
 
 fn create_transaction(instance_id: u32) -> AnyTx {
-    AnyTx {
-        call_info: CallInfo {
-            instance_id,
-            method_id: TEST_TRANSACTION_ID,
-        },
-        arguments: TEST_TRANSACTION_PAYLOAD.to_vec(),
-    }
+    AnyTx::new(
+        CallInfo::new(instance_id, TEST_TRANSACTION_ID),
+        TEST_TRANSACTION_PAYLOAD.to_vec(),
+    )
 }
 
 fn create_node(keypair: (PublicKey, SecretKey)) -> (Node, Receiver<Verified<AnyTx>>) {
--- a/exonum-java-binding/core/rust/src/proxy/runtime.rs
+++ b/exonum-java-binding/core/rust/src/proxy/runtime.rs
@@ -17,11 +17,11 @@
 use exonum::{
     blockchain::Blockchain,
     crypto::{Hash, PublicKey},
-    exonum_merkledb::{BinaryValue, Snapshot},
+    merkledb::{BinaryValue, Snapshot},
     runtime::{
         migrations::{InitMigrationError, MigrationScript},
         versioning::Version,
-        ArtifactId, CallInfo, Caller, ExecutionContext, ExecutionError, InstanceId, InstanceSpec,
+        ArtifactId, Caller, ExecutionContext, ExecutionError, InstanceId, InstanceSpec,
         InstanceStatus, Mailbox, Runtime, RuntimeIdentifier, SnapshotExt, WellKnownRuntime,
     },
 };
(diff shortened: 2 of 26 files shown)


exonum-java-binding/core/rust/integration_tests/tests/node_context.rs
```diff
@@ -78,13 +78,10 @@ fn submit_transaction_to_missing_service() {
...
 fn create_transaction(instance_id: u32) -> AnyTx {
-    AnyTx {
-        call_info: CallInfo {
-            instance_id,
-            method_id: TEST_TRANSACTION_ID,
-        },
-        arguments: TEST_TRANSACTION_PAYLOAD.to_vec(),
-    }
+    AnyTx::new(
+        CallInfo::new(instance_id, TEST_TRANSACTION_ID),
```
vitvakatu: Split to two statements might be more readable... 🤔 I'll add a note to make such refactoring.

exonum-java-binding/core/rust/src/proxy/runtime.rs
```diff
@@ -134,20 +134,28 @@ impl Runtime for JavaRuntimeProxy {
     fn initiate_adding_service(
         &self,
         context: ExecutionContext<'_>,
-        spec: &InstanceSpec,
+        artifact_id: &ArtifactId,
         parameters: Vec<u8>,
     ) -> Result<(), ExecutionError> {
+        // TODO: change Java API to accept ArtifactId, not InstanceSpec
```
vitvakatu: Native method lost InstanceSpec parameter, so we constructing it here in order not to break Java code. Must be changed in the future (along with Java changes). initiate_resuming_service is the same

exonum-java-binding/core/src/main/java/com/exonum/binding/core/blockchain/serialization/BlockSerializer.java
```diff
@@ -94,7 +94,7 @@ private static HashCode toHashCode(Hash hash) {
 
   @VisibleForTesting
   static ImmutableMap<String, ByteString> toHeadersMap(AdditionalHeaders headers) {
-    return headers.getHeaders().getEntryList()
+    return headers.getHeaders().getEntriesList()
```
vitvakatu: The most mysterious change in the entire PR

exonum-java-binding/core/src/test/java/com/exonum/binding/core/blockchain/BlockchainDataIntegrationTest.java
```diff
@@ -148,9 +148,12 @@ void createIndexProofFromForkDisallowed() throws CloseFailuresException {
     Fork fork = db.createFork(cleaner);
     BlockchainData blockchainData = BlockchainData.fromRawAccess(fork, "test-service");
 
-    Exception e = assertThrows(IllegalArgumentException.class,
-        () -> blockchainData.createIndexProof(simpleIndexName));
-    assertThat(e).hasMessageFindingMatch("(?i)(Snapshot)|(Fork)");
+    // TODO: it's impossible to get IndexProof for Fork, but we check only for canModify,
```
vitvakatu: @dmitry-timofeev ^

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/ServiceNodeProxyIntegrationTest.java
```diff
@@ -74,6 +75,7 @@ void restrictSubmitTransaction() {
   }
 
   @Test
+  @Disabled("ECR-3828")
```
vitvakatu: Actually, I don't know why and what is the desired behaviour

exonum-java-binding/core/rust/src/proxy/runtime.rs
dmitry-timofeev: Why would we change that if the {name, id} are required to instantiate a service?
         I'd rather have it as an explicit parameter and not buried in the ExecutionContext.

We can consider making an ExecutionContext proxy (when implementing service interfaces);
         or passing {name, id, artifact_id} separately if it makes native implementation
         easier.

exonum-java-binding/core/src/test/java/com/exonum/binding/core/blockchain/BlockchainDataIntegrationTest.java
dmitry-timofeev: s/IllegalArgumentException/RuntimeException/

exonum-java-binding/core/src/test/java/com/exonum/binding/core/runtime/ServiceNodeProxyIntegrationTest.java
dmitry-timofeev: Because the function to which the transformation is applied is not invoked by the Node. If we could create a NodeProxy from java in tests, we would be able to write an integration test for that (with actual invocation). Until then, I'd just test:

```
  @Test
  void withSnapshot() {
    decorator.withBlockchainData(SNAPSHOT_FUNCTION);

    verify(node).withSnapshot(any(Function.class));
  }
```

exonum-java-binding/core/rust/src/proxy/runtime.rs
vitvakatu: passing name, id, artifact_id would work, I suppose

exonum-java-binding/core/src/main/java/com/exonum/binding/core/blockchain/serialization/BlockSerializer.java
dmitry-timofeev: I think the message was updated earlier (before BlockchainData merge, but after its implementation), and as no CI was configured — we just didn't notice during the merge.

PR #1424, opened 2020-02-13
Run python tests on Travis [ECR-3870]

vitvakatu:
## Overview

---
See: https://jira.bf.local/browse/ECR-3870

### Definition of Done

- [ ] Public API has Javadoc


.travis/run_travis_job.sh
```diff
@@ -53,4 +53,25 @@ else
...
+      # Install exonum-launcher
+      git clone https://github.com/exonum/exonum-launcher.git
+      pip3 install --user -r exonum-launcher/requirements.txt
+      pip3 install --user -e exonum-launcher --no-binary=protobuf protobuf
+      # Install exonum_launcher_java_plugins
+      pip3 install --user -e .
+      # Download latest protobuf compiler
+      wget https://github.com/protocolbuffers/protobuf/releases/download/v3.11.3/protoc-3.11.3-linux-x86_64.zip
+      unzip protoc-3.11.3-linux-x86_64.zip
+      export PROTOC="$(pwd)/bin/protoc"
```
dmitry-timofeev: Shan't we add protoc to `PATH`?

.travis/run_travis_job.sh
```diff
@@ -53,4 +53,25 @@ else
...
+      curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
+      python3.7 get-pip.py --user
+      # Install exonum-launcher
+      git clone https://github.com/exonum/exonum-launcher.git
+      pip3 install --user -r exonum-launcher/requirements.txt
+      pip3 install --user -e exonum-launcher --no-binary=protobuf protobuf
+      # Install exonum_launcher_java_plugins
+      pip3 install --user -e .
+      # Download latest protobuf compiler
+      wget https://github.com/protocolbuffers/protobuf/releases/download/v3.11.3/protoc-3.11.3-linux-x86_64.zip
```
dmitry-timofeev: What's interesting is that we already have several protoc copies downloaded by the Maven plugin, e.g.
```
$ ls common/target/protoc-plugins/     
protoc-3.11.0-osx-x86_64.exe
```

but it's probably better to install explicitly.

.travis/run_travis_job.sh
```diff
@@ -53,4 +53,25 @@ else
...
     fi
+
+    # Test exonum_launcher_java_plugins
+    if [[ "${TRAVIS_JOB_NAME}" == "Linux JDK 8 CHECK_RUST=false" ]]; then
+      cd "${TRAVIS_BUILD_DIR}/exonum-java-binding/exonum_launcher_java_plugins"
+      # Install pip
+      curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
+      python3.7 get-pip.py --user
+      # Install exonum-launcher
+      git clone https://github.com/exonum/exonum-launcher.git
```
dmitry-timofeev: Shallow clone?

.travis/run_travis_job.sh
```diff
@@ -53,4 +53,25 @@ else
...
+      pip3 install --user -r exonum-launcher/requirements.txt
```
dmitry-timofeev: Do we always depend on the master of the launcher?

.travis/run_travis_job.sh
```diff
@@ -53,4 +53,25 @@ else
     if [[ "${TRAVIS_JOB_NAME}" == "Linux JDK 8 CHECK_RUST=false" ]]; then
       mvn org.eluder.coveralls:coveralls-maven-plugin:report
     fi
+
+    # Test exonum_launcher_java_plugins
```
dmitry-timofeev: Would it make sense to move them to a separate job?

.travis.yml
```diff
@@ -17,6 +17,7 @@ addons:
       - sourceline: 'ppa:fsgmhoward/shadowsocks-libev'
       - sourceline: 'ppa:maarten-fonville/protobuf'
       - sourceline: 'ppa:exonum/rocksdb'
+      - sourceline: 'ppa:deadsnakes/ppa'
```
dmitry-timofeev: Мертвые змеи в дополнение к теневым носкам 🙃

.travis/run_travis_job.sh
vitvakatu: It was my initial attempt, but actually exonum-python-client is looking for `PROTOC` env var, so it's enough.

.travis/run_travis_job.sh
vitvakatu: Actually installing via requirements.txt works well - now we depend on a particular exonum-launcher revision.

.travis/run_travis_job.sh
vitvakatu: Python plugins need protobuf files, so they must be run after `core generate-sources`. Do you think a separate job would be better?

vitvakatu:
@dmitry-timofeev actually managed to rewrite exonum-launcher installation with a usage of pip and requirements.txt - works pretty well.

.travis/run_travis_job.sh
dmitry-timofeev: I don't know, it seems to require some separate setup (python, pip, protoc) so it makes some sense. Also it would make the feedback more focused. 

Or, would it make sense to add commands to test the python plugin to the `run_all_tests.sh` script? What would it require from us in terms of keeping them green locally, given the python3 is available (regular `pip install`s?)?

vitvakatu: It would be better to use virtualenv for local tests - so `./run_all_tests.sh` will not affect (and won't be affected by) other python packages installed.

vitvakatu: I think I'd move it to a separate script, add it to run_all_tests for local tests and also add a separate job

dmitry-timofeev: Good, but I think it is rather either/or: if we add it to local tests run with `run_all_tests`, it will be run by each job.

vitvakatu: After some consideration, safe script for local tests seems redundant - changes to plugins are rare and tests can be run manually if needed.

.travis/run_python_tests.sh
```diff
@@ -0,0 +1,35 @@
...
+set -eu -o pipefail
+
+# Echo commands so that the progress can be seen in CI server logs.
+set -x
+
+cd "${TRAVIS_BUILD_DIR}/exonum-java-binding"
+
+# Generate protobuf files needed for plugins
+source ./tests_profile
+mvn install -DskipTests -DskipRustLibBuild -pl core -am
```
dmitry-timofeev: Why install if you need to `generate-sources` (and no Java tests or Rust build will happen to skip)?

dmitry-timofeev:
As a future improvement, we might also skip the Rust installation.

.travis/run_python_tests.sh
vitvakatu: I experienced issues with maven unable to resolve project dependencies when used `generate-sources` on Travis.

.travis/run_python_tests.sh
```diff
@@ -11,7 +11,8 @@ cd "${TRAVIS_BUILD_DIR}/exonum-java-binding"
 
 # Generate protobuf files needed for plugins
 source ./tests_profile
```
dmitry-timofeev: RUSTFLAGS=none , otherwise it is likely to fail with no Rust

.travis/run_python_tests.sh
dmitry-timofeev: That's unexpected — which exactly dependencies? 3rd party or a local subproject?

vitvakatu: Build log here: https://api.travis-ci.com/v3/job/287154757/log.txt

.travis/run_python_tests.sh
vitvakatu: :thinking: I don't think it'll help actually, we're always overwrite RUSTFLAGS

vitvakatu: Ah, actually it works, because our tests_profile does not fail on errors

exonum-java-binding/common/pom.xml
```diff
@@ -148,14 +154,43 @@
         <groupId>org.xolstice.maven.plugins</groupId>
         <artifactId>protobuf-maven-plugin</artifactId>
         <extensions>true</extensions>
+        <configuration>
+          <protocArtifact>com.google.protobuf:protoc:${protobuf.version}:exe:${os.detected.classifier}</protocArtifact>
+          <protoSourceRoot>${basedir}/src/main/proto/exonum/java</protoSourceRoot>
```
dmitry-timofeev: Why is this needed? It recognizes the packages, the protoSourceRoot must remain the default.

PR #1446, opened 2020-02-21
Enable Java 11 [ECR-521]

dmitry-timofeev:
## Overview

Build "client" libraries for Java 8 using javac 11+;
"framework" libraries — for Java 11.

Build the project on Travis with the most recent
JDK (13 at the moment) on each push; and weekly —
on the minimum supported (11), as the developers
use the most recent JDK.

### Todo

- [x] Update Travis config
- [x] Update the sys tests (separately)
- [x] Smallish issues (see Jira)
- [x] Javadoc index is broken (search functionality)
- [x] Consider if the performance hit is acceptable

<details>
<summary>Maven Toolchains Performance Issues</summary>

Here are the observations for a “Baseline” experiment (everything run on JDK 11); and “Toolchains” (Maven is run on 11 and modules use the configured toolchain: either 8 or 11).

The total time of `mvn test-compile`:
-	Baseline: 22 sec
-	Toolchains: **52 sec.**

Time output:
-	Baseline:
real	0m22.937s
user	0m24.761s
sys	0m2.621s

-	Toolchains: 
real	0m53.449s
user	**2m15.315s**
sys	0m6.633s
 
The “compile” goal time for a couple of modules, profiled with https://github.com/jcgay/maven-profiler:
-	A module using 8:
o	Baseline: 1.12 s
o	Toolchains: 2.66 s (+1.54 s)
-	A (bigger) module using 11:
o	Baseline: 1.32 s
o	Toolchains: 4.03 s (+2.71 s)
-	A small module using 11:
o	0.08
o	1.24 (+1.16 s)

Script if needed: https://github.com/dmitry-timofeev/exonum-java-binding-1/blob/eb4538654cd32ab7ea82a3da802b8c6ba7e8e5c2/exonum-java-binding/test-toolchains.sh#L18 

#### Observations
1.	There is a performance hit for each module in the project. The minimum slowdown for a module containing a couple of source files exceeds 1 second.
2.	The slowdown seems to be proportional to the module size, but not significantly.

</details>

---
See: https://jira.bf.local/browse/ECR-521


CONTRIBUTING.md
```diff
@@ -17,7 +17,14 @@ This guide shows how to build the whole project only.
 ### System Dependencies
 You need to install the following dependencies:
   * Linux or macOS. Windows support is coming soon. <!-- TODO: Link Java roadmap when it is published -->
-  * [JDK 1.8+](https://jdk.java.net/).
+  * [JDK 1.8](https://jdk.java.net/) and JDK 11.0+ toolchains.
```
bullet-tooth: How about to inject a link to the toolchains? 
Also, configuration below starts with something related to `jenv` which is confusing ones who don't use it. (for example, I use sdkman for managind java versions, so what should I do then?)
So I propose to provide clear step-by-step instructions about how to install and configure toolchains or add a link to it.

toolchains-template-min.xml
```diff
@@ -0,0 +1,46 @@
...
+       `JDK_BASE_DIR` is a directory containing (symlinks to) JDK installations with their
+       directory names equal to the version (e.g., `1.8` or `11.0.2`)
+       -->
+  <toolchain>
+    <type>jdk</type>
+    <provides>
+      <version>1.8</version>
+    </provides>
+    <configuration>
+      <jdkHome>JDK_BASE_DIR/1.8</jdkHome>
```
bullet-tooth: it would be hard to integrate it with sdkman such as it uses another format for java version distributions:

image

dmitry-timofeev: It is a template for boxes that use such layout. If one doesn't — one shall enter jdkHome manually.

CONTRIBUTING.md
dmitry-timofeev: > starts with something related to jenv 

Not necessarily: it specifies to which directory layout these instructions are appropriate.

> (for example, I use sdkman for managind java versions, so what should I do then?)

It does say: "If you don't use jenv, replace with appropriate location" (assuming you have the expected layout).

I agree though that a line with instructions for those who use a different layout could be added — but it is likely we won't use toolchains during development (maybe, we will keep them on CI only to _run tests_ on 8 for client libraries).

PR #1459, opened 2020-03-02
Adjust the rules for Javadoc generation:

dmitry-timofeev:
## Overview

AdoptOpenJDK 11.0.3+ back-ported the fix which
was originally integrated into JDK 13, hence
the option is only needed on [11.0.0–11.0.2]
range.


dmitry-timofeev:
though it does not fix the problem

PR #1462, opened 2020-03-04
Unify the context in transaction-like methods: [ECR-4303]

dmitry-timofeev:
## Overview

Use `TransactionContext` in transaction-like methods:
- Service#initialize, #resume, #before- and #afterTransactions
- Configurable

That enables easier re-use of @Transaction methods, as they
already accept TransactionContext; and makes the API a little
easier to learn because there are fewer immediate abstractions
(though you will have to use TransactionContext#getBlockchainData
to access the data anyway).

---
See: 
- **ECR-4303**
- **https://chat.bfnet.site/bitfury/pl/9yx4q3z5opnb8g7q49756xrhqo**
- ECR-4037 — might make more sense now.

### Definition of Done

- [ ] Public API has Javadoc


exonum-java-binding/CHANGELOG.md
```diff
@@ -98,6 +98,13 @@ stable Exonum release. See [release notes][exonum-1.0.0-rc.1] for details.
   Any exceptions thrown in these methods are saved in the blockchain
   in `Blockchain#getCallErrors` and can be retrieved by any services or
   light clients.
+- Replaced `BlockchainData` argument in transaction-like `Service` and 
+  `Configurable` methods with `TransactionContext`. The `BlockchainData`
+  remains accessible via `TransactionContext#getBlockchainData`,
+  and service data via `TransactionContext#getServiceData`.
+  <!-- TODO: Shall we rename it into `ExecutionContext` (as
+  TransactionExecutionException has become ExecutionException)?
+  -->
```
dmitry-timofeev: # ⚠️

exonum-java-binding/core/src/main/java/com/exonum/binding/core/transaction/TransactionContext.java
```diff
@@ -54,14 +59,16 @@ default Prefixed getServiceData() {
 
   /**
    * Returns SHA-256 hash of the {@linkplain TransactionMessage transaction message} that
-   * carried the payload of the transaction.
-   * Each transaction message is uniquely identified by its hash; the messages are persisted
+   * carried the payload of the transaction; or a zero hash if no message corresponds to this
```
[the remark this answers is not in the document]
dmitry-timofeev: Yes, it would make sense, since we are on Java 11, it has acquired some methods that make it easier to use than previously, where it added some friction.

exonum-java-binding/CHANGELOG.md
bullet-tooth: I like renaming to `ExecutionContext`. For example, `TransactionContext` at `initialize` leads to a question "what kind of transaction do I have here?".

exonum-java-binding/core/src/main/java/com/exonum/binding/core/transaction/TransactionContext.java
```diff
@@ -98,6 +105,8 @@ static Builder builder() {
     private String serviceName;
     private Integer serviceId;
 
+    // todo: Init hash and author to zeroes? Or always leave to the client (as now)?
```
bullet-tooth: Now is ok because the context is created on the framework site.

exonum-java-binding/CHANGELOG.md
dmitry-timofeev: OK, thanks, will do here.

PR #1467, opened 2020-03-10
Add car registry tutorial code [ECR-4301]

dmitry-timofeev:
## Overview

Adds the car-registry service code; and a car registry client (not covered by the tutorial).

The tutorial to be reviewed first: https://github.com/exonum/exonum-doc/pull/414 

---
See: https://jira.bf.local/browse/ECR-4304

### Definition of Done

- [x] Public API has Javadoc


exonum-java-binding/tutorials/car-registry/README.md
```diff
@@ -0,0 +1,38 @@
+
+# car-registry
```
dmitry-timofeev: This is a generated readme, probably, shall be replaced with a link to the tutorial, once it is published.

exonum-java-binding/tutorials/car-registry/car-registry-client/src/main/java/com/example/car/client/Config.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package com.example.car.client;
+
+/**
+ * The client application rudimentary configuration.
+ */
+final class Config {
+  static final String NODE_PUBLIC_API_HOST = "http://127.0.0.1:3000";
+  static final String NODE_JAVA_API_HOST = "http://127.0.0.1:7000";
```
bullet-tooth: It's worth to add a few words in the tutorial why do we send tx to one endpoint and read the state from another

exonum-java-binding/tutorials/car-registry/car-registry-messages/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
+    </dependency>
+
+    <!--
+    TODO: Shall we include them, if by default Exonum types are not really needed
```
bullet-tooth: it may need for advanced users and I believe they can add the dependency themselves.

exonum-java-binding/tutorials/car-registry/car-registry-service/src/main/java/com/example/car/MyService.java
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  private void addVehicle(AddVehicle args, Prefixed serviceData) {
+    var schema = new MySchema(serviceData);
+    ProofMapIndexProxy<String, Vehicle> vehicles = schema.vehicles();
+
+    // Check there is no vehicle with such id in the registry yet
+    var newVehicle = args.getNewVehicle();
+    var id = newVehicle.getId();
+    if (vehicles.containsKey(id)) {
```
bullet-tooth: `ExecutionPreconditions`?

exonum-java-binding/tutorials/car-registry/car-registry-service/src/main/java/com/example/car/MyService.java
```diff
@@ -0,0 +1,148 @@
...
+    var id = args.getId();
+    if (!vehicles.containsKey(id)) {
+      throw new ExecutionException(NO_VEHICLE_ERROR_CODE, "No vehicle with such id: " + id);
+    }
+
+    // Update the owner
+    // Get the current entry
+    var vehicleEntry = vehicles.get(id);
+    // Update the owner
+    var newOwner = args.getNewOwner();
```
bullet-tooth: It will be good to add a check that this operation is performed by the car owner. And it will add some clarity about why do we need keys. But it will add complexity.

exonum-java-binding/tutorials/car-registry/start-testnet.sh
```diff
@@ -0,0 +1,62 @@
...
+header "3 Finalize Configuration"
+exonum-java finalize \
+    "$TESTNET_PATH"/sec.toml \
+    "$TESTNET_PATH"/node.toml \
+    --public-configs "$TESTNET_PATH"/pub.toml
+
+header "4 Start Test Network"
+exonum-java run \
+    --node-config "$TESTNET_PATH"/node.toml \
+    --artifacts-path "$TESTNET_PATH"/artifacts \
```
bullet-tooth: add here a path to `target`?

exonum-java-binding/tutorials/car-registry/car-registry-service/src/main/java/com/example/car/MyService.java
dmitry-timofeev: I decided against their use them for clarity (you see that an ExecutionException is being thrown).

exonum-java-binding/tutorials/car-registry/car-registry-service/src/main/java/com/example/car/MyService.java
dmitry-timofeev: Yes, we currently don't have a registry of car owners :-) We say this topic (authorization) is out of scope in the beginning, but it would be nice to cover in a more advanced tutorial/how to.

exonum-java-binding/tutorials/car-registry/start-testnet.sh
dmitry-timofeev: On top of https://github.com/exonum/exonum-doc/pull/414#discussion_r390990994 , I'd add that this comes from the archetype; therefore, if we decide it is better to have the target, we shall change the archetype.

exonum-java-binding/tutorials/car-registry/car-registry-service/src/main/java/com/example/car/MyService.java
dmitry-timofeev: ExecutionException already references the preconditions at class-level Javadoc, we may add 

```java
 * @see ExecutionPreconditions
```

to each constructor as well.

exonum-java-binding/tutorials/car-registry/car-registry-service/src/main/java/com/example/car/MySchema.java
```diff
@@ -0,0 +1,54 @@
...
+    this.access = checkNotNull(serviceData);
+  }
+
+  // ci-block ci-vehicles {
+  /**
+   * Provides access to the current state of the vehicles registry.
+   */
+  public ProofMapIndexProxy<String, Vehicle> vehicles() {
+    var address = IndexAddress.valueOf("vehicles");
+    var keySerializer = StandardSerializers.string();
```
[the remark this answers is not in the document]
dmitry-timofeev: There is no particular reason. I don't think integers would be more difficult. We could change that to integers; that would lead to more diversity in message definitions, which is good. However, such invocations

```shell
java -jar car-registry-client/target/car-registry-client-1.0.0.jar \
  add-vehicle -a -n=test-car-registry "My car" "VW" "Polo" "$USER"
```

will have to use some int instead of the id/description "My car".

exonum-java-binding/tutorials/car-registry/car-registry-client/src/main/java/com/example/car/client/Config.java
dmitry-timofeev: But the present tutorial does not show requests to service at that level: both operations are hidden inside this CLI client that comes with the tutorial. That is, the user does not interact with these endpoints directly, but through a pre-configured client, therefore, we can't just add that there are two servers for different uses, we will have to explain a bit more.

Shall we add some explanation here for the curious?

dmitry-timofeev:
There is a question on how to integrate the tutorials into our build. They need a path to java_bindings library to run their ITs. They are not, however, child modules (because they must be independent from the main project, otherwise it won't be possible to build them separately from EJB), hence we cannot pass a property from the parent.

We could:
- Build car-registry from the enclosing project (EJB) as a nested module (non-child); and auto-detect such build from the environment variables we use (e.g., RUSTFLAGS). A little bit hacky. Also, in the unlikely event the user has RUSTFLAGS set, will break :-) 
- Build them separately through a script that explicitly passes the needed property:
```
mvn verify -DnativeLibPath=core/rust/target/debug -f tutorials/car-registry
```

Any better options?

PR #1470, opened 2020-03-13
Python scripts for build [ECR-4308]

vitvakatu:
## Overview

This is the experimental re-implementation of `package_app.sh` and `run_all_tests.sh` scripts in Python. Should work on Windows, Mac and Linux. It does not support some existing features, namely build settings (`--skip-tests`, `--release`, `--skip-cargo-clean`), but it can be easily added in the future.

__Please note: there are possible improvements to Python code to reduce code duplication, improve structure and readability, but this PR is for review of a general approach__

Pros:
- Python is more readable than bash or Powershell
- No code duplication between bash and powershell scripts

Cons:
- Triple code duplication: tests_profile.sh, tests_profile.ps1, tests_profile.py

---
See: https://jira.bf.local/browse/ECR-4308

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/package_app.py
+++ b/exonum-java-binding/package_app.py
@@ -1,0 +1,148 @@
+import os
+import subprocess
+import sys
+import fnmatch
+from shutil import copy
+from pathlib import Path
+
+
+def shared_lib_extension():
+    if os.name == 'posix':
+        return ".so"
+    elif os.name == 'os2':
+        return ".dylib"
+    elif os.name == 'nt':
+        return ".dll"
+    else:
+        raise Exception(f"This OS ({os.name}) is unsupported")
+
+
+def jvm_lib_name():
+    return "jvm" + shared_lib_extension()
+
+
+def get_java_lib_dir(java_home):
+    for root, _, filenames in os.walk(java_home):
+        if jvm_lib_name() in filenames:
+            return root
+    raise Exception(f"Invalid JAVA_HOME, jvm library not found: {java_home}")
+
+
+def get_std_lib_location(rust_sysroot):
+    lib_name = "std-*" + shared_lib_extension()
+    for root, _, filenames in os.walk(rust_sysroot):
+        for filename in fnmatch.filter(filenames, lib_name):
+            return root, filename
+
+
+def get_java_home():
+    if not os.environ["JAVA_HOME"] is None:
+        return os.environ["JAVA_HOME"]
+    result = subprocess.run(["java", "-XshowSettings:properties", "-version"], capture_output=True, check=True)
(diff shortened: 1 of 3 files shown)


exonum-java-binding/tests_profile.py
```diff
@@ -0,0 +1,111 @@
...
+        self.java_binding_lib_name = self._java_bindings_lib_name()
+        self.exonum_java_name = self._exonum_java_name()
+
+        self._set_environment_variables()
+
+    def _set_environment_variables(self):
+        os.environ["JAVA_HOME"] = self.java_home
+        print(f"JAVA_HOME={self.java_home}")
+        self._set_rustflags()
+        os.environ["PATH"] = self._create_path()
```
dmitry-timofeev: We didn't use to modify `PATH` (on Linux/Mac). Why is that needed?

exonum-java-binding/tests_profile.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def _create_path(self):
+        return os.environ["PATH"] + os.pathsep + \
+               self.java_lib_dir + os.pathsep + \
+               self.std_lib_dir + os.pathsep + \
+               os.path.join(self.project_root, "core", "rust", "target", "debug")
+
+    def _set_rustflags(self):
+        if sys.platform == 'win32':
+            new_rustflags = " "
```
dmitry-timofeev: A space, not empty?

exonum-java-binding/package_app.py
```diff
@@ -0,0 +1,74 @@
...
+    os.environ["RUSTFLAGS"] = rustflags
+
+
+def clear_cargo(manifest_dir):
+    manifest_path = os.path.join(manifest_dir, "Cargo.toml")
+    subprocess.run(["cargo", "clean", "--manifest-path", manifest_path], check=True)
+
+
+if __name__ == '__main__':
+    tests_profile = TestsProfile()
```
dmitry-timofeev: As tests profile sets different rust flags, overwritten here, shall it be created here?

vitvakatu: Yes, probably we can implement `PackageProfile` subclass, I also noticed that.

exonum-java-binding/tests_profile.py
vitvakatu: 🤷‍♂ Copied from old windows scripts revision where wasn't such change

exonum-java-binding/tests_profile.py
vitvakatu: It's needed on Windows, but yes, it must be removed on Linux/Mac

PR #1560, opened 2020-05-28
[ECR-4410] Add migration scripts lookup in Java services runtime

bullet-tooth:
## Overview

---
See: https://jira.bf.local/browse/ECR-4410

### Definition of Done

- [ ] Public API has Javadoc

### DIFF
--- a/exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
+++ b/exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
@@ -19,6 +19,9 @@ package com.exonum.binding.core.runtime;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
+import static java.util.function.Function.identity;
+import static java.util.stream.Collectors.counting;
+import static java.util.stream.Collectors.groupingBy;
 
 import com.exonum.binding.common.crypto.PublicKey;
 import com.exonum.binding.common.hash.HashCode;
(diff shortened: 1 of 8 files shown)


exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
```diff
@@ -587,4 +593,77 @@ private void checkStoppedService(Integer serviceId) {
...
   }
+
+  /**
+   * Returns migration script for the given artifact to perform asynchronous data migration.
+   *
+   * @param artifactId Java service artifact id
+   * @param dataVersion base data version migrate from
+   * @return migration script instance or {@link Optional#empty()} if there is no scripts found
+   * @throws IllegalArgumentException if the provided artifact is not deployed
+   * @throws IllegalStateException if there is scripts incompatibility
```
aleksuss, line 604: if a script is incompatible ? Or `scripts` should be singular.
```suggestion
   * @throws IllegalStateException if there is script incompatibility
```

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
```diff
@@ -587,4 +593,77 @@ private void checkStoppedService(Integer serviceId) {
...
+   */
+  public Optional<MigrationScript> migrate(ServiceArtifactId artifactId, String dataVersion) {
```
aleksuss, line 606: Why don't use `Version` instead of `String` for the `dataVersion` ?

exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceRuntime.java
```diff
@@ -587,4 +593,77 @@ private void checkStoppedService(Integer serviceId) {
...
+                "Migration called on unknown artifactId: " + artifactId));
+
+        if (serviceDefinition.getMigrationScripts().isEmpty()) {
+          logger.info("Service data migration is not required, skipping."
+              + " No scripts found for the given artifact {}", artifactId);
+          return Optional.empty();
+        }
+
+        Version artifactVersion = Version.valueOf(artifactId.getVersion());
+        Version baseDataVersion = Version.valueOf(dataVersion);
```
aleksuss, line 620: If in `ServiceArtifactId` class change a type for `version` field then these transformations would be redundant. I don't insist actually. Maybe there are some valuable reasons to use `String` type for version 😄 I didn't dig deep, to be honest.

bullet-tooth: There are the following points here worth to mention:
- Java <-> native integration requires to have base types (i.e. you should serialize/deserialize objects). So, String is a good choice for passing `Version` through.
- `ServiceArtifactId` is a public model object and to bring there `java-semver` dependency (and force users using it) is not so good idea I believe.
So, we use String and documented that versions should be in a valid _Semantic Versioning_ format.

PR #1561, opened 2020-05-29
Add abstract migration script to base on

bullet-tooth:


exonum-java-binding/core/src/main/java/com/exonum/binding/core/service/migration/MigrationScript.java
```diff
@@ -43,6 +43,9 @@
  *   {@linkplain #targetVersion() target version} in the module.</li>
  * </ul>
  *
+ * <p>You shall usually subclass an {@link AbstractMigrationScript} which implements some
```
aleksuss, line 46: Shall what ? Maybe `use` ? 😄 
P.S. AFAIK Shall uses with I or we ?
```suggestion
 * <p>You usually should use a subclass of {@link AbstractMigrationScript} which implements some methods declared in the interface.  
```

